PR #17, opened 2026-05-05
Load workspace config when indexing tracked repos

alex4o:
## Summary
- load each tracked repo's `.gortex.yaml` before `IndexAll` creates its per-repo indexer
- make `server --track` honor repo-local excludes and workspace config from the tracked destination repo
- add regression coverage for single-repo and multi-repo `IndexAll` respecting workspace excludes

## Test plan
- `go test ./cmd/gortex ./internal/config ./internal/indexer ./internal/excludes`
- `go build -o gortex ./cmd/gortex`

### DIFF
--- a/internal/indexer/multi.go
+++ b/internal/indexer/multi.go
@@ -163,6 +163,7 @@ func (mi *MultiIndexer) indexSingleRepo(entry config.RepoEntry) (map[string]*Ind
 	}
 
 	prefix := config.ResolvePrefix(entry)
+	mi.configMgr.LoadWorkspaceConfig(prefix, absPath)
 	cfg := mi.configMgr.GetRepoConfig(prefix)
 
 	idx := New(mi.graph, mi.registry, cfg.Index, mi.logger)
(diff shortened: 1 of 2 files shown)


zzet:
Hey @alex4o!

Thank you for catching this nasty bug and contribution.

I followed further for the change and found MultiIndexer.IndexRepo at https://github.com/zzet/gortex/blob/a296103c9ac9e2af5f81021c39876bedd78c5e55/internal/indexer/multi.go#L359 has the same bug - calls GetRepoConfig without LoadWorkspaceConfig. 
It's reached from the MCP index_repository tool https://github.com/zzet/gortex/blob/a296103c9ac9e2af5f81021c39876bedd78c5e55/internal/mcp/tools_core.go#L552 On a fresh-start daemon where immediately calls index_repository against a repo with a .gortex.yaml, you'd still get the bug. I believe it make sense to patch it the same way for completeness.
Also, tests only cover the `exclude` symptom and they don't lock down the workspace-slug propagation, which is the other (and arguably more impactful) thing your fix unblocks. Adding a workspace-slug assertion would harden against regressions.

Can you please take a look to those cases? If you don't have time for it - I can accept the PR and apply those patches in the follow up commit later.

alex4o:
Thanks, good catch. I patched both cases:

- `MultiIndexer.IndexRepo` now calls `LoadWorkspaceConfig` before `GetRepoConfig`, so `index_repository` after daemon start sees the repo `.gortex.yaml`.
- The `IndexAll` paths now also stamp workspace/project slugs from `.gortex.yaml` onto emitted nodes, not just exercise the exclude behavior.
- Tests now cover single-repo `IndexAll`, multi-repo `IndexAll`, and `IndexRepo` for both excludes and workspace/project slug propagation.

Verified locally:

```bash
go test ./internal/indexer ./cmd/gortex ./internal/config ./internal/excludes
```

zzet:
The fix is available since v0.20.1 version

PR #20, opened 2026-05-08
fix: drop stale --web flag from generated .mcp.json

pzartem:
## Summary

Commit 7208e77 removed --web from `gortex mcp` and `gortex server`, but the Claude Code init template still hardcoded it, so freshly init'd projects produced an .mcp.json that crashed on launch with an unknown-flag error. Also fixes the matching example in docs/onboarding.md.

## Changes

- internal/agents/claudecode/content.go — the ProjectMCPJSON template still hardcoded --web, plus a stale comment claiming the flag was intentional.
- docs/onboarding.md — onboarding example still showed gortex server --index . --web --watch.
 

## Testing

## Checklist


zzet:
@pzartem good catch, thank you for the fix!

PR #75, opened 2026-06-11
feat(temporal): Go in-workflow query/signal/update handlers + env-default dispatch names

avfirsov:
## Summary

Two small, independent additions to the Temporal Go extractor/resolver, each mirroring an existing pattern in the codebase. They close two gaps in the **Go-side** Temporal surface; the larger cross-language Java→Go bridge is proposed separately for design discussion in #77 (not in this PR).

The two commits are independent and can be reviewed / merged / split separately.

## Changes

**1. Detect in-workflow query/signal/update handler declarations** (`feat(temporal): detect Go in-workflow query/signal/update handler declarations`)

- Recognises `workflow.SetQueryHandler` / `GetSignalChannel` / `SetUpdateHandler[WithOptions]` and emits a `via=temporal.handler` `EdgeCalls` edge from the enclosing workflow carrying `temporal_kind` (`query`/`signal`/`update`) + `temporal_name`.
- Mirrors the existing `via=temporal.register` / `via=temporal.stub` emission exactly, and gives the graph a symmetric, queryable record of the named handlers a Go workflow exposes — the Go counterpart to the Java side's `@QueryMethod` / `@SignalMethod` / `@UpdateMethod` annotation edges.
- High-precision: only the canonical `workflow` receiver alias matches and the handler name must be a string literal (runtime-matched names can't be pinned from a variable), consistent with the existing dispatch detector.

**2. Resolve activity/workflow names from env-var-with-default variables** (`feat(temporal): resolve activity/workflow names from env-var-with-default vars`)

- When a dispatch name is a local variable read from an env var with a literal fallback, resolves the call to that literal default instead of leaving it unresolved:
  ```go
  actName := cmp.Or(os.Getenv("CHARGE_ACTIVITY"), "ChargeCard")
  workflow.ExecuteActivity(ctx, actName, id)   // → activity "ChargeCard"

  name := os.Getenv("K"); if name == "" { name = "ChargeCard" }
  ```
- The lookup is narrow and intra-procedural, **anchored on a literal `os.Getenv` / `os.LookupEnv` read** (so the value is provably env-sourced — not general constant propagation), and only considers assignments lexically before the call.
- The resolved edge lands at the **speculative tier** (`OriginSpeculative`, confidence `0.4`, `MetaSpeculative=true`) rather than `ast_resolved` — the runtime env override may name a different handler than the literal default, so the edge is *present but hidden from default queries*, reusing the existing speculative-edge convention.

## Testing

- [x] All tests pass (`go test -race ./...`) — for the touched packages (`internal/parser/languages`, `internal/resolver`, `internal/indexer` temporal suites). The only failures in my environment are the `internal/indexer` **watcher** tests failing with `inotify ... too many open files` (a WSL fd-limit issue, unrelated — no watcher files are touched).
- [x] New tests added for new functionality — parser-level (`go_temporal_test.go`), resolver-level (`temporal_calls_test.go`), and end-to-end (`temporal_e2e_test.go`) for both features, including negative cases (non-literal handler name, plain non-env variable, env read with no literal default, unresolved-stays-placeholder).
- [ ] Benchmarks run if performance-relevant — both passes are AST-local / reuse existing graph walks; no new whole-graph scans.

## Checklist

- [x] Code follows existing patterns in the codebase (mirrors `applyGoTemporalRegisterMeta`, the speculative-dispatch tier, the existing detector's high-precision stance)
- [x] Language extractor includes `Meta["methods"]` for interfaces (if applicable) — n/a, no interface extraction changed
- [x] Methods have `EdgeMemberOf` edges to their containing type (if applicable) — n/a

## Notes for reviewers

- Both features are deliberately scoped to the **Go side only**. The cross-language join (linking a Java `@WorkflowInterface` start to the Go workflow / a Java `@QueryMethod` to the Go `SetQueryHandler` provider added here) raises edge-model and confidence-tier design questions, so it's filed as #77 rather than bundled here.
- The aliased-import limitation (only the canonical `workflow` receiver is matched) is preserved and pinned by a test, consistent with the existing dispatch detector.

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"ba66ac51-bf76-4fbf-9092-e8ebc75117eb","pid":1532766,"procStart":"15266838","acquiredAt":1781242320901}
--- a/internal/indexer/temporal_e2e_test.go
+++ b/internal/indexer/temporal_e2e_test.go
@@ -133,3 +133,99 @@ func setup(w Worker) {
 	assert.Equal(t, child.ID, stubCall.To)
 	assert.Equal(t, "workflow", stubCall.Meta["temporal_kind"])
 }
+
+// TestTemporalE2E_GoEnvDefaultActivity exercises the env-var-with-literal
+// -default dispatch name: the workflow names its activity through a
+// variable read from os.Getenv with a literal fallback. The pipeline must
+// land the call on the default activity but at the speculative tier.
+func TestTemporalE2E_GoEnvDefaultActivity(t *testing.T) {
+	dir := t.TempDir()
+
+	writeFile(t, filepath.Join(dir, "workflow.go"), `package wf
+
+import (
+	"cmp"
+	"os"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+func OrderWorkflow(ctx workflow.Context, id string) error {
+	actName := cmp.Or(os.Getenv("CHARGE_ACTIVITY"), "ChargeCard")
+	return workflow.ExecuteActivity(ctx, actName, id).Get(ctx, nil)
+}
+`)
+	writeFile(t, filepath.Join(dir, "activity.go"), `package wf
+
+import "context"
+
+func ChargeCard(ctx context.Context, id string) error {
+	return nil
+}
+`)
+	writeFile(t, filepath.Join(dir, "main.go"), `package wf
+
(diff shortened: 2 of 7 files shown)


avfirsov:
Split into two focused PRs so each can be reviewed/merged independently:
- #78 — Go in-workflow query/signal/update handler edges (`via=temporal.handler`)
- #79 — env-var-with-default dispatch name resolution (speculative tier)

Closing this combined PR in favour of those two.

Issue #77, opened 2026-06-12
Cross-language Java→Go Temporal linking (starts_temporal_workflow by canonical name)


avfirsov:
**Is your feature request related to a problem?**

Gortex already resolves the Temporal **Go→Go** workflow→activity layer end-to-end (`internal/resolver/temporal_calls.go`, `internal/parser/languages/golang_temporal.go`) and tags the **Java** side (`@ActivityInterface` / `@WorkflowInterface` / `@ActivityMethod` / `@WorkflowMethod` / `@SignalMethod` / `@QueryMethod` / `@UpdateMethod`) with `temporal_role` and propagates roles to implementors via `EdgeImplements`.

What's missing is the **cross-language bridge**: when a Java service starts a workflow whose implementation lives in a Go repo, the two are connected only by a shared canonical name string — and gortex creates **no edge between the Java node and the Go node**. Concretely:

- `pickGoTemporalTarget` (`internal/resolver/temporal_calls.go`) hard-filters `n.Language != "go"` and is fed only by Go `worker.Register*` calls, so a Java workflow-start can never land on a Go workflow.
- The Java extractor **captures** `@WorkflowMethod(name="...")` / `@ActivityMethod(name="...")` arguments into `edge.Meta["args"]` (`internal/parser/languages/java.go`, via `EmitAnnotationEdge`) but **never parses** them — the resolver only ever uses the bare method name (`from.Name`), so the canonical Temporal type name (Java SDK default `{InterfaceSimpleName}_{methodName}`, or the explicit `name=`) is unavailable for matching.

Net effect: in a polyglot Temporal estate (Java services → Go workflows → Go activities), roughly half the call graph — the Java→Go hop — is invisible.

**Describe the solution you'd like**

Two coordinated pieces (naming them G2 then G1 by dependency order):

**G2 — Java canonical Temporal name extraction.** Parse the captured `@WorkflowMethod(name=...)` / `@ActivityMethod(name=...)` argument and compute the Java SDK default canonical name (`{InterfaceSimpleName}_{methodName}`) when no explicit `name=` is present. Store it on the node (e.g. `temporal_name` meta) so both languages key off the same canonical string.

[...]

PR #78, opened 2026-06-12
feat(temporal): detect Go in-workflow query/signal/update handler declarations

avfirsov:
## Summary

Surface Go-side Temporal **in-workflow handler declarations** (query / signal / update) as graph edges, symmetric with the Java side's existing `@QueryMethod` / `@SignalMethod` / `@UpdateMethod` annotation edges.

This is one half of a split from #75 (the other half — env-default dispatch names — is in a separate PR so each can be reviewed independently).

## Changes

- Recognises `workflow.SetQueryHandler` / `workflow.GetSignalChannel` / `workflow.SetUpdateHandler[WithOptions]` and emits a `via=temporal.handler` `EdgeCalls` edge from the enclosing workflow function, carrying `temporal_kind` (`query` / `signal` / `update`) + `temporal_name` (the handler's string name).
- Mirrors the existing `via=temporal.register` emission (`applyGoTemporalRegisterMeta`) exactly — same edge shape, same stamping site — so the graph gains a queryable, per-workflow record of the named handlers a Go workflow exposes.
- High-precision, consistent with the existing dispatch detector: only the canonical `workflow` receiver alias matches, and the handler name must be a string literal (query/signal/update names are matched by string at runtime, so a non-literal can't be pinned).

## Testing

- [x] All tests pass (`go test -race ./...`) for the touched packages (`internal/parser/languages`, `internal/indexer`). The only failures in my environment are the pre-existing `internal/indexer` **watcher** tests failing with `inotify ... too many open files` (a local WSL fd-limit issue, unrelated — no watcher files touched).
- [x] New tests added — parser-level (`go_temporal_test.go`: query/signal/update detection + negative cases for non-literal name and aliased import) and end-to-end (`temporal_e2e_test.go`).
- [ ] Benchmarks — n/a, the detector is a per-call AST check on the existing walk.

## Checklist

- [x] Code follows existing patterns in the codebase (mirrors `applyGoTemporalRegisterMeta`, the canonical-alias-only stance of the dispatch detector)
- [x] Language extractor includes `Meta["methods"]` for interfaces (if applicable) — n/a
- [x] Methods have `EdgeMemberOf` edges to their containing type (if applicable) — n/a

## Notes

The natural consumer of this data is the cross-language Java→Go match (a Java `@QueryMethod(name="status")` consumer ↔ this Go `SetQueryHandler(ctx, "status", …)` provider), which is proposed for design discussion in #77 — out of scope here; this PR only adds the Go-side provider edges.

### DIFF
--- a/internal/parser/languages/golang.go
+++ b/internal/parser/languages/golang.go
@@ -201,6 +201,12 @@ type goDeferredCall struct {
 	tempKind  string
 	tempName  string
 	tempLocal bool
+	// tempHandlerKind is "query" / "signal" / "update" when this call
+	// is a `workflow.SetQueryHandler` / `GetSignalChannel` /
+	// `SetUpdateHandler` in-workflow handler declaration; tempName then
+	// carries the handler's string name. `via=temporal.handler` meta is
+	// stamped on the emitted edge in the call post-pass below.
+	tempHandlerKind string
 }
 
 type goDeferredTypeRef struct {
@@ -344,6 +350,13 @@ func (e *GoExtractor) Extract(filePath string, src []byte) (*parser.ExtractionRe
 				// `w.RegisterActivity(F)` etc.
 				if name := goTemporalRegisterName(expr.Node, src); name != "" {
 					dc.tempKind = "register_" + kind
+					dc.tempName = name
+				}
+			} else if hkind, ok := goTemporalHandlerKind(receiver, method); ok {
+				// Temporal in-workflow handler declaration:
+				// `workflow.SetQueryHandler(ctx, "name", fn)` etc.
+				if name := goTemporalHandlerName(expr.Node, src); name != "" {
+					dc.tempHandlerKind = hkind
 					dc.tempName = name
 				}
 			}
@@ -668,6 +681,7 @@ func (e *GoExtractor) Extract(filePath string, src []byte) (*parser.ExtractionRe
 			}
 			applyGoGRPCRegisterMeta(edge, c, src, tenv)
 			applyGoTemporalRegisterMeta(edge, c)
+			applyGoTemporalHandlerMeta(edge, c)
 			result.Edges = append(result.Edges, edge)
 			emitGoSpawnEdge(c, callerID, target, filePath, result)
 			continue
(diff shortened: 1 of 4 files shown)


internal/parser/languages/golang.go
```diff
@@ -346,6 +352,13 @@ func (e *GoExtractor) Extract(filePath string, src []byte) (*parser.ExtractionRe
 					dc.tempKind = "register_" + kind
 					dc.tempName = name
 				}
+			} else if hkind, ok := goTemporalHandlerKind(receiver, method); ok {
+				// Temporal in-workflow handler declaration:
+				// `workflow.SetQueryHandler(ctx, "name", fn)` etc.
+				if name := goTemporalHandlerName(expr.Node, src); name != "" {
+					dc.tempHandlerKind = hkind
```
zzet, line 359: @avfirsov I need to think about it. My main concern for now: why not do the same as with the `Register` kind?
```go
dc.tempKind = "register_" + kind`
```

In such a case, it could be 
```go
dc.tempKind = "handle_" + kind`
```

without exploding the schema. 

I'll take a look at other PRs and return to this point later.  There might be benefits for resolution purposes, so I'm not sure atm.

avfirsov: Good question — and you're right that the `goDeferredCall` field is the part worth trimming. Let me separate the two axes, because I think they answer your "benefits for resolution?" doubt:

**The `via` value is the real schema; the struct field is just a parser-internal carrier.** The reason I didn't fold this into `tempKind` is that handler declarations sit on a *different axis* than dispatch/register:

- `temporal.stub` (`ExecuteActivity`) and `temporal.register` are about the **activity/workflow** namespace, and `temporal.stub` edges get **rewritten** by `ResolveTemporalCalls` (placeholder → real function). The `tempKind` values (`activity` / `workflow` / `register_activity` / `register_workflow`) all live on that one axis and feed that one resolver pass.
- `temporal.handler` is the **query/signal/update** namespace and is **not** rewritten — it's pure provider metadata (“this workflow serves query `status`”), the Go-side counterpart of Java's `@QueryMethod`/`@SignalMethod`/`@UpdateMethod` annotation edges. A distinct `via` lets a consumer ask "all handlers" (`via == "temporal.handler"`) without decoding a `handle_*` prefix out of `tempKind`, and keeps the dispatch-intercept condition (`tempKind == "activity" || "workflow"`) untouched.

So I'd keep the distinct `via=temporal.handler` (different namespace, different resolver treatment), but I fully agree about **not adding the field**. I can drop `tempHandlerKind` and carry it as `tempKind = "handle_" + kind`, with `applyGoTemporalHandlerMeta` routing on that prefix — exactly mirroring `register_`. Same for the `tempOutKind` field in #81 (`signalout_`/`querycall_`). That removes the schema growth you're flagging while preserving the semantic/`via` distinction.

No rush — happy to push that refactor whenever you've settled on the direction (or leave it as-is if you decide the separate field reads cleaner). Thanks for taking the time across the three PRs.

PR #86, opened 2026-06-13
feat(temporal): resolve executor struct-field dispatch

avfirsov:
## What

Resolves Temporal dispatch where the activity/workflow **name is a receiver field** whose literal is set at struct construction:

```go
type ActivityExecutor struct{ ActivityName string }
func (e ActivityExecutor) Run(ctx workflow.Context) {
    workflow.ExecuteActivity(ctx, e.ActivityName)        // name = receiver field
}
exec := ActivityExecutor{ActivityName: "ChargeCard"}     // the literal lives here
```

The dispatch only sees `e.ActivityName` (not a literal), so it never resolved. This joins the dispatch to the construction-site literal by `(receiver type, field)`.

## How

**Extractor** (`golang.go` / `golang_temporal.go`):
- a `fieldstr` tree-sitter pattern collects `Type{Field: "literal"}` struct-field assignments, emitted as `via=temporal.executor-field` marker edges carrying `executor_type` / `executor_field` / `executor_value`;
- the dispatch stub is stamped `temporal_name_field` + `temporal_recv_type` when the dispatch argument is `<receiver>.<field>`;
- helper `goCompositeLiteralType`; a `recvTypeByID` map; and the call AST node on the deferred-call struct so the calls loop can inspect the dispatch arg.

**Resolver** (`temporal_calls.go`):
- `resolveTemporalExecutorFields` joins a dispatch's `(recv_type, field)` to a construction literal and rewrites the dispatching method's stub name to that literal — so the existing sweep lands it on the registered handler **and** callers of the activity surface the dispatching method (not the constructor). Wired at the top of `ResolveTemporalCalls`; self-guards; recompute-safe; unique-or-nothing on conflicting literals (matching the const-deref policy).

## Scope

- Only executor-field resolution. Cross-language / const-deref / env-default / wrapper-following are **untouched**.

## Tests

- resolver unit (field-meta stub + executor-field marker → resolved edge),
- extractor unit (`temporal_name_field`/`temporal_recv_type` on the stub; `executor_*` on the marker),
- end-to-end index→resolve test on real Go source.

`go build ./...` and `go test -race ./internal/resolver/... ./internal/parser/languages/... ./internal/indexer/...` pass.

### DIFF
--- a/internal/indexer/temporal_e2e_test.go
+++ b/internal/indexer/temporal_e2e_test.go
@@ -576,3 +576,67 @@ func setupWorker(w Worker) {
 	assert.Equal(t, activity[0].ID, stubCall.To,
 		"the wrapper-following stub must resolve to the registered ChargeCard activity")
 }
+
+// TestTemporalE2E_GoExecutorFieldDispatch exercises the full pipeline for
+// executor struct-field dispatch: a struct method that dispatches via a
+// field, constructed with a string literal, must resolve through the
+// real indexer to the registered activity.
+func TestTemporalE2E_GoExecutorFieldDispatch(t *testing.T) {
+	dir := t.TempDir()
+
+	writeFile(t, filepath.Join(dir, "executor.go"), `package wf
+
+import "go.temporal.io/sdk/workflow"
+
+type ActivityExecutor struct{ ActivityName string }
+
+func (e ActivityExecutor) Run(ctx workflow.Context) {
+    workflow.ExecuteActivity(ctx, e.ActivityName)
+}
+`)
+	writeFile(t, filepath.Join(dir, "activity.go"), `package wf
+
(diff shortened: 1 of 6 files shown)


zzet:
@avfirsov needed a rebase against the main branch

avfirsov:
@zzet rebased onto current `main`. Conflicts resolved against the merged #85 (wrapper-dispatch): kept the shared `callNode` field once and added `goExecutorField`; both pre-passes now run in `ResolveTemporalCalls` (`resolveTemporalWrapperCalls` + `resolveTemporalExecutorFields`); test files keep both suites. `go build ./...` + temporal tests (resolver / parser / indexer) green locally. `mergeable: MERGEABLE`.

PR #87, opened 2026-06-13
feat(temporal): resolve func-returning-constant dispatch names

avfirsov:
## What

Resolves Temporal dispatch where the activity/workflow **name comes from a function that returns a string constant**:

```go
func GetChargeActivityName() string { return "ChargeActivity" }
workflow.ExecuteActivity(ctx, constants.GetChargeActivityName())
```

Today the call-expression argument yields no name and the dispatch stays unresolved. This resolves it to the registered `ChargeActivity`.

## How — reuse the existing const-deref machinery

No new resolver pass and no new meta keys. It plugs into the constant-value sidecar + const-deref step that already resolves `const X = "Y"` dispatch names:

- **extractor** records any func/method whose body is a single `return "<literal>"` into `result.ConstValues` (the same sidecar already used for string consts), keyed by the func node id;
- **`goTemporalNameFromExpr`** learns `call_expression`: a call-expr dispatch arg yields the callee function name (`pkg.GetX()` / `GetX()` → `"GetX"`);
- **`buildConstDerefMap`** now also considers `KindFunction` / `KindMethod` candidates, so the existing deref step maps `"GetX"` → `"ChargeActivity"` → registered handler.

## Scope

- Only func-returning-constant resolution. Wrapper-following / executor-field / cross-language / env-default are **untouched**.

## Tests

- extractor unit (ConstValue emitted for the const-returning func; call-expr dispatch arg → `temporal_name == "GetX"`, incl. `pkg.Get…()`),
- resolver unit (deref of a function node's sidecar value → resolved edge),
- end-to-end index→resolve test on real Go source.

`go build ./...` and `go test -race ./internal/resolver/... ./internal/parser/languages/... ./internal/indexer/...` pass.

### DIFF
--- a/internal/indexer/temporal_e2e_test.go
+++ b/internal/indexer/temporal_e2e_test.go
@@ -640,3 +640,71 @@ func setup(w Worker) {
 	assert.Equal(t, activity[0].ID, stubCall.To,
 		"executor-field dispatch must resolve to ChargeCard")
 }
+
+// TestTemporalE2E_GoFuncReturningConstantDispatch exercises the full pipeline
+// for G2: a func that returns a string literal used as Temporal dispatch arg.
+//
+//	func GetChargeActivityName() string { return "ChargeActivity" }
+//	workflow.ExecuteActivity(ctx, constants.GetChargeActivityName())
+//
+// After indexing, the stub edge must point at the real ChargeActivity node.
+func TestTemporalE2E_GoFuncReturningConstantDispatch(t *testing.T) {
+	dir := t.TempDir()
+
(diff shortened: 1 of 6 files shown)


zzet:
@avfirsov needed a rebase against the main branch

avfirsov:
@zzet rebased onto current `main`. The only conflicts were in the two temporal test files (overlap with #85's added tests) — resolved by keeping both suites. Code files merged cleanly. `go build ./...` + temporal tests (resolver / parser / indexer) green locally. `mergeable: MERGEABLE`.

zzet (approved):

Issue #120, opened 2026-06-19
gortex init OOM


vitorhirota:
while initing a repo, it got killed due to OOM

```
$ gortex init
  ● ● ●      gortex init        28s elapsed
● · · · ●    Index repository
● · ● ● ●
● · · · ●
  ● ● ●

             ⠧  Index repository              parsing · 750 / 781
             ◌  Generate skills
             ◌  Configure adapters
Killed
```

**Environment:**
- OS: Ubuntu 24.04
- Gortex version: 
```
$ gortex version
gortex v0.48.0+1c29216c
  commit:  1c29216c
  built:   2026-06-18T15:32:57Z
  go:      go1.26.4
  os/arch: linux/amd64
```

**Additional context**

from `dmesg`

```
[1564293.498501] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=docker-17e84edfd0e708ec73d5de4198aa22b7cee4924ac095f54791daa352dcfe4f55.scope,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/session-24.scope,task=gortex,pid=809350,uid=1000
[1564293.498565] Out of memory: Killed process 809350 (gortex) total-vm:53888152kB, anon-rss:44579472kB, file-rss:2240kB, shmem-rss:0kB, UID:1000 pgtables:89148kB oom_score_adj:0
```

zzet:
Hey @vitorhirota! Can you provide more information about repository and the machine it was used ran? Indexing whole Linux repository (100GB+) isn't OOM on the test stage. There is some unhandled anomaly I want to understand.

vitorhirota:
i use a ubuntu server 24.04 vm with 64GB RAM and here's the cpu info

```
$ lscpu
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             46 bits physical, 48 bits virtual
  Byte Order:                Little Endian
CPU(s):                      20
  On-line CPU(s) list:       0-19
Vendor ID:                   GenuineIntel
  Model name:                13th Gen Intel(R) Core(TM) i5-13500T
    CPU family:              6
    Model:                   191
    Thread(s) per core:      2
    Core(s) per socket:      14
```

and here some file count. 

```
$ find . -type f -not -path '*/.*' | sed 's/.*\.//' | sort | uniq -c | sort -rn | head -20
    396 pptx
    349 pdf
    347 xlsx
    292 txt
    214 txn
[...]

PR #132, opened 2026-06-20
Limit concurrent large file reads during indexing

caioribeiroclw-pixel:
## Summary

Adds a small back-pressure gate around `os.ReadFile` for files larger than 16 MiB during full indexing.

The goal is to address the #120 failure mode where a repo contains a few hundred large content artifacts (`pdf`, `pptx`, `xlsx`, vector/RAG outputs, etc.) and `workers == NumCPU` allows many huge files to be read into memory at once before extraction has a chance to flush/compact anything.

This keeps normal source-file parallelism unchanged, but limits large-file reads to at most 2 concurrent reads (1 when the worker count is <= 1). It is deliberately narrow: it does not change `MaxFileSize` defaults, skip behavior, parsers, graph logic, or DB batching.

## Why this shape

For code-heavy repos, most files are small, so existing worker throughput remains the same.
For content-heavy repos, peak read-side memory is bounded closer to:

`min(2, workers) × large_file_size`

instead of:

`workers × large_file_size`

which matches the issue discussion around memory being dominated by in-flight file reads/extractor expansion rather than the graph itself.

## Test

```bash
go test ./internal/indexer -run 'TestLargeFileReadParallelism|TestIndex_SizeSkipTelemetry|TestIndex_ExtractTimeoutSkip|TestSizeSkipNode'
```

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -142,7 +142,7 @@ type Indexer struct {
 	// path so a query word matching it doesn't earn a useless uniform
 	// path-field boost across every document. "" disables the de-weighting.
 	projectName string
-	logger        *zap.Logger
+	logger      *zap.Logger
 
 	// Crash-isolation parser pool, lazily created and then reused
 	// across single-file re-indexes so the watcher hot path never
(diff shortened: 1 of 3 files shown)


zzet:
@caioribeiroclw-pixel thank you for the help!

PR #133, opened 2026-06-20
CLI parity with the MCP surface, a CLI-driven skill, and a context-budget receipt

### LINKED ISSUE #108: Feature: CLI parity with MCP + first-class plain markdown SKILL.md, so agents can use Gortex without paying the 175-tool context tax
## Summary

`gortex install` is the documented path to use Gortex from a coding agent. It wires the MCP server in and — by default — eagerly publishes 175 tool definitions into `tools/list`. Every one of those JSON schemas sits in the agent's context every turn, used or not. For agents that don't want a constantly-loaded MCP surface, there is no first-class alternative: the CLI exposes only a fraction of the daemon's capabilities, and there is no shipped, low-cost markdown skill for on-demand use.

This issue asks for two things together:

1. **A plain-markdown `SKILL.md`** shipped with Gortex, designed to be dropped into `~/.claude/skills/gortex/` (or any agent's user-level skill surface) and loaded on demand — no MCP transport, no `install` step, no permanent context cost.
2. **CLI parity with the MCP surface** for the edit-safety and memory tool families, so the skill above can actually orchestrate the workflows the project already documents as mandatory in `CLAUDE.md`.

The two are inseparable: a markdown skill is only useful to the extent the CLI underneath can do the work it describes.

## Context cost problem

`tools/list` carries the full schema of every registered tool. With 175 tools eagerly published, that's a substantial baseline token cost on every turn, regardless of whether any Gortex tool is called.

The project already acknowledges this — `GORTEX_LAZY_TOOLS=1` exists precisely to defer registration and let the agent discover tools on demand via `tools_search`. That helps, but:

- It still requires the MCP transport and the install step.
- It still requires the host to honour `notifications/tools/list_changed`.
- It is a workaround inside the MCP model, not an alternative to it.

For agents that don't want any MCP surface mounted, or that want to load Gortex capabilities only for the subset of sessions that need them, the cheap option today is to shell `gortex` from a markdown skill. That works — until you hit the coverage gap.

## CLI coverage gap

The CLI is already daemon-backed. `cmd/gortex/query.go` shows the pattern: `requireDaemonTool(repoPath, \"<mcp-tool>\", args)` shells the daemon's executor and calls the same MCP tool handler the server uses. There is no architectural reason every MCP tool can't have a matching CLI verb.

Today's CLI covers roughly 25 verbs — read-side queries (`query symbol/usages/callers/deps/dependents/implementations/calls/stats`) plus rendered verdicts (`review`, `audit`, `wakeup`, `context`, `docs`, `prs`). The MCP surface is ~175 tools. The workhorses the `CLAUDE.md` \"MANDATORY workflow\" instructs the agent to call are CLI-missing:

**Edit-safety pipeline (none of these have CLI verbs):**
- `smart_context` — the entry point for every new task
- `get_editing_context` — required before editing a file
- `verify_change` — required before changing a function signature
- `get_edit_plan` — dependency-ordered refactor plan
- `batch_edit` — atomic multi-file apply
- `preview_edit` / `simulate_chain` — speculative execution
- `check_guards` — required after every edit
- `get_test_targets` — required after every edit
- `change_contract` — pre-commit verdict
- `safe_delete_symbol`

**Memory layer (none of these have CLI verbs):**
- `save_note` / `query_notes` / `distill_session` — session memory
- `store_memory` / `query_memories` / `surface_memories` — durable memory

**Other high-value tools without CLI parity:**
- `analyze` (61 kinds) — only fragments are reachable via `audit` / `review`
- `flow_between`, `taint_paths`
- `find_clones`, `find_co_changing_symbols`
- Overlay sessions (`overlay_push` / `overlay_list` / `overlay_drop`)
- `feedback` — required by the documented workflow to improve future context

So the trade today is binary: pay the 175-def context tax for the full surface, or use the CLI and lose every guardrail and memory step the project itself calls mandatory.

## Ask

1. **Add CLI verbs for the edit-safety, memory, and analysis tools** listed above. They can be thin shells over the same handlers — the executor plumbing already exists. Reasonable groupings: `gortex edit ...` (context / verify / plan / preview / batch / guards / test-targets), `gortex memory ...` (note / memory / distill / surface), `gortex analyze --kind <one-of-61>`.

2. **Ship a plain-markdown `SKILL.md`** alongside the existing per-community generated skills — a single user-level skill that drives Gortex via the CLI, loaded on demand. Frontmatter triggers on Gortex-related work; body walks the same workflow `CLAUDE.md` documents (`stats` → `smart_context` → `surface_memories` → `get_editing_context` → `verify_change` → `batch_edit` → `check_guards` → `get_test_targets` → `feedback`), with every step expressed as a `gortex` shell command. No MCP transport required.

3. **Document the skill+CLI path as a first-class consumption pattern** in `docs/skills.md`, alongside the MCP/install path. Make the trade explicit: full surface via MCP (high context baseline, push notifications, overlay sessions) vs. on-demand surface via skill+CLI (zero baseline, no push channels, no overlays).

## Why this is worth doing

- Lowers the barrier for any coding agent that doesn't support MCP, or where MCP is impractical.
- Gives MCP-capable agents a low-context fallback for sessions that don't need the full surface.
- Closes the gap between what `CLAUDE.md` says is mandatory and what is reachable without MCP.
- Aligns with the project's own framing in the README — \"access via CLI, MCP Server, and API\" — by making the CLI path actually competitive with the MCP one.

## Out of scope

- Push notifications (`subscribe_*`) — fundamentally streaming, don't fit a CLI shape.
- Overlay sessions tied to MCP session lifecycle — can be CLI-exposed but require explicit session-id handling and are a separate design.

zzet:
Closes #108.

## What this does

Gives the `gortex` CLI parity with the MCP tool surface and ships a plain-markdown skill that drives the documented workflow through CLI verbs — so an agent can use Gortex's edit-safety and memory tools without mounting the MCP surface and paying its per-turn tool-schema context cost.

The enabling fact: the CLI already reaches the daemon over the same socket the MCP server uses, and the daemon dispatches tool calls **by name** regardless of which tools are eagerly published. So every tool is already reachable; this is a CLI presentation layer plus a skill, a receipt, and docs — no handler changes.

## Layers

**Single source of truth.** `AnalyzeKinds()` replaces three drifting copies of the analyze-kind list (the dispatch switch, the schema description, two error strings), pinned by an AST test. `ToolDescriptor` / `ToolDescriptors()` joins the previously-disjoint tool metadata (category, mutating-vs-read, preset membership, summary) into one record, surfaced in `tool_profile`.

**Generic passthrough + discovery.** `gortex call <tool>` invokes any registered tool by name, with httpie-style `--arg key=value` coercion (`key:=raw` forces raw JSON), a `--json` / `--json-file` / stdin base object, `--format`, and `--dry`. `gortex tools list|search|describe|receipt` is the CLI analogue of `tool_profile` / `tools_search`.

**Curated verb groups.**
- `gortex edit` — context, verify, plan, preview, simulate, batch, apply, symbol, rename, guards, tests, contract, safe-delete. A shared structured-input helper accepts inline / file / stdin JSON; `--change 'id=sig'` is sugar for verify; safe-delete is a dry run unless `--apply`.
- `gortex memory` — note, notes, distill, store, recall, surface.
- `gortex analyze --kind <k>` (+ `analyze kinds`), `gortex flow`, `gortex taint`, `gortex clones`, `gortex feedback record|query`.

**The `gortex-cli` skill.** A single user-level, plain-markdown skill installed by `gortex install`, scoped to `Bash(gortex:*)`, that walks the workflow through `gortex` shell verbs with no MCP transport mounted. Covered by the agents-render drift fence across all 17 adapters.

**Context-budget receipt.** `gortex tools receipt` emits an inspectable artifact: the transport, the advertised-versus-deferred tool counts, `registered_tool_schemas: 0` on the CLI path, and — when `GORTEX_SESSION_ID` is set — the session's `cli_verbs_used` and a derived `safety_steps_run`, so a CLI path that silently skips verify / guards / tests is visible rather than merely asserted to be equivalent. Backed by a new consent-free `cli_events` sidecar table that opens no database on the default interactive path.

## Tests

Each verb group has stub-seam lowering tests (flags → exact tool + args) that run without a live daemon. Also covered: the analyze-kinds AST drift test, the tool-catalog completeness test, the structured-input parser, the receipt (both daemon-reachable and no-daemon), the `cli_events` round-trip, and the verb→safety-step derivation. Full build, `go test -race` on the touched packages, `go vet`, golangci-lint, the wire-contract golden, and `agents render --check` (17 adapters) all pass. Binary smoke: `analyze kinds` → 73, `call --dry` lowers correctly, `tools receipt` emits a real receipt.

## Docs

`docs/cli.md` gains the full verb reference, the structured-input convention, and a consumption-path trade-off table; `docs/skills.md` adds the skill as a zero-schema path; `docs/mcp.md` notes the two front doors over one set of handlers; `CLAUDE.md` is softened so the mandatory workflow reads as reachable via MCP **or** the equivalent `gortex` CLI verbs.


grantcarthew:
Wow, that was quick. Well done guys. Thanks.

PR #141, opened 2026-06-21
fix(lsp): enable resolve-time pyright helpers

### LINKED ISSUE #140: Python imported calls are not resolved by resolve-time Pyright LSP
## Summary

Python imported function calls can remain as synthetic external-call stubs even when `pyright` can resolve the call target via `textDocument/definition`.

That makes `query callers` / `get_callers` report a real Python function as unused when the caller imports it from another module and calls it by the imported name.

## Minimal shape

```python
# src/pkg/dispatch.py
def dispatch_attempt():
    pass
```

```python
# src/pkg/scheduler.py
from pkg.dispatch import dispatch_attempt

def scheduler_tick():
    dispatch_attempt()
```

## Observed behavior

With `pyright` enabled and available, Gortex indexes the callee function, but the call edge can stay pointed at an external dependency placeholder:

```text
src/pkg/scheduler.py::scheduler_tick
  -> external-call::dep::pkg.dispatch.dispatch_attempt
  kind=calls
  origin=text_matched
```

Then `query callers src/pkg/dispatch.py::dispatch_attempt` returns no callers / likely-unused, even though `scheduler_tick` calls it.

I also probed `pyright-langserver` directly for the same call site shape, and it returns the correct definition location in `dispatch.py`. So this does not appear to be a Pyright import-resolution failure.

## Likely root cause

The daemon's resolve-time LSP helper registration currently appears to be TypeScript/JavaScript-only:

- daemon warmup registers `typescript-language-server` helpers gated by TS/JS project markers
- Python repos are counted/skipped as non-TS repos in the resolve-time helper log
- `pyright` still runs as a semantic enrichment provider, but it is not used by the hot-path resolver that rewrites unresolved call edges

There is one more wrinkle: `ResolverHelperRegistry` stores one helper per repo prefix, so adding a second helper naively would make mixed Python/TS repos fight over the same prefix. The fix likely needs an extension-aware mux per repo prefix.

## Expected behavior

For Python repos where `pyright` is available and the repo has normal Python project markers, resolve-time LSP should route `.py` call edges through `pyright` in the same way TS/JS files route through `typescript-language-server`, so the call edge becomes:

```text
src/pkg/scheduler.py::scheduler_tick
  -> src/pkg/dispatch.py::dispatch_attempt
  origin=lsp_resolved
```

and `get_callers(dispatch_attempt)` includes `scheduler_tick`.

## Version

Observed on:

```text
gortex v0.50.0+446f223b
```

I have a small PR ready that adds an extension-aware resolver-helper mux and registers `pyright` for Python repos without changing the TS/JS path.

cometto2007:
## Summary

Enable Python resolve-time LSP by routing per-repo helper calls through an extension-aware mux:

- add `ResolverHelperMux` so one repo prefix can serve `.ts/.tsx` via `typescript-language-server` and `.py` via `pyright`
- register Pyright-backed resolve-time helpers for repos with standard Python markers
- keep the existing TS/JS helper path and lazy provider lifecycle intact
- update the feature docs to mention Python on the hot path

## Root Cause

`pyright` could already participate in semantic enrichment, but daemon resolve-time LSP helper registration was hardcoded around `typescript-language-server`. Python repos therefore reached the resolver without a helper, leaving import-bound Python calls at the heuristic/external-call tier even when Pyright could answer `textDocument/definition`.

## Validation

```text
go test ./internal/semantic/lsp ./internal/serverstack
go test ./cmd/gortex
```

Fixes #140


zzet:
Hey @cometto2007, are you planning to add contributions to the PR or this is complete (it still in the draft state)?

cometto2007:
Thanks for checking! This is complete from my side; I had opened it as a draft while validating the approach and forgot to flip it afterwards. I marked it ready for review now. CI is green, and I am not planning additional changes unless you would prefer a different shape for the helper mux or Pyright detection.

cometto2007:
Thanks again for merging #141. I did the proper end-to-end smoke test after the merge and found that #141 fixed the registration layer, but there was one more Pyright-specific initialization issue underneath it.

What happened:

- `main` now registers a resolve-time Pyright helper for Python repos.
- Pyright itself can resolve the imported call in the fixture.
- But Gortex was initializing Pyright with `rootUri` while omitting the primary `workspaceFolders` entry when there were no additional roots.
- In that initialization shape, Pyright returns `null` for `textDocument/definition`, so Gortex still falls through to the external-call placeholder path.

I opened follow-up PR #154 with the narrower fix and a real Pyright regression test:

- always include the primary root in `workspaceFolders`
- add an integration test for cross-file `from ... import ...` resolution
- validated with `go test ./...` plus an isolated daemon smoke test where `get_callers` now returns the real `lsp_resolved` Python edge

So #141 was useful and necessary, but not quite sufficient for the user-facing Python caller result.

PR #162, opened 2026-06-23
feat(graph,mcp): annotate_nodes — write external metadata back onto graph nodes

avfirsov:
## What

Adds a sanctioned, additive **write-back path for external metadata** so a downstream tool — an LLM enrichment pass, an analysis stage, an editor extension — can attach its own metadata to existing graph nodes **without re-indexing** and **without racing the sharded in-memory store**.

Today there is no safe way for a caller holding a `graph.Store` (rather than a concrete `*Graph`) to mutate `Node.Meta`: the shard lock the in-memory backend takes is private, so reaching into `GetNode(id).Meta` directly races the sharded map and panics on a live daemon. This PR closes that gap.

## How

- **`graph.Store.MergeNodeMeta(id, kv) (changed, found)`** — the only sanctioned `Node.Meta` mutation path for a `Store` holder. Additive and **idempotent** (deep-equal delta via the pure `metaDelta` helper), taken under the node's shard write lock, and it never touches structural fields (id / kind / name / path / lines). Implemented for **both** the in-memory and SQLite backends and covered by a shared store-conformance case so the two behave identically.
- **`annotate_nodes` MCP tool** (also served at `POST /v1/tools/annotate_nodes` through the shared registry — no separate HTTP code): merges a free-form per-node `meta` map under a caller-chosen **`namespace`** (default `ext`). Every key is stored as `<namespace>_<key>`, so an annotation can never shadow an indexer-owned Meta key. Optionally adds idempotent `semantically_related` edges between node pairs. Returns `{annotated, unchanged, not_found, edges_added}`.

The merge is deliberately scoped: never deletes keys, never mutates structural data, non-fatal per item (an unknown id is recorded in `not_found`, the batch continues).

## Tests

- Pure `metaDelta` unit (delta/idempotency/nil-handling).
- In-memory + SQLite **conformance** (`MergeNodeMeta`): additive merge, deep-equal idempotency, found semantics for an unknown id, lazy `Meta` init, structural fields untouched.
- MCP handler: round-trip, idempotent re-run, `namespace` prefixing (incl. no double-prefix), `semantically_related` edge add + dedup, default score, bad input, and a registration guard.

All green; `go build ./...`, `go vet`, and `gofmt` clean on the changed files.

## Notes / scope

- Cross-restart durability of annotations is intentionally out of scope for this v1: the merge mutates in-memory state and rides the gob+gzip shutdown snapshot (the SQLite backend re-persists the node). Happy to follow up with explicit per-call persistence if you'd prefer.

### DIFF
--- a/internal/graph/meta_merge.go
+++ b/internal/graph/meta_merge.go
@@ -1,0 +1,89 @@
+package graph
+
+import "reflect"
+
+// metaDelta returns the subset of kv whose value differs from (or is
+// absent in) existing — i.e. the keys a merge would actually change.
+//
+// It is a pure calculation: no locks, no I/O, no mutation of either
+// argument. Equality is reflect.DeepEqual so the comparison is correct
+// for the value shapes that ride on Node.Meta — scalars (string, the
+// JSON-decoded float64 for numbers), []string / []any tag slices, and
+// nested map[string]any blobs — where a == comparison would either
+// panic (slices/maps are not comparable) or compare identity rather
+// than contents.
+//
+// A nil existing is treated as "no keys present", so every entry in kv
+// is returned. An empty/nil kv returns nil (nothing to merge). The
+// returned map is freshly allocated and shares no structure with kv's
+// values beyond the value pointers themselves (callers store them as-is
+// — these are caller-owned, JSON-decoded values).
+//
+// Idempotency falls straight out of this: once a key has been written,
+// a second metaDelta over the now-updated map omits it, so the second
+// MergeNodeMeta reports changed=false.
+func metaDelta(existing, kv map[string]any) map[string]any {
+	if len(kv) == 0 {
+		return nil
+	}
+	var delta map[string]any
+	for k, v := range kv {
+		if existing != nil {
+			if cur, ok := existing[k]; ok && reflect.DeepEqual(cur, v) {
+				// Key already present with an equal value — nothing to do.
+				continue
+			}
+		}
+		if delta == nil {
+			delta = make(map[string]any, len(kv))
+		}
+		delta[k] = v
+	}
+	return delta
+}
+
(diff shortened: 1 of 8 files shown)


zzet:
@avfirsov thank you for submitting an interesting feature.

Overall, it looks good to me, excluding only one aspect:
The Meta attribute contains data that Gortex relies on, and external writes to that attribute can be both useful and harmful. Considering the prefix, half of the potential issues are mitigated; however, an uncontrolled increase in the JSON column can have the following impact on performance (which can lead to a chain of consequences, up to MCP tool penalisation during the session).

What about adding an additional attribute to the table and persisting user input metadata into an additional JSON column, and loading it when it is necessary (not on every node extraction)?

Second point, which may not be expected from the user POV - an in-session reindex of a changed file evicts and re-parses that file's nodes — which drops both the merged ext_* Meta and the semantically_related edges, on the SQLite backend too (reindex INSERT OR REPLACE overwrites Meta). So annotations are lost on any edit to the annotated file, not just on restart.

Also, Minor DRY: the SQLite path re-implements the metaDelta loop inline instead of reusing the pure helper. It's forced today (the helper is unexported in package graph); exporting graph.MetaDelta and calling it would remove the duplication the PR itself flags in a comment.

zzet:
Thanks again for this contribution. The feature is valuable, but the current implementation cannot be merged because external annotations enlarge the core Node.Meta payload and are lost when an annotated file is re-indexed.

To continue, the PR needs:
  - Separate, lazily loaded storage for external annotations.
  - Annotation and semantic-edge survival across file re-indexing.
  - Reuse of the common metadata-delta helper.
  - Rebase and passing current CI.

Please let us know by August 14, 2026 if you would like to continue. Otherwise, I’ll close this PR as inactive; it can be reopened or replaced by a new PR later.

PR #163, opened 2026-06-23
feat: add dry-run intake manifest

caioribeiroclw-pixel:
## Summary

Implements the privacy-safe intake diagnostic discussed in #120:

- adds `gortex init --dry-run-intake`
- exits before parsing/extraction/storage/writes
- emits JSON with aggregate intake counts, top admitted/skipped extensions by bytes, hashed top admitted dirs, hashed largest admitted files, repo ref, and explicit privacy flags
- reuses the index walk's language/ignore/prune/max-file-size gates so the report answers "what enters the expensive index pipeline?" without exposing raw paths or file contents

This is meant for private mixed repos where maintainers need to distinguish corpus admission from downstream extractor/flush memory growth.

## Validation

- `go test ./internal/indexer -run TestNonexistent -count=0`
- `go test ./cmd/gortex -run 'TestInitDryRunIntakeJSONDoesNotWrite|TestInitDryRunSkipsProjectMarker' -count=1`
- `go test ./cmd/gortex -run TestNonexistent -count=0`
- `git diff --check`
- smoke: built `/tmp/gortex-intake`, ran `gortex init --dry-run-intake` on a temp repo with source + binary-ish assets, verified schema/privacy flags/files counts

Full `go test ./internal/indexer ./cmd/gortex` was attempted first, but exceeded the local 180s command timeout while compiling/running the larger cmd/indexer package set; the targeted compile/tests above passed.

### DIFF
--- a/cmd/gortex/init.go
+++ b/cmd/gortex/init.go
@@ -61,13 +61,14 @@ var (
 	initSkillsMaxSkills int
 
 	// Non-interactive / reporting knobs
-	initYes         bool
-	initInteractive bool
-	initAgents      string
-	initAgentsSkip  string
-	initJSON        bool
-	initDryRun      bool
-	initForce       bool
+	initYes          bool
+	initInteractive  bool
+	initAgents       string
+	initAgentsSkip   string
+	initJSON         bool
+	initDryRun       bool
+	initDryRunIntake bool
+	initForce        bool
 )
 
 var initCmd = &cobra.Command{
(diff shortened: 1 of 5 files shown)


caioribeiroclw-pixel:
Follow-up after the initial CI pass: I added a small `docs/cli.md` entry so the diagnostic is discoverable, not only hidden in flag help.

For testing from this PR branch before a release, the intended private-repo workflow is:

```bash
git fetch https://github.com/caioribeiroclw-pixel/gortex.git caio/dry-run-intake-manifest
# build this branch, then from the affected repo:
gortex init --dry-run-intake --json > intake.json
```

The JSON report should be safe to share by default: `raw_paths_included=false`, `raw_file_contents_included=false`, hashed path-depth buckets only, plus admitted/skipped counts and top file classes by bytes. That should be enough to decide whether #120 is still corpus admission, extractor expansion, or later accumulation.

PR #185, opened 2026-06-27
Fix Spring datasource config and @Bean consumer resolution

mhmmdd:
Detect Spring application config files with Windows paths, stamp @ConditionalOnProperty and @ConfigurationProperties reads onto the owning Java node, and capture method params so @Bean consumers link back to matching factory methods.

Adds regression coverage for datasource config reads, DataSource bean consumer linkage, and Spring config path detection.

## Summary

Fixes Spring datasource/config graph gaps in the Java indexer so Spring config readers and explicit `@Bean` consumers are linked to the right graph nodes.

## Changes

- Normalize Spring application config file paths so Windows `\` paths are detected.
- Parse `@ConditionalOnProperty` and stamp the resolved config keys on the owning Java class or method.
- Stamp `@ConfigurationProperties(prefix)` reads on the annotated Java class instead of the synthetic annotation node.
- Capture Java method parameter source text so Spring `@Bean` consumer methods can link back to matching factory methods by type.
- Add regression tests for datasource config reads, `DataSource` bean consumer linkage, and Windows Spring config path detection.

## Testing

- [x] All tests pass (`go test -count=1 ./...` in clean WSL/LF clone)
- [x] Benchmarks run if performance-relevant

## Checklist

- [x] Methods have `EdgeMemberOf` edges to their containing type (if applicable)

### DIFF
--- a/internal/contracts/spring_config.go
+++ b/internal/contracts/spring_config.go
@@ -38,6 +38,7 @@ func springConfigKeyID(key string) string {
 // springConfigFile reports whether path is a Spring application config file and
 // returns its profile (the `-prod` in application-prod.yml; "" for the base).
 func springConfigFile(path string) (profile string, ok bool) {
+	path = strings.ReplaceAll(path, "\\", "/")
 	base := path
 	if i := strings.LastIndexByte(path, '/'); i >= 0 {
 		base = path[i+1:]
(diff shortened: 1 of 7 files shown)


zzet:
@mhmmdd thank you for your contribution!

PR #189, opened 2026-06-27
fix(daemon): keep daemon responsive when LSP enrichment wedges

pacaya:
keep daemon responsive when LSP enrichment wedges
A repo containing .NET projects could make the daemon look crashed
(control socket present, every tool call i/o timeout, CPU pegged): the
C# LSP gets stuck loading an MSBuild/Roslyn workspace and an enrichment
findReferences Call blocks forever, while the status path runs a slow
SQLite COUNT … GROUP BY under the coarse controller mutex and stalls
every queued control request.

Fix https://github.com/zzet/gortex/pull/1 (core) — bound a single post-initialize LSP request:
- lsp.Client gains an atomic callTimeout + SetCallTimeout; Call selects
  on a timer (nil channel = unbounded, preserving prior behaviour).
- Provider.ensureClient applies it AFTER the initialize/initialized
  handshake (default 30s, GORTEX_LSP_CALL_TIMEOUT, "0"/"off"/"none" to
  disable) so a cold workspace load stays unbounded but a server that
  wedges mid-session can no longer block enrichment forever.
- Manager.runEnrichOne runs the provider on a goroutine bounded by a
  per-repo deadline (default 10m, GORTEX_LSP_ENRICH_TIMEOUT); on
  deadline it logs and moves on while the detached goroutine drains.

Fix https://github.com/zzet/gortex/pull/2 — read-loop spin guard: readResponses drops the connection after
maxMalformedFrames (64) consecutive Content-Length-less header blocks
instead of spinning on continue and pinning a core.

Fix #3a — lock-light Status: AllRepoMemoryEstimates is memoised in the
SQLite store with a 3s TTL (mutex held across recompute = singleflight),
and realController.Status computes it off the controller mutex.

Tests: client_timeout_test.go, enrich_deadline_test.go,
store_memest_test.go. go build ./... clean; go test -race green on
internal/semantic, internal/semantic/lsp, internal/graph/store_sqlite,
cmd/gortex.
- 

## Testing

## Checklist

### DIFF
--- a/cmd/gortex/daemon_controller.go
+++ b/cmd/gortex/daemon_controller.go
@@ -613,6 +613,21 @@ func resolveSearchBackend(b search.Backend) searchBackendInfo {
 // fields (PID, uptime, socket, session count) are filled in by the
 // daemon itself before the response goes out.
 func (c *realController) Status(_ context.Context) (daemon.StatusResponse, error) {
+	// Compute the per-repo memory estimate BEFORE taking the coarse
+	// controller mutex. On the SQLite backend AllRepoMemoryEstimates is a
+	// COUNT … GROUP BY scan that turns pathologically slow under
+	// enrichment write load; holding c.mu across it stalls every other
+	// control request (status / track / reload) queued on the mutex — the
+	// daemon-looks-crashed symptom. Snapshot the graph handle under a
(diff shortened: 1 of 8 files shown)


zzet (approved):

zzet:
@pacaya thank you for your contribution!

PR #193, opened 2026-06-28
fix(hooks): route file-indexed probe through daemon socket so hard-de…

vitaliyslion:
## Summary

The PreToolUse hook is meant to **hard-deny** whole-file `Read`/`cat`/`Edit`/`Write`
of indexed source and redirect to graph tools. Commit `7208e77a` ("server: drop
--web, salvage /v1/graph + /v1/events, delete internal/web", 2026-04-18) deleted
the `internal/web` server, removing the `/api/graph/file` endpoint the hook
probed. Since then `queryFileIndexed` always returned `(false, 0)`, so the deny
silently degraded to soft guidance for **every** agent. Masked because the tests
mocked the endpoint and the fallback still printed advice.

This re-points the file-indexed probe at the daemon's MCP socket
(`get_file_summary`) - the always-present, authoritative graph owner.

> ⚠️ Intentionally hook-local and minimal, **not** a complete fix. The deeper
> path-resolution work is left as inline TODOs for a maintainer to direct; I'm
> happy to add follow-up commits once the direction is decided.

## Changes

- Replace the dead `/api/graph/file` HTTP probe with a daemon socket call to
  `get_file_summary`, behind a `fileIndexedFn` test seam.
- Thread `cwd` (from `HookInput.CWD`) through `enrich -> enrichRead/enrichBash/
  enrichEdit/enrichWrite -> queryFileIndexed` so the probe scopes to the workspace
  and resolves the repo-relative graph key.
- Memoize the daemon status fetch (5s TTL) so a wide `postGlob` makes ~one
  control-status round-trip instead of one per file.
- Migrate the HTTP-mock hook tests to stub the new seam.
- **TODOs left for maintainer review** (documented inline): route the path
  through the server's existing `resolveFilePath` (abs -> repo-relative + symlink
  confinement) instead of hand-rolling root resolution; reuse one MCP connection
  per batch / count-only probe. Note: the `file-callers` + `symbol-at` probes in
  `posttooluse.go` were never served — out of scope here.

## Testing

  ## Checklist

### DIFF
--- a/internal/hooks/edit_test.go
+++ b/internal/hooks/edit_test.go
@@ -2,8 +2,6 @@ package hooks
 
 import (
 	"encoding/json"
-	"net/http"
-	"net/http/httptest"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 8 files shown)


zzet (approved):

zzet:
@vitaliyslion thank you for your contribution!

I'll take a look at TODOs performance references later

Issue #172, opened 2026-06-26
Intent-based default scoping for multi-repo workspaces


pacaya:
# Proposal.

*A discussion starter - not a finished decision. I think the workspace/project/repo model is the right foundation; this is about the **default breadth** of a query and the **uniformity** of the scope overrides. Corrections welcome, especially where I've misread the intent.*

## Summary

In a workspace that indexes many repos, the query tools default to a broader scope than a developer usually wants while working inside one repo. Two changes, separable:

- **Layer A - uniform scope overrides.** Let every scoped tool accept `repo` / `project` / `workspace` / `scope`, resolved the same way and clamped to the session workspace. Pure capability; **no behavior change**.
- **Layer B - intent-based defaults.** Default *locate* tools (symbol search) to the **current repo** and *reach* tools (usages/callers/contracts) to the **workspace**, with `project` as an explicit-only middle rung. **Behavior change; config-gated, default off.**

Layer A stands on its own. Layer B is opt-in.

## How scoping works today (my reading - please correct)

- Three nesting levels: **repo** (`Node.RepoPrefix`) ⊂ **project** (`Node.ProjectID`, soft) ⊂ **workspace** (`Node.WorkspaceID`, hard isolation boundary).
- A daemon session resolves its scope from cwd - `ScopeForCWD` returns `(workspace, project, repo)` - and caches the home repo, but uses it **only for relevance ranking**, never for filtering.
- There are **two** scope paths with **two different defaults**:
  - *Locate / traversal* (`search_symbols`, `find_usages`, `get_callers`) → `resolveQueryScope` → `QueryOptions{WorkspaceID, ProjectID}`. Defaults to the session **project**.
  - *Analyze / by-id / review* (`get_symbol`, `analyze`, `review`, sast) → `resolveRepoFilter` → a repo allow-set. Defaults to the whole **workspace**.
- `QueryOptions` has no repo dimension, so the locate path can't filter below project even though the session already knows the exact repo.

## Problem

1. **Default is too broad.** Working inside one repo, `search_symbols` returns hits from every repo in the project (or workspace). The noise scales with workspace size.
2. **The two paths are inconsistent.** They default differently (project vs workspace) and expose different overrides - `find_usages` can't take a `repo:`, and symbol search can't take a saved `scope:`.
3. **No single granularity is a good default for everything:**
   - *Locate* ("where is X defined / find a symbol") wants the **repo**.
[...]

PR #202, opened 2026-06-30
feat(scope): intent-based default scoping for multi-repo workspaces

pacaya:
Implementation of the issue #172 

In a multi-repo workspace, graph queries had no principled default scope: every tool read the full index, so results leaked across project and workspace boundaries unless the caller manually scoped each call. This PR makes scope a first-class, intent-derived default while keeping explicit overrides uniform and workspace a hard isolation boundary.

## Behavior change (default ON)

Ships enabled by default (`scope.intent_defaults: true`, override with `GORTEX_SCOPE_INTENT_DEFAULTS=0`). In a multi-repo workspace, results that previously spanned every indexed repo now default to the session's repo (lookups) or workspace (reachability / analysis). Set the flag to false to restore the prior whole-index behavior. Single-repo setups are unaffected.

## Two-layer model

- Layer A — uniform overrides: an explicit `repo:` / `project:` / `scope:` (workspace) argument overrides the default identically across query, MCP, and analyze tools. `repo:"*"` is a sentinel escape hatch meaning "all repos", not a post-filter.
- Layer B — intent-based defaults: with no explicit scope, the default is chosen by the tool's intent —
    - Locate (symbol search / lookup) — repo (narrowest)
    - Reach  (callers / usages / dependents / walk) — workspace
    - Analyze — workspace Defaults only ever narrow; they never widen past the session workspace.

## Analyze made scope-aware

The analyze dispatcher now scopes its kinds in tiers: kinds built on the scoped-node accessors inherit scoping automatically; dead_code, hotspots, cycles, and releases are per-row filtered via analyzeNodeVisible; the edge / algorithm / framework and file / AST-scan kinds were taught to honor the resolved scope. Kinds that read the full graph directly (community / git-mining / per-id / synthesizer) are not narrowed in v1 and now stamp a `scope_note` saying their results may span the whole index.

## Tests & docs

- New tests: scope_resolve_test.go, analyze_scope_test.go, workspace_isolation_test.go, scope_allows_test.go, plus search_text and field_query cases. `go build ./...` and `go test -race ./internal/mcp/...` pass.
- docs/multi-repo.md documents the defaults and the narrow-only invariant; docs/proposals/scope-intent-defaults.md records the design and the default-ON decision.

## Summary

Brief description of what this PR does.

## Changes

- 

## Testing

## Checklist


docs/proposals/scope-intent-defaults.md
```diff
@@ -0,0 +1,95 @@
+# Proposal: intent-based default scoping for multi-repo workspaces
```
zzet: @pacaya, the docs directory should contain up-to-date documentation and shouldn't be mixed with any specs. The main problem I see is a quick drift of this kind of specs' state from the real functionality. While I see it as important to keep the documentation up to date, I'm not ready to spend time on actualisation specs condition. Let's keep the scope of this file for documentation purposes (or remove it and add a mention of critical aspects in another file where it is necessary)

zzet:
Hey @pacaya! Thank you for your contribution and your time. 
The scope of changes seems bigger than I imagined, and it will require additional time to review the change.

docs/proposals/scope-intent-defaults.md
pacaya: Thanks @zzet  - that's a fair call, and I agree the docs/ tree shouldn't carry specs that drift out of sync. I've gone with your second suggestion: removed docs/proposals/scope-intent-defaults.md entirely.

The critical user-facing aspects were already captured in docs/multi-repo.md - the scope.intent_defaults flag, the intent table (locate -> repo, reach/analyze -> workspace), the opt-out, and the upgrade/behavior-change note all live there and stay maintained. I also removed the now-dangling cross-reference link that pointed back to the proposal. The design rationale and discussion stay preserved in the merged commit and this PR thread, which feels like the right home for it.

Net effect: multi-repo.md is the single source of truth for the behavior, and there's no proposals/spec directory to keep actualised going forward.

PR #205, opened 2026-06-30
feat(agents): add a Pi coding-agent adapter

vitaliyslion:
## Summary

This is a **proposal** for first-class [Pi](https://github.com/earendil-works/pi) support, and it deliberately departs from how every other adapter in this repo works — so it's worth reviewing as an RFC rather than a routine adapter addition.

Every existing adapter configures an already-MCP-capable client and writes instruction/markdown files. **Pi has no MCP support by design**, so the only way to expose Gortex's graph tools and read-discipline to it is to ship an executable **Pi extension written in TypeScript**. That is a new kind of artifact for this codebase — *code we now have to maintain*, pinned against Pi's pre-1.0 extension API.

To keep that surface as small as possible this is a **minimal implementation**:
- an **MCP-equivalent bridge** — the extension registers Gortex's graph tools natively (each shelling `gortex call`) and bridges Pi lifecycle events to `gortex hook --agent=pi`, so the Go side keeps owning all posture logic;
- the **read-discipline orientation** injected into the model's context (as a tail user message, cache-safe);
- the **communities routing block** merged into `AGENTS.md` (with `--skills`).

The extension is intentionally a single self-contained file with **zero runtime dependencies** (Node built-ins only), written verbatim by the adapter — no `package.json` / `node_modules`.

**Proposed follow-up:** Pi has a [packages system](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/packages.md). Publishing Gortex as a proper Pi package in the near future would let users install it the idiomatic Pi way, and let us version and distribute the extension instead of writing a file into `.pi/`. This PR is the groundwork; packaging is the natural next step.

## Changes

- New `pi` adapter (`internal/agents/pi/`) — detection (`.pi/` dir or `pi` on `PATH`), project + global install modes, renders an embedded `extension/index.ts` with templated sentinels (binary path, hook argv, enforce flag, tools preset).
- Native graph-tool registration in the extension: discovers the eager `core` preset via `gortex tools list`, plus a `gortex_tools_search` meta-tool that mirrors MCP's tools_search + promotion for the deferred catalogue; re-registers per session (Pi resets the registry on `/new`) and brings up the shared daemon (`ensureDaemon`).
- Read-discipline enforcement bridged to Go: new `internal/hooks/pi.go` defines the `PiEvent` / `PiDecision` wire contract and reuses the shared `enrich` + `applyMode` pipeline; supports the deny / enrich / consult-unlock / nudge postures and `--no-hooks` (orientation + tools still wired).
- Orientation briefing injected via Pi's `context` hook as a tail user message (avoids a `systemPrompt` mutation that would bust prompt caching).
- `AGENTS.md` community-routing block on `--skills` (mirrors the opencode adapter); read-discipline rules ride the extension at runtime rather than an instructions file.
- Shared `markGraphConsulted` helper extracted in `internal/hooks/` so the consult-unlock handshake stays identical across the Claude, Hermes, and Pi handlers (also touches `pretooluse.go`, `hermes.go`).
- Wiring + docs: `gortex hook --agent=pi` dispatch (`cmd/gortex/hook.go`), `init` / `init_wizard` registration, `docs/agents.md` + `README.md`, and an `agent-render` golden (`testdata/agent-render/pi.txt`).

## Testing

## Checklist


zzet:
@vitaliyslion thank you for your contribution!

Issue #206, opened 2026-06-30
feat: Add Codex PreToolUse/PostToolUse hooks for graph context injection


Maple0517:
## Summary

Gortex currently implements rich hooks for Claude Code (PreToolUse, PostToolUse, PreCompact, Stop, SessionStart) but has **no hooks for Codex CLI**. The Codex adapter only writes MCP config and AGENTS.md communities block, missing the opportunity to proactively inject graph context during tool calls.

## Problem

Codex CLI now supports lifecycle hooks (PreToolUse, PostToolUse, SessionStart) similar to Claude Code. Without hooks, Codex users experience:

1. **Deferred tool discovery**: Codex's `tool_search` may omit expected MCP tools on first query (tracked upstream as openai/codex#21503)
2. **No graph context injection**: When Codex runs `Grep`, `Glob`, or `Bash`, it doesn't get enriched with graph context
3. **No stale index detection**: After `git commit`, Codex doesn't get notified to reindex
4. **Inferior experience**: Claude Code gets proactive graph integration; Codex gets passive MCP-only access

### What Claude Code gets (from `internal/agents/claudecode/hooks.go`)

| Hook | Behavior |
|------|----------|
| PreToolUse | Enrich Read/Grep/Glob/Bash with graph context, deny mode available |
| PostToolUse | Augment tool output with graph context (enrich mode) |
| PreCompact | Inject orientation snapshot before context compaction |
| Stop | Post-task diagnostics (detect_changes, check_guards, dead_code) |
| SessionStart | Prime first turn with graph orientation |

### What Codex gets today (from `internal/agents/codex/adapter.go`)

| Integration | Status |
|-------------|--------|
| MCP config | ✅ `~/.codex/config.toml` |
| AGENTS.md | ✅ Communities block |
| PreToolUse | ❌ Missing |
| PostToolUse | ❌ Missing |
| SessionStart | ❌ Missing |

## Proposed Solution

Add Codex hooks support to the Codex adapter, following the same patterns as Claude Code hooks:

### 1. SessionStart hook (low-risk, high-value)

Inject a one-line reminder to use Gortex graph tools first:

```toml
# ~/.codex/config.toml
[[hooks.SessionStart]]
matcher = "startup|resume|clear|compact"

[[hooks.SessionStart.hooks]]
type = "command"
command = "echo 'Prefer gortex MCP tools (search_symbols, find_usages, analyze, get_callers) over Read/Grep/Glob for indexed source code.'"
```

[...]

PR #207, opened 2026-06-30
feat(codex): install SessionStart graph orientation hook

blankanswer:
## Summary

  Adds minimal Codex hook support by installing a `SessionStart` orientation hook into `~/.codex/config.toml`.

  When Codex starts or resumes a session, the hook prints a short reminder to prefer Gortex graph tools before large direct file reads. This mirrors the existing agent onboarding intent without changing MCP
  behavior or adding tool interception.

  This intentionally keeps the first Codex hook PR small: it only covers `SessionStart`, not `PreToolUse` / `PostToolUse`.

  ## Changes

  - Install an idempotent Codex `SessionStart` hook when hooks are enabled
  - Match Codex start sources with `startup|resume|clear|compact`
  - Preserve existing Codex hooks and existing `mcp_servers.gortex` config
  - Respect `--no-hooks` / `InstallHooks=false` for Codex hook installation
  - Add POSIX and Windows hook commands
  - Update install/init copy and docs for supported agent hooks
  - Add focused tests for fresh install, repeated install, existing config merge, and no-hooks behavior

  Out of scope:

  - Codex `PreToolUse` / `PostToolUse` hooks
  - Dynamic hook handling or context injection
  - Changes to Claude Code hook behavior
  - Changes to MCP tools, indexing, daemon, LSP, resolver, or graph storage
  - Expanding `--hooks-only` beyond its current Claude Code behavior

  ## Testing

  - [x] `go test ./internal/agents/...`
  - [x] `go test ./cmd/gortex/...`
  - [x] `go test -race ./internal/agents/codex ./internal/agents`
  - [x] `git diff --check`
  - [x] `go test -run '^$' -bench . ./internal/agents/...`
  - [x] Benchmarks run if performance-relevant

  Benchmarks are not performance-relevant for this config-writing change; the scoped benchmark build passed and there are no benchmark cases in the changed agent packages.

  ## Checklist

Related to https://github.com/zzet/gortex/issues/206

### DIFF
--- a/internal/agents/codex/adapter.go
+++ b/internal/agents/codex/adapter.go
@@ -26,7 +26,7 @@ const Name = "codex"
 const DocsURL = "https://developers.openai.com/codex/mcp"
 
 const codexSessionStartMatcher = "startup|resume|clear|compact"
-const codexSessionStartMessage = "[Gortex] Prefer Gortex graph tools (search_symbols, get_callers, get_file_summary) before Read/Grep/Glob over large files."
+const codexSessionStartMessage = "IMPORTANT: Prefer Gortex MCP tools (search_symbols, get_callers, get_file_summary, edit_file) over Read/Grep/Glob/Edit."
 const codexSessionStartCommand = "printf '%s\\n' '" + codexSessionStartMessage + "'"
 const codexSessionStartWindowsCommand = "powershell -NoProfile -Command \"Write-Output '" + codexSessionStartMessage + "'\""
 
--- a/internal/agents/codex/adapter.go
+++ b/internal/agents/codex/adapter.go
@@ -25,6 +25,11 @@ import (
 const Name = "codex"
 const DocsURL = "https://developers.openai.com/codex/mcp"
 
+const codexSessionStartMatcher = "startup|resume|clear|compact"
+const codexSessionStartMessage = "[Gortex] Prefer Gortex graph tools (search_symbols, get_callers, get_file_summary) before Read/Grep/Glob over large files."
+const codexSessionStartCommand = "printf '%s\\n' '" + codexSessionStartMessage + "'"
+const codexSessionStartWindowsCommand = "powershell -NoProfile -Command \"Write-Output '" + codexSessionStartMessage + "'\""
+
 type Adapter struct{}
 
 func New() *Adapter                { return &Adapter{} }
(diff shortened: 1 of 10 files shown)


internal/agents/codex/adapter.go
```diff
@@ -25,6 +25,11 @@ import (
 const Name = "codex"
 const DocsURL = "https://developers.openai.com/codex/mcp"
 
+const codexSessionStartMatcher = "startup|resume|clear|compact"
+const codexSessionStartMessage = "[Gortex] Prefer Gortex graph tools (search_symbols, get_callers, get_file_summary) before Read/Grep/Glob over large files."
```
zzet: I would love to get feedback from Codex users, but intuitively:
1. I see zero value for duplication Gortex with `[Gortex] `
2. I don't think`before Read/Grep/Glob over large files.` will be a clear instruction:
- Why before?
- Why only large files?

```suggestion
const codexSessionStartMessage = "IMPORTANT: Prefer Gortex graph tools (search_symbols, get_callers, get_file_summary) over Read/Grep/Glob."
```

One more point: tools like `edit_file` can be very useful to mention, too. Overall, the initial message makes sense to test.

blankanswer: Yep, good point.  I agree the original wording was too fuzzy: “before” made it sound like an ordering rule, and “large files” narrowed the guidance unnecessarily; I updated the message to remove the duplicate label, use “over” instead of “before”, drop the large-file qualifier, and mention `edit_file` as well: `IMPORTANT: Prefer Gortex MCP tools (search_symbols, get_callers, get_file_summary, edit_file) over Read/Grep/Glob/Edit. `This keeps the PR scoped to the lightweight SessionStart orientation hook; PreToolUse/PostToolUse behavior remains out of scope for a follow-up PR~

zzet (approved):

zzet:
@blankanswer merged. Thank you for contribution. Follow up question if you're planning to cover other hooks, are you?

blankanswer:
> @blankanswer merged. Thank you for contribution. Follow up question if you're planning to cover other hooks, are you?已合并。感谢您的贡献。如果您打算涵盖其他钩子函数，请告知我们。

Yep, I’m happy to continue working on the remaining Codex hook surface, but I’d prefer to split it into small follow-up PRs rather than trying to reach full Claude Code parity in one large change.


My next step will be to validate the Codex PreToolUse and PostToolUse input/output contract against the current Codex CLI behavior. After that, I plan to open a lightweight PR for PreToolUse only, focused on soft nudging / additional context rather than deny or enforcement behavior. I’ll keep PostToolUse enrichment, PreCompact/Stop-style lifecycle handling if Codex supports equivalent events, and any broader hook dispatcher work as separate follow-up PRs~

Issue #220, opened 2026-07-02
clangd enrichment can trigger clang-tidy crashes and high CPU


darthghandi:
## Summary
Gortex auto-registers `clangd` for C/C++ repos and currently launches it with `--background-index --header-insertion=never`. On a large C/C++ repo, semantic enrichment drives a very large LSP workload through clangd. If the repo has a broad `.clang-tidy` (`bugprone-*`, `readability-*`, `clang-analyzer-*`, `performance-*`, etc.), clangd 22.1.6 can repeatedly crash while running clang-tidy matchers. Gortex then reconnects and repeats the work, keeping clangd at high CPU.

## Evidence
A daemon log showed:
- `semantic enrichment starting` for `lsp-clangd`
- clangd argv: `--background-index --header-insertion=never`
- repeated crash stack dumps in clang-tidy checks:
  - `bugprone-pointer-arithmetic-on-polymorphic-object`
  - `bugprone-not-null-terminated-result`
  - `performance-unnecessary-copy-initialization`
- one completed clangd enrichment run took ~113 minutes and included tens of thousands of LSP requests.

## Proposed fix
For gortex semantic enrichment, default clangd should avoid lint diagnostics and background indexing, e.g. `--clang-tidy=false --background-index=false --header-insertion=never -j=1`. Gortex needs semantic graph signal, not clang-tidy diagnostics, and should avoid reconnect loops amplifying repeated clangd crashes.

PR #227, opened 2026-07-03
Use safer clangd defaults for enrichment

darthghandi:
## Summary

- Change the built-in clangd enrichment defaults to disable background indexing and clang-tidy, keep header insertion disabled, and limit clangd to one worker.
- Add a regression test that pins the default clangd argument list.
- Document the new defaults and how users can override them with `semantic.providers[].args`.

## Test Plan

- [x] `GOCACHE=/tmp/gortex-go-cache GOMODCACHE=/tmp/gortex-gomod-cache go test ./internal/semantic/lsp -count=1`
- [x] Manual llama.cpp diagnostic run confirmed gortex launched clangd with `--background-index=false --clang-tidy=false --header-insertion=never -j=1`.
- [ ] `go test ./...` currently fails on both this branch and main in `internal/parser/forest` due to `github.com/alexaandru/go-sitter-forest/perl` C compilation error: `bsearch.h:72:7: error: expected identifier or ‘(’ before ‘_Generic’`.

Related: #220


internal/semantic/lsp/registry.go
```diff
@@ -319,16 +319,18 @@ var Servers = []ServerSpec{
...
-		// server at high CPU for the whole pass.
-		Args:      []string{"--background-index", "--header-insertion=never", "--clang-tidy=false"},
+		// Gortex uses clangd for request-driven graph evidence during
+		// enrichment, not lint diagnostics or a persistent project index.
+		// Keep clang-tidy and background indexing off by default so broad
+		// repo .clang-tidy configs and large C++ trees do not dominate CPU
+		// or crash-loop the enrichment subprocess. Users can opt back in via
+		// a semantic.providers override in .gortex.yaml.
+		Args: []string{
+			"--background-index=false",
```
zzet, line 329: @darthghandi `--background-index=false` has caveats:

It can impact the C/C++ cross-file graph fidelity:
  - find_usages / get_callers on a C/C++ function -> only same-TU callers; call sites in other .cpp files are not found. Ambiguous call edges never get confirmed to the lsp_resolved tier, so min_tier filtering drops them → missing caller edges.
  - find_implementations -> misses implementers in other files.
  - Type hierarchy ->  misses subtypes defined elsewhere.

I want to benchmark it before accepting it.

PR #283, opened 2026-07-10
instructions: switchable session profiles — one table drives body, tool preset, skills, hook tier

zzet:
## What

Agent sessions pay an ambient prefix on every API call (tool schemas + installed guidance + hook output). This PR makes that surface a first-class, switchable profile instead of a fixed cost:

- **`internal/profiles`**: one table defines each profile's instruction body, MCP tool preset, skills subset, and hook-verbosity tier — all four surfaces generate from the same source so they cannot drift (a no-drift test asserts every eagerly-listed tool is cued in the rendered body).
- **`gortex instructions list|show|switch|regen`**: switching atomically copies the profile body over \`~/.gortex/instructions/active.md\` (never a symlink; asserted). The installed CLAUDE.md section becomes a thin \`@\`-include pointer block — no commented-out alternatives, since inactive text still bills every turn. The switch output states explicitly that changes take effect for new sessions only.
- **\`localization\` profile**: a 13-tool task-focused surface (19.1KB cold tools/list vs the 25.9KB agent preset; 2.0KB body) with lean hook output. The default profile remains \`core\` — machines that never switch see zero behavior change. Preset precedence: forwarded spec > operator pin > profile > client default > server default.
- Per-profile byte-ceiling tests; migration gate rewrites the legacy fat block to the pointer on re-install; skills sync prunes only byte-identical shipped copies.

## Measured (localization benchmark, 28 real tasks)

The lean profile delivered exactly its byte math — per-turn tokens 25.0k → 22.0k — and adoption held at 100% (9/9 spot-check sessions, first-tool 100%). The first lean surface traded the savings back by dropping \`batch_symbols\` (multi-symbol follow-ups cost a turn each; median turns 5.5 → 6.5) — that tool is now part of the surface, and the lesson is recorded in the docs: an ambient diet is bounded by turn count, so every focused profile must keep a multi-symbol read.

## Notes for review

- Positioning cues and the memory-workflow triggers survive in every profile body (they have no other home).
- Non-Claude adapters keep their existing full-body writes; profile-izing those surfaces is follow-up scope.
- After the explore-verb PR merges, the localization profile gains that verb (opener text, eager set, ceiling 2048→2304) — the rebase is mechanical and documented in the session notes.

### DIFF
--- a/cmd/gortex/instructions.go
+++ b/cmd/gortex/instructions.go
@@ -1,0 +1,153 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/zzet/gortex/internal/agents"
+	"github.com/zzet/gortex/internal/agents/claudecode"
+	"github.com/zzet/gortex/internal/profiles"
+)
+
+// instructions.go is the `gortex instructions` command tree — the CLI
+// front end for instruction profiles (internal/profiles): named
+// bundles of instructions body + MCP tool preset + skills subset +
+// hook-verbosity tier, generated from one table. `switch` atomically
+// repoints the @-included active.md copy; agents pick the change up at
+// their next session start.
+
+var instructionsCmd = &cobra.Command{
+	Use:   "instructions",
+	Short: "Manage instruction profiles (guidance depth per machine)",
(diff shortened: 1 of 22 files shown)


zzet:
Consolidated into #282 (the profiles commits ship there together with the explore verb they configure).

PR #292, opened 2026-07-11
fix(upgrade): run script installer pipeline via shell

### LINKED ISSUE #281: gortex upgrade --run fails for script method: pipeline exec'd as raw argv (curl: could not resolve host '|')
## Summary

`gortex upgrade --run` fails for the **script** install method because it executes the shell pipeline `curl -fsSL https://get.gortex.dev | sh` as a raw argv vector instead of through a shell. The `|` and `sh` tokens are handed to `curl` as additional hostnames, so curl tries to resolve them as hosts and exits `6`.

## Version

```
gortex v0.58.3+447588b   (also present in the v0.60.0 codebase — verified by inspecting the printed command)
os/arch: darwin/arm64
```

## Repro

```console
$ gortex upgrade
Upgrading gortex 0.58.3 → v0.60.0 (install method: script)
Run:
  curl -fsSL https://get.gortex.dev | sh

$ gortex upgrade --run
...
curl: (6) Could not resolve host: |
curl: (6) Could not resolve host: sh
Error: upgrade command failed: exit status 6
```

## Diagnosis

The printed command is a shell pipeline (`curl ... | sh`), but `--run` appears to `exec` the argv list `["curl","-fsSL","https://get.gortex.dev","|","sh"]` directly (e.g. `exec.Command("curl", args...)`), so `|` and `sh` become curl positional args → treated as hostnames. No shell ever interprets the pipe.

## Expected

`--run` should run pipeline-containing commands through a shell, e.g.:

```go
exec.Command("sh", "-c", cmdString)
```

for the script method (and any method whose command contains shell metacharacters). Non-pipeline methods (brew/scoop/go install) are unaffected because they're plain argv.

## Workaround

Run the printed command manually, or install the tarball by hand (download release asset + `checksums.txt`, verify sha256, extract to `$GORTEX_INSTALL_DIR`). Both preserve the documented verification.

zerone0x:
## Summary

Fixes #281

`gortex upgrade --run` now executes the installer-script command through `sh -c`, so the documented installer pipeline is interpreted by a shell instead of being split into raw curl arguments.

## Changes

- Add `upgradeExecCommand` to choose shell execution only for commands with shell metacharacters
- Keep plain argv execution for commands such as `go install ...`
- Add regression coverage for the script pipeline and plain `go install` path

## Validation

- `gofmt -w cmd/gortex/upgrade.go cmd/gortex/upgrade_test.go` — blocked: this runner does not have `gofmt`/Go installed
- Static diff review only in this environment; change is limited to fixed upgrade command templates and tests

### DIFF
--- a/cmd/gortex/upgrade.go
+++ b/cmd/gortex/upgrade.go
@@ -196,9 +196,8 @@ func runUpgrade(cmd *cobra.Command, args []string) error {
 	if !upgradeRun {
 		fmt.Fprintf(out, "Run:\n  %s\n", command)
 	} else {
-		parts := strings.Fields(command)
 		fmt.Fprintf(out, "$ %s\n", command)
-		ex := exec.CommandContext(cmd.Context(), parts[0], parts[1:]...) //nolint:gosec // command is one of the fixed install-method templates
+		ex := upgradeExecCommand(cmd, command)
 		ex.Stdout, ex.Stderr, ex.Stdin = out, cmd.ErrOrStderr(), os.Stdin
 		if rerr := ex.Run(); rerr != nil {
 			return fmt.Errorf("upgrade command failed: %w", rerr)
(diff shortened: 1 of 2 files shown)


zzet (dismissed):

zzet:
Thanks for this @zerone0x — your diagnosis and the `sh -c` approach were spot on. I've carried your intent and test forward in #316 (you're credited as co-author on the commit).

Two small corrections were needed relative to this PR's test, which its notes mention was never run (no Go in that environment):

- It asserted on `cmd.Path`, but `exec.Command` resolves a bare name via `LookPath`, so `Path` becomes `/bin/sh` and `/usr/local/go/bin/go` once those are on `PATH` — the assertion would fail in CI. #316 asserts on `cmd.Args[0]` instead.
- It passed `rootCmd`, whose `Context()` is `nil` outside cobra's `Execute()`, so `exec.CommandContext(nil, …)` panics. #316 has the helper take a `context.Context` directly, which removes that foot-gun and keeps it unit-testable.

#316 supersedes this PR — closing in favour of it. Really appreciate the fix. 🙏

PR #293, opened 2026-07-12
Retry AtomicWriteFile's rename past transient Windows sharing violations

yimingll:
## Summary

On Windows, `os.Rename` — which `AtomicWriteFile` uses for its temp→dest replace — maps to `MoveFileEx(MOVEFILE_REPLACE_EXISTING)`, which fails with `ERROR_SHARING_VIOLATION` / `ERROR_ACCESS_DENIED` when the destination is briefly held open **without `FILE_SHARE_DELETE`** by another process: an editor's language server, antivirus, a search indexer, or **Gortex's own file watcher re-indexing the file we just wrote**.

Because `AtomicWriteFile` is the single choke point for every MCP file mutation (`write_file`, `edit_file`, `move`, safe-delete, overlay-branch), this surfaces to coding agents as spurious errors like:

```
rename ...\entry.cs.gortex.tmp-3024971810 -> ...\entry.cs: The process cannot access the file because it is being used by another process.
```

on rapid successive edits to the same file — the second write races the reindex the first write triggered. Retrying the same edit typically succeeds, confirming a transient lock rather than a real error.

POSIX renames replace an open destination atomically, so this never happens off Windows.

## Changes

- `internal/agents/writer.go`: route the temp→dest rename through a new `renameWithRetry` — a bounded retry (10 attempts, ~225 ms total backoff) that retries **only** when the failure is a transient Windows sharing/lock violation and returns every other error immediately.
- `internal/agents/writer_windows.go` (new, `//go:build windows`): `isRetryableRenameErr` matching `ERROR_SHARING_VIOLATION` (32), `ERROR_LOCK_VIOLATION` (33), `ERROR_ACCESS_DENIED` (5) via `errors.Is` on the `*os.LinkError` chain.
- `internal/agents/writer_other.go` (new, `//go:build !windows`): the predicate is always `false`, so the loop collapses to a single `os.Rename` with no added latency on POSIX.

The build-tag split mirrors the vendored `renameio` package's existing Windows/non-Windows layout. No behavior change off Windows; on Windows the only change is that a transient sharing violation is retried instead of surfaced.

## Testing

- [x] New `internal/agents/writer_windows_test.go` reproduces the exact failure: it opens the destination without `FILE_SHARE_DELETE`, then asserts `AtomicWriteFile` retries and succeeds once the handle releases. A second test bounds the retry (destination held throughout → gives up and returns the error, leaving the original file intact).
- [x] Portable `writer_test.go` coverage: create/overwrite happy path (asserts no leftover `*.gortex.tmp-*`) and non-retryable errors fast-fail.
- [x] `go test ./internal/agents/` passes on Windows (all four new tests); `gofmt -l` and `go vet` clean; `GOOS=linux CGO_ENABLED=0 go build ./internal/agents/` succeeds.

> Note: the `test` job runs on ubuntu/macos only and the `build-windows` job is build-only (per `.github/workflows/ci.yml`), so the Windows regression tests are validated locally / on-demand rather than in the CI matrix.

## Checklist

- [ ] Language extractor includes `Meta["methods"]` for interfaces (n/a)
- [ ] Methods have `EdgeMemberOf` edges to their containing type (n/a)

### DIFF
--- a/internal/agents/writer.go
+++ b/internal/agents/writer.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"time"
 )
 
 // writer.go centralises every write `gortex init` performs. Going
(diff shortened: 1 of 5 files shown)


zzet (approved):

zzet:
@yimingll thank you for fixing this issue.

PR #294, opened 2026-07-13
Windows: fix slash-path duplicate indexing and orphaned atomic-write temp files

yimingll:
## Windows robustness: fix slash-path duplicate indexing and orphaned temp files

Two self-defects surfaced while hardening the Windows atomic-write path, plus the rename-retry that this branch already carried.

### 1. Slash-path duplicate indexing (Windows)

The graph keys a file's nodes under **OS-native separators** — the cold bulk walk stamps `repo/pkg\sub\x.go` on Windows (`filepath.Rel`, no `ToSlash`), and every runtime file→node lookup produces the same form. But the **incremental** re-index path derived its graph lookup/eviction key with `relKey`, which slash-normalises to `repo/pkg/sub/x.go`.

`graph.GetFileNodes` / `graph.EvictFile` match the key **byte-for-byte with no separator folding**, so on Windows:

- an incremental re-index **missed** the backslash-keyed cold nodes → the evict was a no-op → the re-parse appended a **second, forward-slash-keyed copy**. A duplicate node set grew on every save.
- deleted files leaked their nodes the same way (the deletion loops built the evict key from the slash-form mtime map).
- the watcher's change-detection reported **every symbol as added/removed** (its `GetFileNodes` used the slash key too).

On POSIX the two key forms coincide (`filepath.Rel` already yields `/`), so this only ever bit Windows — which is why it hid behind a green Linux CI.

**Fix:** introduce `graphRelKey` (OS-native separators + NFC) as the single graph-node key, and keep `relKey` (slash + NFC) for the `fileMtimes` map only. Every graph-key derivation now routes through `graphRelKey`: `indexFile` (evict + new-node form), `EvictFile`, `ReresolveFileScoped`, `graphFilePaths`, the scoped-deletion loops (via `filepath.FromSlash` on the mtime key), the cold-walk size/content skip nodes, and the watcher's `GetFileNodes` / `snapshotSymbols`. The mtime map, `IsStale`, and staleness detection are deliberately untouched.

This also fixes the pre-existing `TestIncrementalReindexPaths_EvictsDeletedFileInScope` and `TestIncrementalReindex_ConvergesToFullIndex` failures on Windows. New regression guard: `TestIncrementalReindex_NestedFileNoDuplicate` (verified to fail without the fix).

### 2. Orphaned atomic-write temp files never cleaned up

`AtomicWriteFile` writes through a `<name>.gortex.tmp-<rand>` temp file and renames it into place. A crash (or kill) between `CreateTemp` and the rename strands that temp on disk **forever** — the success path only removes its own temp — so orphans accumulate wherever Gortex writes agent config. The doc comment literally told users to "clean up manually".

**Fix:** every `AtomicWriteFile` now calls `cleanStaleTempFiles(dir)` first, reaping files that carry the `.gortex.tmp-` infix and have been untouched for over an hour. The age gate sits far above a single write's lifetime (data write + `renameWithRetry`'s ~225 ms worst case), so a temp another writer is still filling in is never touched. Best-effort and error-swallowing — a write never fails because a leftover couldn't be removed. New test: `TestAtomicWriteFileReapsStaleTempOrphans`.

### Verification

- `go build ./...` and `go vet` on the touched packages pass.
- `internal/indexer` full package: **zero new test failures** vs. the base branch (compared by test name), and the two reindex failures above now pass. The remaining failures reproduce on an unmodified checkout (pre-existing Windows-environment issues, unrelated to this change).
- Both regression tests were confirmed to fail when the fix is reverted.

### DIFF
--- a/internal/agents/writer_test.go
+++ b/internal/agents/writer_test.go
@@ -331,7 +331,7 @@ func TestRenameWithRetryReturnsNonRetryableErr(t *testing.T) {
 }
 
 // TestAtomicWriteFileReapsStaleTempOrphans guards the cleanup of temp
-// files a crashed write orphaned ("历史临时文件未清理"): a *.gortex.tmp-*
+// files a crashed write orphaned ("not cleared"): a *.gortex.tmp-*
 // stranded when a process died between CreateTemp and the rename must be
 // reaped by a later AtomicWriteFile — but a temp still in flight (fresh
 // mtime) and any unrelated file must be left untouched.
--- a/internal/agents/writer_test.go
+++ b/internal/agents/writer_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 )
 
 // TestWriteIfNotExistsCreatesAndSkips covers both the create and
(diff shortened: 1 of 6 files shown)


internal/agents/writer_test.go
```diff
@@ -328,3 +329,53 @@ func TestRenameWithRetryReturnsNonRetryableErr(t *testing.T) {
 		t.Fatal("expected an error renaming a missing source, got nil")
 	}
 }
+
+// TestAtomicWriteFileReapsStaleTempOrphans guards the cleanup of temp
+// files a crashed write orphaned ("历史临时文件未清理"): a *.gortex.tmp-*
```
zzet:
```suggestion
// files a crashed write orphaned ("not cleared"): a *.gortex.tmp-*
```

internal/indexer/slash_path_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/zzet/gortex/internal/graph"
+)
+
+// TestIncrementalReindex_NestedFileNoDuplicate guards the
+// "斜杠路径重复索引" (slash-path duplicate indexing) regression on Windows.
```
zzet:
```suggestion
// slash-path duplicate indexing regression on Windows.
```

yimingll:
Done — pushed 65e3065, dropping the non-English text from both test doc comments. Thanks for the review!

By the way, I really love Gortex — it's become my daily driver as an MCP server. Thanks for building it. 🙏

zzet (approved):

zzet:
Thank you @yimingll. I'll take another look at the problem of orphan files later (I don't 100% like the solution for checking files for the past hour only, but it's better than nothing).

PR #337, opened 2026-07-25
fix(init): refuse to index the filesystem root or $HOME

vitaliyslion:
## Summary

`gortex init` had no guard against being pointed at a dangerous root — running it against `$HOME` (or `/`) walks the entire tree by default, since `--skills` defaults to on. `gortex track` already refused these paths via `indexer.UnsafeIndexRootReason`; `init` never called it. This wires the same guard into `init`, checked before the wizard, any indexing, or any file writes.

## Changes

- `cmd/gortex/init.go`: `runInit` now resolves `absRoot` immediately after parsing the `[path]` arg and, unless `--force` is set, calls `indexer.UnsafeIndexRootReason(absRoot)` before doing anything else (wizard, indexing, or writes). Refuses with the reason plus `; pass --force to init it anyway` when the target is the filesystem root, a Windows drive root, or exactly `$HOME`. Removed the now-redundant later `filepath.Abs(root)` recomputation. Updated the `--force` flag help text to mention the new bypass.
- `cmd/gortex/init_integration_test.go`: added `TestInitRefusesHomeDirectory` (asserts the refusal, its error message, and that no `.gortex/` marker is written) and `TestInitForceBypassesHomeDirectoryGuard` (asserts `--force` still proceeds).

## Testing

- [x] New tests added for new functionality (`TestInitRefusesHomeDirectory`, `TestInitForceBypassesHomeDirectoryGuard`)
- [x] `go test ./cmd/gortex/... -run TestInit` — all init/init-wizard tests pass
- [x] `go build ./cmd/gortex/...`
- [x] Manually verified with a built binary: `gortex init "$HOME" --yes` refuses with the home-directory message; `gortex init "$HOME" --yes --force` proceeds
- [ ] Full `go test -race ./...` not run (scoped to the touched package; happy to run the full suite before merge if wanted)
- [ ] Benchmarks — not performance-relevant

## Checklist

- [x] Code follows existing patterns in the codebase (mirrors the `unsafeRootBlocked`/`pass force to track it anyway` convention already used by `gortex track`)
- [ ] Language extractor includes `Meta["methods"]` for interfaces (not applicable)
- [ ] Methods have `EdgeMemberOf` edges to their containing type (not applicable)

### DIFF
--- a/cmd/gortex/init.go
+++ b/cmd/gortex/init.go
@@ -108,7 +108,7 @@ func init() {
 	initCmd.Flags().BoolVar(&initJSON, "json", false, "emit a structured JSON report on stdout")
 	initCmd.Flags().BoolVar(&initDryRun, "dry-run", false, "plan writes without modifying disk")
 	initCmd.Flags().BoolVar(&initDryRunIntake, "dry-run-intake", false, "emit a privacy-safe corpus intake manifest and exit before parsing or writing")
-	initCmd.Flags().BoolVar(&initForce, "force", false, "overwrite keys we would otherwise preserve during a merge")
+	initCmd.Flags().BoolVar(&initForce, "force", false, "overwrite keys we would otherwise preserve during a merge; also bypasses the filesystem-root/home-directory safety check")
 
 	rootCmd.AddCommand(initCmd)
 }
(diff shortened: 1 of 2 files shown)


zzet:
@vitaliyslion can you please check the failed test?

UPD: Looks like a flaky test

PR #343, opened 2026-07-25
fix(mcp): run graph analysis on demand instead of eagerly

zzet:
Stacked on #340 — review that first; this branch targets it, not `main`.

## Problem

The analysis pass (six whole-graph analyzers) took **79s–366s** on a 20-repo workspace and ran twice over: once at daemon startup, then again after *every* graph change (`WatchForReanalysis`, 500ms debounce, `// any event triggers reanalysis`).

That was self-defeating. The pass is *triggered by* a graph mutation, then runs for minutes while more mutations land, so it can never be published against a stable revision. Combined with the validation bugs in #340, **no generation had ever activated** — `analysis_active_generation` was empty across the daemon's entire history, `cache_hit:false` on every pass. The daemon recomputed the same snapshot on every start and discarded it, while the ~400 MB result stayed resident.

## Change

Analysis starts when a consumer first needs it. Nothing is lost by dropping the eager pass: the currency token is derived from graph state, so a mutation already makes the snapshot non-current — the watcher only has to stop recomputing.

Since the pass far exceeds any agent's tool budget, a consumer that finds the snapshot missing **does not block**. It starts the pass in the background and returns the collection empty plus `retry_after_seconds`, derived from the previous pass's runtime and clamped so a hint is neither a busy-poll nor longer than a plausible budget:

```json
{
  "communities": [],
  "status": "analysis_pending",
  "retry_after_seconds": 30,
  "message": "graph analysis is running in the background — retry this call in ~30s. ..."
}
```

The empty collection keeps the response shape identical to success, so a caller that just iterates sees "none yet" rather than a decode error. This replaces `"run index_repository first"`, which was misleading against a fully populated graph.

**Two lock fixes make it safe.** `ensureAnalysis` uses `TryRLock` — a plain `RLock` queues behind the running pass and blocks for minutes, the exact failure it exists to prevent. And `releaseTransientAnalysisIfIdle`, which runs on the tail of **every** tool call, now uses `TryLock`: taking it unconditionally stalled every request — including ones that never touch analysis — behind a background pass.

## Measured on a live 20-repo daemon

| | Before | After |
|---|---|---|
| Time to `ready` | 2m21s – 11m29s | **21s** |
| `tool not found` reply during a pass | 66s | **3.5s** |
| `search_symbols` during a pass | — | **3.5s** |
| Active generation | never, in the daemon's whole history | **generation 10, state=ready** |
| Analysis pass on restart | `cache_hit:false`, 79–366s | **`cache_hit:true`, 0.19s** |
| `daemon status memory` | 507 MiB | **135.6 MiB** |
| Retained heap (pprof `?gc=1`) | 497 MB | **141 MB** |

The analyzer structures are now absent from the heap profile entirely — `columnText` 186 MB, Leiden 128.67 MB, adjacency 98.85 MB, HITS 86.64 MB, PageRank 82.37 MB all gone, served from SQLite instead. `get_communities` returns all 3,251 communities from the durable generation.

Note this needed no new residency mechanism: `releaseTransientAnalysisIfIdle` already drops the seven in-memory views once a generation is published. It simply never fired, because publication never succeeded.

## Deliberately not included

I also tried removing the `DELETE FROM analysis_active_generation` in `invalidateAnalysisGenerationTx`, on the theory that marking the generation stale was sufficient. **It is not, and the tests caught it:** `LoadActiveAnalysisHeader` treats an active pointer aimed at a non-ready generation as *corruption*, not as a cache miss. The invariant is "an active pointer must point at a ready generation," and the `DELETE` is how invalidation maintains it. Reverted; `store_sqlite` is green.

## Verification

- `go build ./...`, `go vet`, `golangci-lint` — clean (0 issues)
- `go test -race ./internal/mcp/ ./internal/graph/store_sqlite/` — pass (92s / 395s)
- Unit tests cover the retry-hint countdown, both clamps, the unknown-duration default, and the pending payload's response shape

### DIFF
--- a/cmd/gortex/daemon.go
+++ b/cmd/gortex/daemon.go
@@ -594,19 +594,17 @@ func runDaemonStart(cmd *cobra.Command, _ []string) error {
 		if v1EventHub != nil && mw != nil {
 			go v1EventHub.Run(mw.Events())
 		}
-		// Community detection and process discovery only run when a
-		// repo is tracked or indexed via MCP — a daemon coming up off
-		// a snapshot never triggers them. Fire once here so
-		// get_communities / get_processes / dashboards reflect the
-		// loaded graph instead of returning "run index_repository
-		// first" against a fully populated state.
+		// Community detection and process discovery are a whole-graph pass
+		// costing minutes on a large workspace, and most sessions never ask
+		// for them. Running it here delayed readiness and kept the result
+		// resident for every daemon whether or not anything read it, so the
+		// pass is now started by the first consumer that needs it: those
+		// tools answer with a retry hint while it runs in the background
+		// instead of the old "run index_repository first", which was
+		// misleading against a fully populated graph.
 		if state.mcpServer != nil {
-			analysisStart := time.Now()
-			publishReadinessPhase(state, "analysis", true, nil)
-			state.mcpServer.RunAnalysis()
-			warmup.analysis = time.Since(analysisStart)
-			publishReadinessPhase(state, "analysis_done", true, map[string]any{
-				"elapsed_ms": warmup.analysis.Milliseconds(),
+			publishReadinessPhase(state, "analysis_deferred", true, map[string]any{
+				"reason": "analysis runs on first use",
 			})
 			// Co-change pre-warm: fire the git-history mine in the
 			// background so the first user-visible
--- a/cmd/gortex/mcp.go
+++ b/cmd/gortex/mcp.go
@@ -382,12 +382,8 @@ func runMCP(cmd *cobra.Command, args []string) error {
 			eventHub := hub.New()
 			go eventHub.Run(watcher.Events())
 
-			srv.WatchForReanalysis(eventHub, 500)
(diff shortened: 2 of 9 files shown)


zzet:
Follow-up: removed `WatchForReanalysis` outright rather than leaving it subscribed.

After dropping the recompute, the goroutine's `timer` could never be non-nil, so both `if timer != nil` branches were unreachable and `_ = debounce` / `_ = subID` were noise — all it did was drain the channel. That drain isn't needed either: `Hub.broadcast` uses `select`/`default` and drops events for slow subscribers rather than blocking, so a subscriber that never reads cannot stall the hub or anyone else. A subscription whose only job was the recompute is just a goroutine holding a channel, so the function and its call site are gone.

Also removed the eager `srv.RunAnalysis()` in standalone `gortex mcp`, which I'd missed — it had the same problem as the daemon's and is now covered by `ensureAnalysis`. `eventHub.Run(watcher.Events())` stays: it drains the watcher's channel, which is a separate concern.

Re-verified: build, vet, `golangci-lint` (0 issues), `go test -race ./internal/mcp/` green (95s).

PR #347, opened 2026-07-25
fix(indexer): evict present-but-excluded files on reindex

### LINKED ISSUE #321: bug: newly-excluded files are never evicted from the graph — the walk skips them, reconcile only evicts deleted files; stale nodes persist until untrack+track
**Env:** gortex daemon `vv0.61.0+58dc39b`, Windows 11 Pro (10.0.26200), one tracked repo (~2.6k files, 124,734 nodes), MCP facade v1.

## Repro

1. A directory of generated markdown (4 files) and a legacy TS helper tree (~250 files) are indexed.
2. Add both to `exclude:` in the repo's `.gortex.yaml`; restart the daemon so the config is actually loaded (see the companion reload issue).
3. Full `workspace_admin {operation:"reindex"}` → the walk honors the exclusion: `file_count` drops 2613 → 2305 (= exactly the newly-excluded files).
4. But `node_count`/`edge_count` are unchanged, and `search {operation:"files"}` still returns the excluded files. A scoped reindex over just the excluded dirs doesn't evict them either.
5. Full `workspace_admin {operation:"index"}` (cold path): same — nodes persist.

## Analysis

Reindex "evicts nodes for deleted files". An excluded-but-still-present file is simply never visited by the walk, so it is neither re-parsed nor treated as deleted — its nodes and edges become permanent orphans. The only purge path today is `untrack` + `track` (full retrack), which is expensive and risks daemon-side session/notes state.

This is also a steady source of the "graph holds purgeable rows" problem reported in #311: every exclude-list refinement strands another batch of nodes.

## Expected

After a config change, files that fall outside the admission set should be evicted on the next walk (present-but-excluded ⇒ same as deleted), or a `prune`/`reindex --evict-excluded` escape hatch should exist.

Pitchfork-and-Torch:
## Summary
Fixes #321.

When exclude patterns grow, the discovery walk correctly skips those paths (`file_count` drops), but `IncrementalReindex` only treated `ENOENT` as deletion. Present-but-excluded files kept their graph nodes/edges forever until `untrack` + `track`.

## Change
- If a previously indexed path is missing from the walk but still on disk **and** `shouldExclude` is true → treat as deleted and run the normal eviction path
- If it is on disk and **not** excluded (e.g. language extension set shrank) → still preserve
- Transient stat errors still preserve
- Same rule on scoped deletion detection
- Regression: `TestIncrementalReindex_EvictsExcludedFiles` (replaces the old preserve-excluded test that encoded the opposite policy)

## Test plan
- [ ] CI: `go test ./internal/indexer/ -run EvictsExcluded|EvictsTrulyDeleted`
- Local Go toolchain not available on the contributor machine for this session; logic mirrors existing deletion path after classification.

## Notes
This is a deliberate policy shift from "preserve all present-but-undiscovered files" to "exclude list is admission intent." Non-exclude undiscovered files remain preserved.

### DIFF
--- a/internal/indexer/incremental_reindex_test.go
+++ b/internal/indexer/incremental_reindex_test.go
@@ -17,12 +17,11 @@ import (
 	"github.com/zzet/gortex/internal/graph"
 )
 
-// TestIncrementalReindex_PreservesExcludedFiles is the regression test for
-// the deletion-classification bug: a file that's tracked in fileMtimes but
-// absent from the discovery walk must NOT be purged unless it's truly gone
-// from disk. Otherwise an exclude-list change between passes silently
-// destroys legitimate graph state.
-func TestIncrementalReindex_PreservesExcludedFiles(t *testing.T) {
+// TestIncrementalReindex_EvictsExcludedFiles is the regression for #321:
+// when a previously-indexed file becomes excluded, IncrementalReindex must
+// purge its nodes even though the file still exists on disk. Otherwise
+// file_count drops while node_count/search keep serving orphans.
+func TestIncrementalReindex_EvictsExcludedFiles(t *testing.T) {
 	dir := t.TempDir()
 	writeFile(t, filepath.Join(dir, "keep.go"), `package main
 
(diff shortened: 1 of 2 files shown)


zzet (approved):

zzet:
@Pitchfork-and-Torch thank you for your contribution and for fixing the issue!

Pitchfork-and-Torch:
> @Pitchfork-and-Torch thank you for your contribution and for fixing the issue!

:)

PR #352, opened 2026-07-26
hooks: attribute the invocation log to the agent that ran it

zzet:
**Stacked on #351** — the base is `feat/gortex-doctor-runtime`, since this scopes the findings that PR introduces. Merge #351 first and this diff reduces to the single commit.

Closes the follow-up #351 flagged against itself.

## The problem

The invocation log is the one thing that tells *hook installed* from *hook running* — and it recorded neither which agent ran the hook nor, therefore, which agent was silent. On a machine running Claude Code and Codex side by side the two streams merged, so a busy Claude session vouched for hooks Codex was skipping: precisely the failure the log exists to catch. Doctor had to caveat its own table.

## Why it's cheap

Nothing about a hook command changes. `--agent` already selects the wire protocol, and its empty default already means Claude Code — the value was on the command line the whole time and only needed recording.

That constraint is not incidental. **Codex records trust against each hook command's hash and skips hooks whose definition changed**, so rewriting one to add a flag would have silently disabled every Codex hook until each user re-approved it in `/hooks` — reintroducing the exact bug #348 was about, in the course of building the tool that detects it.

## The asymmetric fallback

Reads are partitioned per agent. The two unknowns are not equally safe, so they are not treated the same:

| log state | `ForAgent` returns | why |
|---|---|---|
| no attribution at all | the union | a log predating the field cannot say an agent was silent; concluding otherwise manufactures a blocker out of missing metadata |
| attributed, agent has no rows | empty | real evidence of silence, and reported as such |

Without that split, the first `gortex doctor` run after an upgrade would accuse every working install of having dead hooks. With it, the blocker still fires the moment the log can actually support it.

Doctor scopes findings to the agent under diagnosis, names the agents it saw, and reports how many rows predate attribution rather than implying a precision the data lacks — those rows clear as the window rolls forward:

```
    531 row(s) in window, 35708 in the whole log
    note: 531 row(s) predate per-agent attribution and are counted for
          every agent; they clear as the window rolls forward.
```

## Tests

- `SetAgent` resolution, including the empty-flag default and case/whitespace forms.
- Reader partitions by agent, tracks `Attributed` and `UnattributedRows`, keeps the union intact as the machine-wide denominator.
- **Both fallback directions** — an unattributed log must not read as proof of silence; an attributed log with no rows for an agent must.
- Writer/reader round trip: `SetAgent("codex")` → log → `ForAgent("codex")` sees it and `ForAgent("claude-code")` does not.
- Diagnose: a busy Claude Code stream no longer suppresses Codex's `none has run` blocker, and the same evidence asked about Claude reads healthy.

`go build ./...`, `go vet ./...`, `go test ./internal/... ./cmd/gortex/` — 12,011 tests across 169 packages, green.


zzet:
## First real run on a mixed machine found two bugs — both fixed

Pushed as `1242f38b`.

### 1. False blockers on a healthy install (critical)

The attribution guard was all-or-nothing: the union stood in only while **no** row carried an agent. But a log isn't attributed or unattributed — it's **mixed**, for as long as the window still reaches back past the upgrade. One attributed row flipped the summary to strict mode and hid every older row behind it.

`PreToolUse` fires constantly and attributes itself within minutes. `SessionStart` fires once per session and hadn't run again yet. Result on a working install:

```
✗ BLOCKER SessionStart is configured but never ran, while other hooks did (not once, ever).
✗ BLOCKER UserPromptSubmit is configured but never ran, while other hooks did (not once, ever).
✗ BLOCKER PostToolUse is configured but never ran, while other hooks did (not once, ever).
```

…against 17,317 rows in the same window saying otherwise. The guard meant to prevent that accusation is what produced it — my tests only covered the two clean states, never the transition every machine passes through.

Unattributed rows are now rolled up **per event** and held out of every agent's counts rather than shared into all of them:

| agent runs | unattributed runs | verdict |
|---|---|---|
| > 0 | any | normal analysis |
| 0 | 0 | genuine silence → blocker |
| 0 | > 0 | **cannot tell** → withheld, reported as INFO |

The table grows an `unattrib` column so a zero explains itself, and dates the last run from unattributed rows when that's all there is — `never` beside evidence of 48 invocations is a contradiction, not a diagnosis:

```
    event                 runs  injected  daemon-up unattrib  last seen
    SessionStart             0         0        n/a       48  2026-07-26T07:44:36
    PreToolUse              46         4      46/46    15764  2026-07-26T08:32:52
```

The regression test reproduces the reported state and fails against the old rule with all three false blockers verbatim.

### 2. `--redact` was a no-op

It promised to hash repo paths and branch names, and rewrote only `Adoption` fields the human report never renders — while printing every absolute path, repo name and account name included, in both sections. Now paths under the repo root render as `<repo>/…` and under home as `~/…`, installed before anything renders so the static report, runtime report, paths quoted inside findings, and `--json` all agree. Branch names aren't paths and can carry a ticket or customer name, so they're hashed.

```
  ✓ daemon handshake: v0.61.4-43-g2b7cdde1 (~/.gortex/cache/daemon.sock)
      ✗ <repo>/.mcp.json (init would create)
    ✗ rule block in ~/.codex/AGENTS.md
```

After both fixes, the same machine reports one genuine finding — it hasn't re-run `gortex install` since #348 merged:

```
! WARN  No Gortex rule block in ~/.codex/AGENTS.md
        → gortex install
✓ OK    Gortex tools are 78% of tool calls (377 Gortex vs 108 shell)
```

`go build ./...`, `go vet ./...`, `go test ./internal/... ./cmd/gortex/` — 11,954 tests across 170 packages, green.

PR #372, opened 2026-07-27
Stop dropping C# reference forms for type names in caseless scripts

yimingll:
## Summary

`emitCSharpReferenceForms` gates every emission on `isCSharpTypeNameCapitalized`, which tests whether the **first byte** is in `'A'..'Z'`. A type named in a script without case — Chinese, Japanese, Korean, Hebrew, Arabic, Thai — can never satisfy that. In such a codebase the gate does not lose precision, it disables the pass: no instantiations, no casts, no `as` / `is`, no `typeof`, no attributes, no generic arguments, no static access.

Measured on a 2.1k-file C# repository whose identifiers are Chinese, gortex v0.61.4:

| | types/interfaces/enums | with an incoming `references` or `instantiates` edge |
|---|---|---|
| Han-initial names | 2003 | **0 (0.0%)** |
| ASCII-initial names | 2292 | 1128 (49.2%) |

All 6691 `instantiates` edges in that graph pointed at ASCII names, while the source contains 2609 `new <Han type>(` sites. A 170-line message-factory file — one `[Kind.X] = () => new X(),` entry per protocol message — carried 17 edges in total, all from its imports and its two declarations.

After this change, on the same repository: Han-initial 1858/2003 (**92.8%**), ASCII-initial 1132/2292 (49.4%, i.e. unchanged), `instantiates` targets 3682 Han-initial / 6712 ASCII.

## Changes

- A name in a **syntactically guaranteed type position** — `new T()`, `new T[]`, `(T)x`, `x as T`, `x is T`, `typeof(T)`, `[T]`, `List<T>` — is emitted on the grammar's authority. No naming heuristic is needed there at all; the capitalisation test was being applied to positions that never had to guess. Primitives are still skipped.
- Type-ness is inferred only for a **member-access receiver**, where `T.X` and `local.X` are genuinely the same shape. An uppercase receiver is accepted exactly as before, so cased-script behaviour is untouched. For a caseless name the question the gate was really asking — *is this receiver bound to a value?* — is answered from the file's own declarations (`csharpDeclaredValueIdents`: locals, fields, parameters, `foreach` variables, pattern designations, catch bindings) instead of from its spelling. The collection is file-scoped rather than block-scoped, which errs toward dropping a reference rather than inventing one.

`isCSharpTypeNameCapitalized` also becomes rune-aware rather than byte-aware, which is what it always meant.

## Testing

- [x] `go test ./internal/parser/languages/` — full package green, no regressions
- [ ] Benchmarks run if performance-relevant — one extra tree walk per C# file to collect declared value identifiers; the pass already walks the tree twice

New tests:

- `TestCSharpTypeUse_CaselessScriptTypeNames` — instantiation, `as`, `is` and static access on Han-named types all mint edges. Verified red before the change (every assertion) and green after.
- `TestCSharpTypeUse_CaselessLocalReceiverIsNotStaticAccess` — a caseless receiver the file declares as a local must not become a type reference; this is the case the capitalisation gate was standing in for.
- `TestCSharpTypeUse_CasedScriptUnchanged` — uppercase receiver still a static access, lowercase still rejected, `new Widget()` still an instantiation.

## Not included

Even with this fix, a reference to an enum **member** (`Kind.Value` read as a value) is recorded against the owning enum type, never the member: `references` targets in that repository's graph are `type`, `method`, `interface` and `param` only, so "is this enum value used?" answers nothing for all 2726 `enum_member` nodes. I prototyped emitting the member as `unresolved::*.<name>` and backed it out — the resolver does not bind that form for `references` edges, so it produced ~85k unresolved edges, delivered no member attribution, and measurably diluted resolution elsewhere (the ASCII column above dropped to 47.3% with it in, and returned to 49.4% with it out). Landing it properly looks like resolver work rather than extractor work, so I left it out rather than guess at your binding rules. Happy to open a separate issue with the measurements if it is useful.

## Checklist

- [ ] Language extractor includes `Meta["methods"]` for interfaces — n/a
- [ ] Methods have `EdgeMemberOf` edges to their containing type — n/a

### DIFF
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -26,7 +26,8 @@ var extractorVersions = map[string]int{
 	// Languages default to version 1 (no salt). Raise an entry here in
 	// the same change that alters a language's extraction logic, e.g.
 	//   "go": 2,
-	"php": 2, // class/interface inheritance now emits typed structural edges
+	"php":    2, // class/interface inheritance now emits typed structural edges
+	"csharp": 2, // null-conditional call edges; reference forms for caseless type names
 }
 
 // extractorSaltExtLang maps a lower-case file extension to the language
--- a/internal/parser/languages/csharp_caseless_typeuse_test.go
+++ b/internal/parser/languages/csharp_caseless_typeuse_test.go
@@ -1,0 +1,107 @@
+package languages
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/zzet/gortex/internal/graph"
+)
+
+// refTargets collects the reference-form edge targets of one kind, keyed by
+// target id, with the ref_context that produced them.
+func refTargets(edges []*graph.Edge, kind graph.EdgeKind) map[string]string {
+	out := map[string]string{}
+	for _, e := range edges {
+		if e.Kind != kind {
+			continue
+		}
+		ctx := ""
+		if e.Meta != nil {
+			if v, ok := e.Meta["ref_context"].(string); ok {
+				ctx = v
+			}
+		}
+		out[e.To] = ctx
+	}
+	return out
+}
+
+// A type named in a script without case — Chinese here, but Japanese, Korean,
+// Hebrew, Arabic and Thai are the same shape — can never be "Capitalized".
+// Gating the reference-form pass on capitalisation therefore did not merely
+// lose precision in such a codebase, it dropped every instantiation, cast and
+// static access in the repository.
+func TestCSharpTypeUse_CaselessScriptTypeNames(t *testing.T) {
+	src := []byte(`public class 消息工厂 {
+    public object 创建(object 输入) {
+        var 请求 = new 钓鱼点覆盖请求();
+        var 类型 = 消息类型.钓鱼点覆盖请求;
+        var 转换 = 输入 as 钓鱼点覆盖请求;
+        if (输入 is 钓鱼点列表响应) { }
+        return 请求;
+    }
+}`)
+	e := NewCSharpExtractor()
+	res, err := e.Extract("消息工厂.cs", src)
+	require.NoError(t, err)
+
+	inst := refTargets(res.Edges, graph.EdgeInstantiates)
+	require.Contains(t, inst, "unresolved::钓鱼点覆盖请求",
+		"`new 钓鱼点覆盖请求()` must mint an instantiates edge")
+
+	refs := refTargets(res.Edges, graph.EdgeReferences)
+	assert.Equal(t, "cast", refs["unresolved::钓鱼点覆盖请求"],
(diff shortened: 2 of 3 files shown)


yimingll:
Pushed one more commit to this branch: `extractorVersions["csharp"] = 2`.

That registry is exactly the mechanism for "extraction logic changed, already-indexed files whose content did not change must be re-extracted", and neither this PR nor #367 (null-conditional call edges, merged earlier today) bumped it. Without the bump, an existing index silently keeps the old C# graph after an upgrade — no new `?.` call edges, no reference forms for caseless type names — until the files happen to change or the store is rebuilt. The bump re-flags only `.cs` files, which is the surgical behaviour the registry is designed for.

So the entry covers both changes; #367 needs no separate follow-up.

`go test ./internal/indexer/` compared against `origin/main` on Windows: 22 pre-existing failures on both, identical sets, no regressions.

zzet:
Hey @yimingll! Thank you for improving C# related quality!

Please, take a look at linter (CI failed)

yimingll:
Thanks for the quick review — fixed.

The linter was right: after the refactor `isCSharpTypeNameCapitalized` had no callers left, since `csharpStaticReceiverEligible` does the uppercase test itself. I removed the function rather than wiring it back in; it was a one-line check with no other users, and dropping it also takes my rune-awareness edit to it out of the diff.

I've installed golangci-lint v2.11.4 locally and run `golangci-lint run` against this branch: the only remaining issue is a pre-existing `errcheck` in `internal/agents/writer_windows_test.go`, which is Windows-only and so never reaches the Linux CI job. `go test ./internal/parser/languages/` is green.

Sorry for making you catch that one — I'll run the linter before pushing from now on.

yimingll:
Lint is green now. The `test (macos-latest, 1.26)` failure in that run looks unrelated to this branch:

```
--- FAIL: TestAtomicBatchConcurrentIdempotencyWritesOnce (0.02s)
FAIL	github.com/zzet/gortex/internal/mcp	176.890s
```

Evidence that it is not this change:

- The branch touches three files — `internal/parser/languages/csharp_typeuse.go`, its new test, and `internal/indexer/extractor_version.go`. Nothing in `internal/mcp`, where that test lives (`internal/mcp/batch_transaction_test.go`).
- The previous CI run on this same branch had byte-identical `internal/mcp` content and both `test` jobs passed; the only delta since is deleting the now-unused `isCSharpTypeNameCapitalized`.
- Locally: `go test ./internal/mcp/ -run TestAtomicBatchConcurrentIdempotencyWritesOnce -count=20` passes 20/20.

So it reads as a flaky concurrency test rather than a regression — `test (ubuntu-latest, 1.26)` was cancelled by the matrix rather than failing on its own.

I can't re-run the jobs myself (fork PR, no admin on this repo). Could you kick a re-run? I'd rather not push a no-op commit and burn the full ~20-minute matrix just to retrigger, but say the word and I will. Happy to open a separate issue for the flake if you want it tracked.

Issue #310, opened 2026-07-20
bug: CLI `gortex call search_text` returns count=0 for every query, including known-positive controls; widening with repo:"*" does not help


tiendungdev:
## Environment

- gortex v0.60.0+dd3bb30, Windows 11 Pro (10.0.26200)
- 1 tracked repo; `daemon status` shows `search sqlite-fts5 docs=72707`
  (so the text index exists and is populated)

## Repro (100% reproducible here)

```
$ gortex call search_text --arg query="setupOutboxDispatcher" --arg limit=3
{"count": 0, "matches": [], "query": "setupOutboxDispatcher",
 "scope_note": "0 results within the active scope (repo:domain-manage) — widen with repo:\"*\" or pass an explicit repo:/project:"}

$ gortex call search_text --arg query="setupOutboxDispatcher" --arg repo="*" --arg limit=3
{"count": 0, ...}

$ gortex call search_text --arg query="CREATE TABLE" --arg repo="*" --arg limit=5
{"count": 0, ...}
```

Control facts:

- `search_symbols` finds `setupOutboxDispatcher` instantly (it exists at
  `services/domain-core/cmd/dispatcher/main.go:388`).
- `git grep -c "CREATE TABLE"` → 39 files under `db/migrations` alone.
- FTS index reports 72,707 docs.

So this is not "term not indexed" — every query returns zero, and the
`scope_note`'s suggested escape hatch (`repo:"*"`) also returns zero, which
makes the note actively misleading.

## Confirmed repro — broken even in SOLO mode on a fresh daemon

Isolated home, fresh store, ONE tracked repo of 3 files containing the
literal `UNIQUE_LITERAL_ALPHA_7391` in an indexed `.go` file (verified via
`search_symbols` finding its enclosing function):

```
SOLO:            count: 0   scope_note: 0 results within the active scope (repo:repo-a) ...
MULTI (2 repos): count: 0
AFTER UNTRACK:   count: 0
repo="*":        count: 0
```

So this is not an artifact of our production daemon's history — the CLI
search_text path returns zero in the **most basic configuration possible**.
Note the scope resolves to `repo:repo-a` while solo-mode matches carry
unprefixed paths (see #309): if the scope filter compares
match-path prefixes against `RepoAllow`, every solo match fails the test.
`repo:"*"` also returning 0 suggests the wildcard from the scope_note hint is
taken as a literal repo name rather than a widen.

## Why this matters more than a broken verb

[...]

PR #409, opened 2026-07-31
fix(store): a passive checkpoint that never ran is not incomplete

### LINKED ISSUE #325: log noise: periodic PASSIVE WAL checkpoint reports "checkpoint incomplete ... context deadline exceeded" with all-zero stats during bulk indexing
**Env:** gortex `v0.61.0`, Windows 11 Pro, during a long `gortex init` bulk index.

## Observed

Every other 5-minute tick while the bulk indexer is writing:

```
2026/07/21 07:56:32 store_sqlite: wal checkpoint incomplete mode=PASSIVE busy=0 wal_frames=0 checkpointed_frames=0 error="context deadline exceeded"
2026/07/21 08:06:31 store_sqlite: wal checkpoint incomplete mode=PASSIVE busy=0 wal_frames=0 checkpointed_frames=0 error="context deadline exceeded"
```

## Analysis (internal/graph/store_sqlite/store.go)

- `walCheckpointInterval = 5 * time.Minute`, `walPassiveCheckpointTimeout = 1 * time.Second` (store.go:522–531).
- The periodic path is documented as best-effort with a non-blocking gate; the 1s context "only protects an unexpected writer-pool wait or a slow driver call".
- During a bulk index the writer pool is occupied ~continuously, so the 1s context expires before the PRAGMA ever runs — which is why every stat in the log line is zero (`busy=0 wal_frames=0 checkpointed_frames=0`): they're the zero-value result struct, not measurements.

So the message reads like a checkpoint that ran and failed ("incomplete"), when it actually never started. Functionally benign (the WAL is checkpointed later), but it's exactly the kind of line a user pastes into a bug report.

## Suggestion

- When the context expires during acquisition (before the PRAGMA), log at debug as `wal checkpoint skipped: writer busy` (or once per bulk phase), reserving the current warning for a checkpoint that genuinely ran and returned an error / left frames behind.
- Alternatively suppress the periodic warning entirely while a bulk/import phase is active.

tiendungdev:
## Summary

The periodic PASSIVE WAL checkpoint reports every failure as `incomplete` with
counters attached. During a bulk index that produces this line every five
minutes:

```
store_sqlite: wal checkpoint incomplete mode=PASSIVE busy=0 wal_frames=0 checkpointed_frames=0 error="context deadline exceeded"
```

Those counters are not measurements. `checkpointWALOnce` returns its
`walCheckpointResult` by value, so a failed `QueryRowContext` hands the caller
the zero value — `busy`, `wal_frames` and `checkpointed_frames` are unset
fields, not numbers SQLite produced. The one-second window is spent waiting for
the single writer connection, which a bulk index holds almost continuously, so
`PRAGMA wal_checkpoint(PASSIVE)` never executes at all. The line nevertheless
reads like a checkpoint that ran and fell short, which is exactly the kind of
line a user pastes into a bug report.

Closes #325.

## Changes

- `checkpointWALPassive` now classifies the failure. An expired window logs
  `deferred mode=PASSIVE reason=writer_busy` — the third shape of a deferral,
  alongside the `writer_gate` and `bulk_writer` branches the function already
  distinguishes a few lines above.
- A checkpoint that did run keeps its `incomplete` warning together with its
  real counters, so a reader holding frames back stays visible.
- `passiveCheckpointTimeout` makes the window injectable so the deadline can be
  exercised deterministically instead of by racing a real bulk load. Its zero
  value selects `walPassiveCheckpointTimeout`, mirroring `busyRetryTimeout` in
  the same struct.
- No change to draining behaviour: the WAL is still checkpointed on a later
  tick, or by the final TRUNCATE checkpoint in `Close`.

## Testing

`go test -race -count=1 ./internal/graph/store_sqlite/` — Windows 11, go1.26.5.

Two new tests in `wal_checkpoint_log_test.go`:

- `TestPassiveCheckpointWriterDeadlineIsNotReportedAsIncomplete` — an expired
  window prints neither `incomplete` nor zero-value counters, and is reported
  as a `writer_busy` deferral.
- `TestPassiveCheckpointReaderLimitedStillWarns` — a checkpoint that genuinely
  ran and left frames behind must still warn.

Verified by sabotage, because the second test is what keeps the fix honest:
disabling the new classification branch turns the first test red, and treating
every failure as a deferral turns the second one red. Deleting the log call
outright — the tempting one-line "fix" — is caught by the second test.

Pre-existing failures on this platform, unchanged by this PR:
`TestBundlePackageKey`, `TestBundleCache_CrossRepoIsolation`,
`TestImportAdjacencyProjectionSkipsUnrelatedOutgoingRows`, `TestSweepPlanLocks`,
`TestSweepPlanLockReceiverRebindBatch`, `TestSweepWarnPlanLocks`,
`TestPreparedStatementPlansNeverScanBigTables`. The same seven fail identically
on `74d2296` without this change; four are `TempDir` cleanup races on Windows,
and `TestBundlePackageKey` is a path-separator mismatch (`pkg\sub` vs
`pkg/sub`) of the same class as #310 — happy to send that one separately.

## Checklist

- [ ] `go test -race ./...` across the whole repository — I ran the touched
      package only; a full-repo race run does not finish in reasonable time on
      this Windows machine, so I am leaving that to CI rather than claiming it.
- [ ] Benchmarks — not performance-relevant (one log line changes wording).

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -11,6 +11,7 @@ terms (Apache 2.0, Section 5).
 | Name | GitHub | Added |
 |------|--------|-------|
 | Andrey Kumanyaev | [@zzet](https://github.com/zzet) | 2024-01-01 |
+| Tien Dung Dao | [@tiendungdev](https://github.com/tiendungdev) | 2026-07-31 |
 
 ## Past Contributors
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -11,7 +11,6 @@ terms (Apache 2.0, Section 5).
 | Name | GitHub | Added |
 |------|--------|-------|
 | Andrey Kumanyaev | [@zzet](https://github.com/zzet) | 2024-01-01 |
-| Tien Dung Dao | [@tiendungdev](https://github.com/tiendungdev) | 2026-07-31 |
 
 ## Past Contributors
 
--- a/internal/graph/store_sqlite/store.go
+++ b/internal/graph/store_sqlite/store.go
@@ -55,6 +55,11 @@ type Store struct {
 	busyRetries        atomic.Uint64
 	busyRetryExhausted atomic.Uint64
 
+	// passiveCheckpointTimeout bounds one periodic PASSIVE checkpoint. The zero
+	// value selects walPassiveCheckpointTimeout; tests shorten it to exercise
+	// the writer-pool deadline deterministically.
+	passiveCheckpointTimeout time.Duration
+
 	// dbPath is the on-disk SQLite file path, retained for size
 	// telemetry — the WAL high-water mark surfaces in daemon_health so a
 	// runaway -wal is observable rather than silently filling the disk.
@@ -561,6 +566,13 @@ func (s *Store) runCheckpointLoop(interval time.Duration) {
 	}
 }
 
+func (s *Store) passiveCheckpointWindow() time.Duration {
+	if s.passiveCheckpointTimeout > 0 {
+		return s.passiveCheckpointTimeout
+	}
+	return walPassiveCheckpointTimeout
+}
+
 func (s *Store) checkpointWALPassive() {
 	if !s.writeMu.TryLock() {
 		log.Printf("store_sqlite: wal checkpoint deferred mode=PASSIVE reason=writer_gate")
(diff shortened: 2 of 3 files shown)


CONTRIBUTORS.md
```diff
@@ -11,6 +11,7 @@ terms (Apache 2.0, Section 5).
 | Name | GitHub | Added |
 |------|--------|-------|
 | Andrey Kumanyaev | [@zzet](https://github.com/zzet) | 2024-01-01 |
+| Tien Dung Dao | [@tiendungdev](https://github.com/tiendungdev) | 2026-07-31 |
```
zzet: Hey @tiendungdev! Such changes better to submit in a dedicated PR

zzet (changes requested):

CONTRIBUTORS.md
tiendungdev: Okay bro, Claude push haha.

tiendungdev:
Agreed — dropped it in 1fa79bb5, so this PR is now just the checkpoint fix and its test. Will send the CONTRIBUTORS entry separately if you want it at all.

zzet (changes requested):
Hey @tiendungdev, sorry it sumbitter previous feedback without a full review. I found one blocker in the PR’s core logic.

A deadline does not prove the checkpoint never ran. The new classification https://github.com/zzet/gortex/blob/e21ed878bb601599bc1e61c6ff79505d1538698d/internal/graph/store_sqlite/store.go#L587-L605 labels every expired context as `writer_busy`. However, that context explicitly covers both pool waiting and slow driver execution https://github.com/zzet/gortex/blob/e21ed878bb601599bc1e61c6ff79505d1538698d/internal/graph/store_sqlite/store.go#L533-L536 while `QueryRowContext(...).Scan(...)` https://github.com/zzet/gortex/blob/e21ed878bb601599bc1e61c6ff79505d1538698d/internal/graph/store_sqlite/store.go#L658-L670 combines connection acquisition, checkpoint execution, and result scanning.

Consequently, SQLite could have partially executed the checkpoint—or returned genuine incomplete counters just before the deadline—and the code would discard that evidence and log `writer_busy`. The broad `ctx.Err()` check can also hide an unrelated SQLite error if the deadline expires immediately after it returns.

The 1 ns test https://github.com/zzet/gortex/blob/e21ed878bb601599bc1e61c6ff79505d1538698d/internal/graph/store_sqlite/wal_checkpoint_log_test.go#L43-L61 uses a free writer pool, so it only proves behavior with an already-expired context, not connection contention.

The fix should acquire `writerDB.Conn(ctx)` separately and classify only an acquisition timeout as `writer_busy`. Execution timeouts should be reported separately, and `errSQLiteCheckpointIncomplete` should be handled before inspecting `ctx.Err()`. The test should explicitly pin the sole writer connection.

tiendungdev:
You are right, and the assumption was load-bearing — thanks for the second pass.

`QueryRowContext` folds pool acquisition, PRAGMA execution and result scanning into one call, and the window covers all three, so an expired context could equally mean the writer was never obtained, that SQLite was midway through a drain, or that it had just returned genuine counters. Labelling all of it `writer_busy` threw away the only measurement in the last two cases.

Fixed in a0508467:

- **Acquire separately.** `writerDB.Conn(ctx)` first; only a failure to get it proves the PRAGMA never executed, and only that is reported as a deferral next to `writer_gate` / `bulk_writer`.
- **Order by how much is known** once execution began:
  - `errSQLiteCheckpointIncomplete` + counters → warn with the counters, even if the window expired right after
  - execution timeout → `timed out ... phase=execute`, no counters printed
  - anything else → `failed ... error=...`
- `errSQLiteCheckpointIncomplete` is now checked before `ctx.Err()`, so a real result outranks a deadline that landed on its heels. `checkpointWALOnce` gained a variant over a `walCheckpointQueryer` (satisfied by both `*sql.DB` and `*sql.Conn`) so only the PASSIVE path changes.

On the test — also a fair hit, the 1 ns version proved nothing about contention.
`TestPassiveCheckpointPinnedWriterIsDeferredNotIncomplete` now pins the sole writer connection (`writerDB` is capped at one), so the deadline is genuinely spent waiting for acquisition.

One thing I could not pin with a live deadline: whether SQLite returns counters just before or just after the window expires is a race, and that ordering is the substance of the fix. I split `passiveCheckpointReport` out so the three cases can be driven directly.

Sabotage-checked both ways — swapping the deadline check ahead of the counters check turns the ordering test red; removing the acquisition split turns the pinned-writer test red with `timed out ... phase=execute` where a deferral belongs.

Local: `golangci-lint v2.11.4` clean, `-race` clean over the checkpoint tests, and `internal/graph/store_sqlite` reports the same seven failures as the baseline (two separator cases, four plan-lock TempDir cleanups, one adjacency projection) — none new.

zzet (approved):

PR #410, opened 2026-07-31
fix(indexer): a route's file and line must describe the same site

### LINKED ISSUE #322: bug: `analyze kind=routes` mixes fields — `line` is the mux registration line but `file` is the handler's file (Go net/http 1.22 method patterns)
**Env:** gortex daemon `vv0.61.0+58dc39b`, Windows 11 Pro, Go repo using net/http 1.22 method patterns, MCP facade v1.

## Repro shape

Package with the registration and the handlers in **different files**:

```go
// handler.go
func (h handlers) Register(mux *http.ServeMux) {
    ...
    mux.HandleFunc("POST /api/admin/things/{id}/request-update", h.requestUpdate) // handler.go:166
}

// actions.go
// requestUpdate handles POST /api/admin/things/{id}/request-update.
func (h handlers) requestUpdate(w http.ResponseWriter, r *http.Request) { ... } // actions.go:~200
```

`analyze {kind:"routes"}` returns:

```
file:    .../actions.go        <-- handler's file
line:    166                   <-- registration line, which is in handler.go
handler: .../actions.go::handlers.requestUpdate
method:  POST
path:    /api/admin/things/{id}/request-update
```

Verified with `git grep`: `actions.go:166` is unrelated code inside another handler's body; the literal registration lives at `handler.go:166`. Method/path/handler symbol are all correct — only the `file`+`line` pair is incoherent, so consumers who cite `file:line` point at the wrong place.

## Expected

`file:line` should describe one site consistently (the registration site), or the row should carry both explicitly, e.g. `route_file:route_line` + `handler_file:handler_line`.

tiendungdev:
## Summary

`analyze kind=routes` reports a `file` and a `line` that come from two
different places whenever a route is registered in one file and handled in
another. Minimal reproduction — registration literal at `handler.go:8`,
handler body at `actions.go:11`:

```json
{
  "file":    "route-repro/actions.go",
  "line":    9,
  "handler": "route-repro/actions.go::handlers.requestUpdate",
  "method":  "POST",
  "path":    "/api/admin/things/{p1}/request-update"
}
```

`actions.go:9` is a comment. The method, path and handler symbol are all
correct; only the `file`+`line` pair is incoherent, so anything citing a
route's location points at arbitrary code. The compact renderer prints
exactly that pair — `(%s:%d)` — and an agent auditing an API surface ends up
citing a line that registered nothing.

Closes #322.

## Root cause

`resolveProviderHandlers` (internal/indexer/indexer.go) re-points a route
contract at its resolved handler so `EnrichHTTPContractWithTree` can find the
handler body in that file's tree:

```go
patched := c
patched.SymbolID = handlerNode.ID
patched.FilePath = handlerNode.FilePath
...
contracts.EnrichHTTPContractWithTree(&patched, lines, nodes, lang, tree)
...
matches[i] = patched          // written back into the registry
```

The comment above it describes the swap as temporary — "swap it in
temporarily to the resolved handler so the lookup works" — but `patched` is
written back, and `Line` is never swapped along with `FilePath`. So `FilePath`
follows the handler while `Line` stays on the registration literal, and every
downstream consumer inherits the mixed pair: the contract node, the
`EdgeHandlesRoute` edge (which takes both from the contract), and the
analyzer row.

Single-file registration hides this completely, because both spellings
coincide — which is why the existing fixtures never caught it.

## Fix

Restore `FilePath` after enrichment. `SymbolID` stays re-pointed: the resolved
cross-file handler is the answer this pass exists to produce, and the `handler`
field already carries the handler's file, so naming the registration site in
`file:line` loses no information.

```diff
+registrationFile := patched.FilePath
 patched.SymbolID = handlerNode.ID
 patched.FilePath = handlerNode.FilePath
 ...
 contracts.EnrichHTTPContractWithTree(&patched, lines, nodes, lang, tree)
+patched.FilePath = registrationFile
```

## One existing assertion changed — please sanity-check this call

`TestResolveProviderHandlersUsesOneNameAndFileBatch` asserted the re-homed
`FilePath`. That test is about batching — it counts `findNodesByNames` and
`getFileNodesByPaths` calls — and used `FilePath` only as evidence that the
patch had been applied. It now checks `SymbolID` for that instead, which is the
resolution the pass produces, and pins `FilePath` to the registration site. Its
fixture already separated the two (contract at `repo/router.go`, handler at
`repo/handlers.go`), so the case was expressible; only the expectation encoded
the mixed pair.

If you consider the re-homed `FilePath` intentional, the alternative is to
carry both sites explicitly (`route_file`/`route_line` plus
`handler_file`/`handler_line`) as #322 suggests. That changes the response
shape, so I did not do it unilaterally — happy to switch if you prefer it.

## Testing

`go test -count=1 ./internal/indexer/ ./internal/mcp/ ./internal/contracts/` —
Windows 11, go1.26.5.

New `TestRouteContractKeepsRegistrationSiteCoherent`: a route registered in one
file and handled in another keeps `file:line` on the registration and still
resolves its handler across files.

Verified by sabotage, since either half alone would look like a fix: dropping
the restore turns the coherence assertions red, and dropping the `SymbolID`
re-point turns the cross-file assertion red. The pair pins "borrow and give
back" rather than abandoning the re-point.

Pre-existing failures on this platform: `internal/indexer` and `internal/mcp`
fail **85** tests on `74d2296` without this change (largely `TempDir` cleanup
races and clangd/compile-commands cases specific to Windows). With this change
the same run reports 89; the four-test difference is
`TestResolveProviderHandlersUsesOneNameAndFileBatch` (the assertion updated
above, recorded before that edit) plus `TestReviewPackCoverageEvidenceIsRecorded`,
`TestSiblingDiffContext_GCXAndTOONAndBudget` and `TestVerifyCitation_BadSHA` —
all four pass when run targeted on this branch, and the latter three touch
nothing this PR changes.

## Checklist

- [ ] `go test -race ./...` across the whole repository — I ran the three
      affected packages without `-race`; a full-repo race run does not finish
      in reasonable time on this Windows machine, so I am leaving that to CI
      rather than claiming it.
- [ ] Benchmarks — not performance-relevant.

### DIFF
--- a/internal/indexer/contract_import_resolve.go
+++ b/internal/indexer/contract_import_resolve.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"sort"
 	"strings"
 	"sync"
 
@@ -20,12 +21,27 @@ func (mi *MultiIndexer) disambiguateBareTypesViaImportsBatch(registries []*contr
 	srcCache := map[string][]byte{}
 	importCache := map[string]map[string]string{}
 	nameSet := map[string]struct{}{}
+
+	// A bare type is resolved through the imports of the file that mentions
+	// it — the file defining the contract's symbol. For a route registered in
+	// one file and handled in another that is not c.FilePath, which names the
+	// registration site, so resolve the symbol's own file first. One batched
+	// node lookup, mirroring the FindNodesByNames projection below.
+	srcByContract := handlerSourceFiles(registries, g)
+	srcFile := func(c contracts.Contract) string {
+		if p := srcByContract[c.SymbolID]; p != "" {
+			return p
+		}
+		return c.FilePath
+	}
+
 	for _, cr := range registries {
 		if cr == nil {
 			continue
 		}
 		for _, c := range cr.All() {
-			if c.Meta == nil || !isImportResolvableLang(c.FilePath) {
+			src := srcFile(c)
+			if c.Meta == nil || !isImportResolvableLang(src) {
 				continue
 			}
 			for _, key := range []string{"response_type", "request_type"} {
@@ -34,8 +50,8 @@ func (mi *MultiIndexer) disambiguateBareTypesViaImportsBatch(registries []*contr
 					continue
 				}
 				nameSet[name] = struct{}{}
-				if isRustFile(c.FilePath) {
-					for _, candidateName := range mi.rustImportCandidateNames(c.FilePath, name, srcCache) {
+				if isRustFile(src) {
+					for _, candidateName := range mi.rustImportCandidateNames(src, name, srcCache) {
 						nameSet[candidateName] = struct{}{}
 					}
 				}
@@ -52,8 +68,40 @@ func (mi *MultiIndexer) disambiguateBareTypesViaImportsBatch(registries []*contr
 		if cr == nil {
 			continue
 		}
-		mi.disambiguateBareTypesViaImportsPrefetched(cr, srcCache, importCache, candidatesByName)
+		mi.disambiguateBareTypesViaImportsPrefetched(cr, srcCache, importCache, candidatesByName, srcByContract)
 	}
+}
+
+// handlerSourceFiles maps a contract's SymbolID to the file its symbol is
+// defined in. Contracts whose symbol is unknown or absent from the graph are
+// omitted; callers fall back to Contract.FilePath for those.
+func handlerSourceFiles(registries []*contracts.Registry, g graph.Store) map[string]string {
+	idSet := map[string]struct{}{}
+	for _, cr := range registries {
+		if cr == nil {
+			continue
+		}
+		for _, c := range cr.All() {
+			if c.SymbolID != "" {
+				idSet[c.SymbolID] = struct{}{}
+			}
+		}
+	}
+	if len(idSet) == 0 {
+		return nil
+	}
+	ids := make([]string, 0, len(idSet))
+	for id := range idSet {
+		ids = append(ids, id)
+	}
+	sort.Strings(ids)
+	out := make(map[string]string, len(ids))
+	for id, n := range g.GetNodesByIDs(ids) {
+		if n != nil && n.FilePath != "" {
+			out[id] = n.FilePath
+		}
+	}
+	return out
 }
 
 func (mi *MultiIndexer) disambiguateBareTypesViaImportsPrefetched(
@@ -61,17 +109,26 @@ func (mi *MultiIndexer) disambiguateBareTypesViaImportsPrefetched(
 	srcCache map[string][]byte,
 	importCache map[string]map[string]string,
 	candidatesByName map[string][]*graph.Node,
+	srcByContract map[string]string,
 ) {
 	for _, c := range cr.All() {
 		if c.Meta == nil {
 			continue
 		}
-		if !isImportResolvableLang(c.FilePath) {
+		// The imports to consult are the ones in the file defining this
+		// contract's symbol, not the file it was registered in.
+		src := c.FilePath
+		if p := srcByContract[c.SymbolID]; p != "" {
+			src = p
+		}
+		if !isImportResolvableLang(src) {
 			continue
 		}
 		patched := false
 		items := cr.ByID(c.ID)
 		for i := range items {
+			// FilePath still identifies which registry entry this is —
+			// registration site, unchanged.
 			if items[i].FilePath != c.FilePath || items[i].Meta == nil {
 				continue
 			}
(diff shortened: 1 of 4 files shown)


zzet (changes requested):
He @tiendungdev. I see a potential issue in the change:

Contract.FilePath is restored before handler-based enrichment finishes.
 Restoring the registration path at indexer.go:6305–6316 https://github.com/zzet/gortex/blob/954dfc4fa46b6b7aa1dca016ae9f7fa30eade7e3/internal/indexer/indexer.go#L6305-L6316 correctly fixes the persisted route location. However, resolveCallReturnTypes runs afterward https://github.com/zzet/gortex/blob/954dfc4fa46b6b7aa1dca016ae9f7fa30eade7e3/internal/indexer/indexer.go#L5857-L5908

 That pass obtains body facts from the resolved handler via c.SymbolID, but then constructs a semantic binding site using c.FilePath together with the handler’s line number at indexer.go:6866–6881 https://github.com/zzet/gortex/blob/954dfc4fa46b6b7aa1dca016ae9f7fa30eade7e3/internal/indexer/indexer.go#L6866-L6881 The lookup uses that exact file/line pair indexer.go:6802–6820 https://github.com/zzet/gortex/blob/954dfc4fa46b6b7aa1dca016ae9f7fa30eade7e3/internal/indexer/indexer.go#L6802-L6820 while compiler bindings are stored against the handler’s actual source file provider.go:1074–1090 https://github.com/zzet/gortex/blob/954dfc4fa46b6b7aa1dca016ae9f7fa30eade7e3/internal/semantic/goanalysis/provider.go#L1074-L1090

For a route registered in router.go and handled in actions.go, this therefore queries router.go:<handler-line> instead of actions.go:<handler-line>. It can silently lose the response type or select an unrelated binding at the same line.

The same overloaded field is used for TypeScript/Rust import disambiguation contract_import_resolve.go:28–39 https://github.com/zzet/gortex/blob/954dfc4fa46b6b7aa1dca016ae9f7fa30eade7e3/internal/indexer/contract_import_resolve.go#L28-L39 contract_import_resolve.go:65–84 https://github.com/zzet/gortex/blob/954dfc4fa46b6b7aa1dca016ae9f7fa30eade7e3/internal/indexer/contract_import_resolve.go#L65-L84 causing it to inspect the router’s imports instead of the handler’s.

The new test only checks the final edge and its handler has no schema behavior route_registration_site_test.go:51–88 https://github.com/zzet/gortex/blob/954dfc4fa46b6b7aa1dca016ae9f7fa30eade7e3/internal/indexer/route_registration_site_test.go#L51-L88 so it cannot detect this regression.

Please keep FilePath:Line as the registration location, but derive handler-source paths from the SymbolID node—or represent the two locations separately—and add a cross-file schema/type-enrichment test.

tiendungdev:
Confirmed on both counts — thanks, this was a real regression and my test could not have caught it.

`prepareContractTypeResolution` was pairing `c.FilePath` with `binding.Line`, and the line comes from the handler body. For a cross-file route that asks for `<registration file>:<handler line>`, which describes nothing. The fix in 23a88943 takes both halves from one source: `handlers[c.SymbolID]` is already in hand there — it is what produced the body facts the line came from — so the site now uses its `FilePath`.

`contract_import_resolve` had the same shape: a bare type resolved through the imports of `c.FilePath` would inspect the router file. It now derives the path from `SymbolID` via `handlerSourceFiles()` — one batched `GetNodesByIDs`, mirroring the `FindNodesByNames` projection next to it — falling back to `Contract.FilePath` when a contract has no symbol or the symbol is not in the graph.

Left alone deliberately: the `items[i].FilePath != c.FilePath` comparison. That is an identity check between registry entries sharing an ID, and `Contract.FilePath` also serves as a dedup key (`contract_bridge.go`, `incremental_contracts.go`) and the changed-file re-scan trigger. Only the "which file do I read source from" meaning moved.

On the test gap: `TestContractBindingSitesUseTheHandlersOwnFile` records the sites the pass queries through a store implementing `SemanticBindingTypeReader`, so it needs no live compiler-facts provider and runs anywhere. Sabotage-checked — restoring `c.FilePath` there turns it red.

Worth being explicit about the other test, since it is the same trap you flagged: `TestContracts_CrossFileHandlerStillResolvesResponseType` (the two-file form of `TestContracts_ResponseFromMakeSlice`) does **not** fail under that sabotage — its response type resolves through the literal fallback, not a binding site. I found that by sabotaging it rather than trusting the green. It is kept for the invariant it does pin (FilePath on the registration site, SymbolID still resolving cross-file); the recorder test is the one guarding this regression.

Local before pushing: `golangci-lint v2.11.4` clean, `go build ./cmd/gortex/` ok, `go vet` clean, `internal/indexer` 24 failures — none new against 74d2296 (TempDir-cleanup and clangd cases) — and all three new tests pass under `-race`.

zzet (approved):

PR #411, opened 2026-07-31
fix(paths): package identity is slash-only, not filepath-shaped

tiendungdev:
## Summary

A package key, a cluster label and a cache fingerprint key are **identifiers
inside the graph** — none of them is ever opened from disk. They are built with
`path/filepath` anyway, so on Windows they come out spelled with backslashes
while everything they are compared against stays forward-slash, and the
comparison quietly fails.

`bundlePackageKey` is the clearest case:

```go
dir := filepath.Dir(filepath.ToSlash(filePath))
```

`ToSlash` rewrites the separator and `filepath.Dir` immediately cleans it back,
so the normalisation is undone by the line wrapping it. The result is then
looked up in a fingerprint map keyed `"repoA/pkg"` while the key itself reads
`"repoA\pkg"` — no entry ever validates, and **the bundle cache serves nothing
on Windows**.

`longestCommonDirPrefix` shows the same mismatch inside a single function:

```go
pfx := filepath.Dir(paths[0])
...
cut := strings.LastIndex(pfx, "/")
if cut < 0 { pfx = ""; break }
```

The trimming loop cuts at `"/"` — the author knew these are slash paths — but
the prefix is built with the OS separator, so the cut never lands and the
function bails out empty on the first mismatch. Every cluster label on Windows
is either `""` or `"gortex\internal\parser\languages"`.

## Changes

Switch these to `path.Dir`. This is not a normalisation layered on top:
`path` is the correct package for identifiers, `path/filepath` is for talking
to the filesystem. `filepath.Base` is deliberately left alone — it accepts both
separators and returns the same answer either way.

| site | package | effect on Windows |
|---|---|---|
| `bundlePackageKey` | `graph/store_sqlite` | bundle cache never hits |
| `packageKey` | `analysis` | the function the cache mirrors; same key, same mismatch |
| `longestCommonDirPrefix`, `mixedClusterLabel` | `analysis` | cluster labels empty or backslash-spelled |
| `commonFilePrefix` | `mcp` | containment test joins with `"/"`, so an OS-separator prefix can never match |

## Testing

Windows 11, go1.26.5, measured against `74d2296` as the baseline:

| package | baseline | with this change |
|---|---|---|
| `internal/analysis` | 6 failures | **2** |
| `internal/graph/store_sqlite` | 7 failures | **5** |

The six that flip: `TestBundlePackageKey`,
`TestBundleCache_CrossRepoIsolation`, `TestPureClusterLabel`,
`TestMixedClusterLabel`, `TestInferCommunityLabelDistinguishesSameBasename`,
`TestDetectCommunitiesLeidenIncremental`. Nothing regresses — I ran the same
packages on `74d2296` and diffed the failure sets rather than eyeballing them.

What remains in those packages is `TempDir`-cleanup contention on Windows and
one diff-parsing case (`TestParseDiffLinesNewSide`), both unrelated to this
change and failing identically on the baseline.

`TestBundlePackageKeyNeverUsesOSSeparator` states the contract on its own
instead of leaving it implicit in a table that only fails off-POSIX. It skips
where the separator already is `"/"`, so it is a no-op on the platforms CI runs
today.

## Why this class of bug reaches users

The test matrix is `[ubuntu-latest, macos-latest]` and `build-windows` only
builds — so nothing in CI ever evaluates a path comparison on a platform whose
separator differs. On POSIX both spellings coincide and every one of these
sites looks correct.

I am not proposing a matrix change here (a good number of tests fail on
Windows today for unrelated reasons, mostly `TempDir` cleanup contention, so
turning it on now would just paint CI red). Mentioning it because it explains
why these went unnoticed, and I am happy to work through the remaining Windows
failures separately if that is a direction you want.

There is a sibling case in `internal/analysis/diffmap.go` — `parseDiffLines`
cleans git's forward-slash paths with `filepath.Clean` — but fixing it in
isolation breaks the review-pack consumers, which normalise inconsistently
(`tools_review.go` has both `filepath.Clean` and `filepath.ToSlash(filepath.Clean(...))`).
That needs the whole chain audited together, so I left it out of this PR
rather than half-fixing it.

## Checklist

- [ ] `go test -race ./...` across the whole repository — I ran the affected
      packages without `-race`; a full-repo race run does not finish in
      reasonable time on this machine, so I am leaving that to CI rather than
      claiming it.
- [ ] Benchmarks — not performance-relevant in the measured sense, though it
      does restore a cache that currently never hits on Windows.

### DIFF
--- a/internal/analysis/communities.go
+++ b/internal/analysis/communities.go
@@ -541,7 +541,7 @@ func inferCommunityLabel(members []string, nodeMap map[string]*graph.Node, files
 		return np
 	}
 
-	return path.Dir(files[0])
+	return path.Dir(filepath.ToSlash(files[0]))
 }
 
 // pureClusterLabel returns a name for clusters whose files share a
@@ -567,9 +567,10 @@ func pureClusterLabel(files []string) string {
 func mixedClusterLabel(files []string) string {
 	dirCount := make(map[string]int)
 	for _, f := range files {
-		// Slash-only: this directory is rendered into the label and trimmed
-		// by helpers that split on "/".
-		dirCount[path.Dir(f)]++
+		// Graph file paths keep OS-native separators (see graphRelKey), so
+		// normalise before taking the directory: the result is rendered into
+		// the label and trimmed by helpers that split on "/".
+		dirCount[path.Dir(filepath.ToSlash(f))]++
 	}
 	if len(dirCount) == 0 {
 		return ""
@@ -606,13 +607,14 @@ func longestCommonDirPrefix(paths []string) string {
 	if len(paths) == 0 {
 		return ""
 	}
-	// path.Dir, not filepath.Dir: the trimming loop below cuts at "/", so a
-	// prefix cleaned to the OS separator can never be shortened and the
-	// function bails out with "" on the first mismatch — on Windows that made
-	// every cluster label empty.
-	pfx := path.Dir(paths[0])
+	// Graph file paths keep OS-native separators (see graphRelKey), so
+	// normalise with ToSlash first, then take the directory with path.Dir —
+	// the trimming loop below cuts at "/", so a prefix carrying the OS
+	// separator could never be shortened and the function would bail out with
+	// "" on the first mismatch.
+	pfx := path.Dir(filepath.ToSlash(paths[0]))
 	for _, p := range paths[1:] {
-		dir := path.Dir(p)
+		dir := path.Dir(filepath.ToSlash(p))
 		for pfx != "" && !isPathPrefix(dir, pfx) {
 			cut := strings.LastIndex(pfx, "/")
 			if cut < 0 {
--- a/internal/analysis/communities.go
+++ b/internal/analysis/communities.go
@@ -3,6 +3,7 @@ package analysis
 import (
 	"fmt"
 	"math"
+	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
@@ -540,7 +541,7 @@ func inferCommunityLabel(members []string, nodeMap map[string]*graph.Node, files
 		return np
 	}
 
-	return filepath.Dir(files[0])
+	return path.Dir(files[0])
 }
 
 // pureClusterLabel returns a name for clusters whose files share a
@@ -566,7 +567,9 @@ func pureClusterLabel(files []string) string {
 func mixedClusterLabel(files []string) string {
 	dirCount := make(map[string]int)
 	for _, f := range files {
-		dirCount[filepath.Dir(f)]++
+		// Slash-only: this directory is rendered into the label and trimmed
+		// by helpers that split on "/".
+		dirCount[path.Dir(f)]++
 	}
 	if len(dirCount) == 0 {
 		return ""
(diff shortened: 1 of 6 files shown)


zzet (changes requested):
Hey @tiendungdev, I see potentially blocking finding

`communities.go` now calls path.Dir directly on graph file paths:
`return path.Dir(files[0])` https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/analysis/communities.go#L544
`dirCount[path.Dir(f)]` https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/analysis/communities.go#L567-L573
`path.Dir(paths[0])` and `path.Dir(p)` https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/analysis/communities.go#L609-L615

However, graph paths are explicitly OS-native on Windows. `graphRelKey` https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/indexer/indexer.go#L1453-L1472 preserves native separators, and in `indexFile` nodes/FilePaths are backslash-keyed on Windows https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/indexer/indexer.go#L3819-L3833
Community finalization passes `Node.FilePath` through unchanged communities.go:770–784 https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/analysis/communities.go#L770-L784

Because `path.Dir` only recognizes `/`, a production Windows path such as `internal\analysis\a.go` becomes `"."`. In multi-repo mode, `repo/internal\analysis\a.go` becomes only `"repo"`. Consequently, unrelated directories collapse onto the same community label.

The fix should normalize first, matching the correct implementation in `packageKey` https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/analysis/incremental_communities.go#L223-L235

`path.Dir(filepath.ToSlash(filePath))`

The community tests use hard-coded forward-slash inputs communities_label_test.go:5–86
https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/analysis/communities_label_test.go#L5-L86 so they do not represent production Windows graph paths. The new separator test also skips on POSIX https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/internal/graph/store_sqlite/bundle_cache_test.go#L109-L118 while Windows CI only builds and tests internal/agents
https://github.com/zzet/gortex/blob/9a7beacf5781aac30d3cf2f413e733e130153ad6/.github/workflows/ci.yml#L36-L57 That explains why all 11 checks pass (https://github.com/zzet/gortex/pull/411/checks).

The `packageKey`, `bundlePackageKey`, and `commonFilePrefix` changes otherwise look correct. I found no additional blocker.

tiendungdev:
You are right, and the branch was worse than what it replaced — thanks for catching it.

I had assumed graph paths were forward-slash by contract. `graphRelKey` says otherwise in its own comment: OS-native separators are kept deliberately so an incremental evict lookup matches the exact string the cold walk stamped. So a real Windows path is `gortex/internal\parser\languages\cpp.go` — prefix joined with `/`, remainder not — and `path.Dir` on that collapses to the prefix. Unrelated directories sharing a label is exactly what would have happened.

Fixed in 0ffad383: the four `communities.go` sites now normalise first, the way `packageKey`, `bundlePackageKey` and `commonFilePrefix` already did.

    path.Dir(filepath.ToSlash(p))

On the test gap — also a fair hit. The existing label fixtures are forward-slash, so they pass either way, and my `...NeverUsesOSSeparator` test skipped on POSIX, which makes it worthless against the CI you actually run.

`TestClusterLabelsNormaliseOSNativeGraphPaths` builds its inputs with `filepath.FromSlash`, reproducing the `<prefix>/<os-native remainder>` shape nodes really carry. That is an identity on POSIX (so it runs as an ordinary passing test there, not a skip) and a real assertion on Windows. Sabotage-checked: dropping the `ToSlash` calls turns it red with precisely your scenario —

    pureClusterLabel(...)        = ""         want "parser/languages"
    longestCommonDirPrefix(...)  = "gortex"   want "gortex/internal/parser"

Package results on this machine, measured against 74d2296: `internal/analysis` 6 failures -> 2, `internal/graph/store_sqlite` 7 -> 5, nothing new anywhere. The remainder is TempDir-cleanup contention and one diff-parsing case, both unrelated.

zzet (approved):

PR #415, opened 2026-07-31
test(mcp): release sidecar handles so TempDir can remove them on Windows

tiendungdev:
## Summary

`OpenSidecar` keeps handles in a process-wide cache keyed by absolute path, so
a manager built over a test directory holds that file open for the life of the
test binary. POSIX unlinks an open file happily and nothing shows. Windows
refuses, `t.TempDir()`'s `RemoveAll` fails, and the test is reported as failed
even though every assertion in it passed:

```
testing.go:1464: TempDir RemoveAll cleanup: unlinkat
...\001\sidecar.sqlite: The process cannot access the file because it is
being used by another process.
```

**28 tests** across the `mcp` package fail this way. Nothing in the message
points at the sidecar, so each one reads as an unrelated flake rather than one
shared cause.

## Changes

**`persistence.CloseSidecar(path)`** — closes the cached sidecar rooted at
`path` and drops it from the shared cache.

`5980a1f3` reached for `OpenSidecar(path).Close()` to release one test's
handle. That works, but it *creates* the file and its parent directory when
nothing was open — repeated across 28 call sites it would leave a trail of
empty databases behind every run. `CloseSidecar` returns `nil` when the cache
has nothing for that path, so a caller can release unconditionally.

**Test helpers**, registered through `t.Cleanup`. Ordering matters here:
cleanups run LIFO, so registering *after* `t.TempDir()` has queued its own
removal closes the handle first. Three shapes, because managers root their
sidecar differently and the difference is not visible at the call site:

| helper | for |
|---|---|
| `tempSidecarDir` | the directory *is* the data dir — notes, memories, promoted tools, scopes (via `filepath.Dir`) |
| `tempNotebookRepo` | `newNotebookManager` roots at `<repo>/.gortex` |
| `closeSidecarOnCleanup` | explicit, for tests deriving a cache dir themselves (memories) or passing two (learned tools) |

## Testing

Windows 11, go1.26.5, measured against `6d97beb1`:

| | before | after |
|---|---|---|
| affected test set | 29 failures | **1** |
| of those, `TempDir` cleanup | **28** | **0** |

The whole `mcp` package still reports 33 failures, **none of them new** — I ran
the same package on the baseline and diffed the failure sets. The remainder are
ordinary assertion failures unrelated to teardown, including
`TestNotesManager_SaveQueryDelete`, which fails identically without this
change.

Also clean: `golangci-lint v2.11.4` on `./internal/mcp/... ./internal/persistence/...`,
`go vet`, and `-race` over the affected tests. (`internal/persistence`'s
`TestFeedbackDir` fails on the baseline too — a separate path-separator issue,
same class as #411, untouched here.)

## Why this matters beyond the red

The test matrix is `[ubuntu-latest, macos-latest]` and `build-windows` only
builds, so nothing in CI ever exercises a teardown path where an open file
cannot be unlinked. That is why a whole cluster of tests can be red on Windows
indefinitely without anyone noticing.

I am not proposing a matrix change in this PR. Two smaller clusters remain
first — the plan-lock sqlite handles in `store_sqlite` and one git-dir case in
`analysis` — and both are a different mechanism than this one, so they want
separate changes. Happy to work through them if that is a direction you want.

## Checklist

- [x] New tests added for new functionality — n/a; this fixes existing tests'
      teardown. `CloseSidecar` is exercised by every helper that calls it.
- [ ] `go test -race ./...` across the whole repository — I ran the affected
      packages plus `-race` over the touched tests; a full-repo race run does
      not finish in reasonable time on this machine, so I am leaving that to
      CI rather than claiming it.
- [ ] Benchmarks — not performance-relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,8 +37,8 @@ jobs:
     # CGO build smoke-test on native Windows. tree-sitter needs a C/C++
     # compiler — the GitHub windows runner ships mingw-w64 on PATH, so no
     # extra toolchain setup is required. `go build ./...` compiles every
-    # production package; the focused agents test also exercises Windows-only
-    # integration paths without opting into the unsupported full test suite.
+    # production package; the focused agents and sidecar tests also exercise
+    # Windows-only paths without opting into the unsupported full test suite.
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -55,6 +55,13 @@ jobs:
 
       - name: Test Windows agent integrations
         run: go test -timeout=5m -count=1 ./internal/agents
+
+      # Windows refuses to unlink an open file, so a leaked sidecar handle
+      # pins its directory and fails the enclosing t.TempDir() cleanup. That
+      # asymmetry is invisible on the linux/macos matrix, which is why this
+      # one test runs here rather than only there.
+      - name: Test sidecar handle release
+        run: go test -timeout=5m -count=1 -run TestCloseSidecar ./internal/persistence
 
   build-linux-static:
     # The linux release ships a statically linked binary so it runs on any
--- a/internal/mcp/index_health_tool_surface_test.go
+++ b/internal/mcp/index_health_tool_surface_test.go
@@ -19,7 +19,7 @@ func TestIndexHealth_SurfacesToolPreset(t *testing.T) {
 	require.False(t, hasLearned)
 
 	// After a learned promotion, the count + names surface.
-	srv.InitLearnedTools(t.TempDir(), t.TempDir())
+	srv.InitLearnedTools(tempSidecarDir(t), tempSidecarDir(t))
 	srv.RecordLearnedPromotion("get_architecture", "")
 	payload = srv.buildIndexHealthPayload()
 	require.Equal(t, 1, payload["learned_tools"])
--- a/internal/mcp/memories_test.go
+++ b/internal/mcp/memories_test.go
@@ -169,6 +169,7 @@ func TestMemoryManager_PersistenceRoundTrip(t *testing.T) {
 	repoPath := filepath.Join(tmp, "fake-repo")
 	require.NoError(t, os.MkdirAll(repoPath, 0o755))
 	cacheDir := filepath.Join(tmp, "cache")
+	closeSidecarOnCleanup(t, cacheDir)
 
 	mm := newMemoryManager(cacheDir, repoPath)
 	_, err := mm.Save(persistence.MemoryEntry{
--- a/internal/mcp/notes_test.go
+++ b/internal/mcp/notes_test.go
@@ -98,7 +98,7 @@ func TestNotesManager_UpdateMissingID(t *testing.T) {
 }
 
 func TestNotesManager_PersistenceRoundTrip(t *testing.T) {
-	cache := t.TempDir()
+	cache := tempSidecarDir(t)
 	repo := "/tmp/notes-test-repo"
 
 	nm1 := newNotesManager(cache, repo)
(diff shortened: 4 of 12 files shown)


zzet (changes requested):
I like the implementation, but I’d request one test change before merging.

The new `CloseSidecar` https://github.com/zzet/gortex/blob/8df1c84edd303edf14a3977556981b100a6861b7/internal/persistence/sidecar_sqlite.go#L337-L361 looks correct for test teardown. Path normalization matches `OpenSidecar`, cache misses do not create files, and the cleanup helpers https://github.com/zzet/gortex/blob/8df1c84edd303edf14a3977556981b100a6861b7/internal/mcp/sidecar_testing_test.go#L23-L47 compute the correct paths and run before `TempDir` removal.

The significant gap is that none of this is tested by CI:

  - On Linux/macOS, the modified tests would pass even if `CloseSidecar` were a no-op because POSIX allows deleting open files.
  - The Windows job only builds and tests internal/agents.

I’d add a focused persistence test covering:

  1. Closing an unopened path creates nothing.
  2. Open → CloseSidecar → reopen returns a fresh, usable store.
  3. Repeated closing remains harmless.
  4. On Windows, the containing directory can be removed afterward.

Then run only that test in the Windows job; enabling the whole MCP suite is unnecessary.

One non-blocking caveat: `CloseSidecar` unlocks between finding and closing the store https://github.com/zzet/gortex/blob/8df1c84edd303edf14a3977556981b100a6861b7/internal/persistence/sidecar_sqlite.go#L354-L360 so a concurrent OpenSidecar could receive the soon-to-be-closed handle https://github.com/zzet/gortex/blob/8df1c84edd303edf14a3977556981b100a6861b7/internal/persistence/sidecar_sqlite.go#L247-L251 

That is acceptable for documented, quiescent teardown but should be hardened if this becomes a general lifecycle API.

tiendungdev:
Fair — the fix was unverified exactly where it matters. Addressed in 5ca38c4d.

`TestCloseSidecar_ReleasesTheHandleAndStaysHarmless` covers the four cases you listed:

- closing a path that was never opened creates neither the database nor its parent directory (the reason this is not `OpenSidecar(path).Close()`)
- open → `CloseSidecar` → reopen returns a *different*, usable store — usable meaning it serves a query, not merely non-nil
- closing repeatedly is harmless
- the containing directory can then be removed

The reopen case is what gives it teeth on every platform: with `CloseSidecar` stubbed to `return nil` the cache still holds the old store, so the reopen hands back the same pointer and the assertion fails on Linux too. The `RemoveAll` case is the one that only bites on Windows — sabotage-checked, the stub turns the reopen case red.

Wired into `build-windows` next to the existing agents step, and I spelled the asymmetry out in the comment so the reason it lives there does not get lost:

```yaml
# Windows refuses to unlink an open file, so a leaked sidecar handle
# pins its directory and fails the enclosing t.TempDir() cleanup. That
# asymmetry is invisible on the linux/macos matrix, which is why this
# one test runs here rather than only there.
- name: Test sidecar handle release
  run: go test -timeout=5m -count=1 -run TestCloseSidecar ./internal/persistence
```

Agreed on keeping the whole MCP suite out of it — that would need the plan-lock and git-dir clusters cleaned up first, and those are a different mechanism.

Also took the non-blocking note: the entry is now dropped under the same lock that found it, so a concurrent `OpenSidecar` misses the cache and opens a fresh handle rather than receiving one about to close. `Close` still runs outside the lock, since it takes `sidecarMu` itself to clear any other keys pointing at the same store.

Local: `golangci-lint v2.11.4` clean, `-race` clean, YAML parses. `internal/persistence` reports one failure — `TestFeedbackDir` — which fails identically on a clean `origin/main` (`\home\user\.cache\gortex` vs `.cache/gortex`, same class as #411, untouched here).

zzet (approved):

PR #416, opened 2026-07-31
Fix four stacked search-recall failures: post-filter wipeouts, cross-repo FTS floods, exact-name container hijack, warm-restart degradation

pbednarcik:
## Summary

`search_symbols` for a suffix-convention name — `Extensions`, `Controller`, `Repository`, the shapes .NET codebases name by — returned **zero results** on a production C# solution whose graph demonstrably held 67 matching classes (exact-name lookup found each one). Root-causing that single symptom uncovered four independent mechanisms stacked on top of each other, each one masking the next. This PR fixes all four, each with a failing-test-first repro:

1. **Post-filter wipeout** (`internal/mcp/tools_core.go`). Corpus / repo / kind filters run after a bounded fetch; when noise rows own the whole BM25 head, the filters empty the page and the existing zero-result fallbacks re-run the same poisoned fetch. Fix: when the raw fetch had candidates but post-filtering emptied it, refetch at ×5 then ×25 depth under the same scope — before the fallbacks that relax the caller's filters. Response carries `fetch_escalated: true`.

2. **Cross-repo floods outrun any bounded over-fetch** (`store_fts.go` + engine threading). The FTS query is repo-blind; the engine's scope filter runs over a head capped at 200 rows. One tracked Go repo contributes 232 exact-token `Extensions` symbols — deeper than the cap, so a scoped session could never see the other repo's classes, and BM25's IDF makes the flood direction flip as index composition changes. Fix: `SearchSymbolsRepoScoped` / `SearchSymbolBundlesRepoScoped` filter `repo_prefix` inside the SQL (the same UNINDEXED-literal-column shape the per-repo staleness wipe uses), threaded to the engine via a `ScopedSymbolBundleSearcher` capability with `Swappable` / `HybridBackend` forwards — the `SymbolBundleSearcher` pattern. Unowned rows (empty prefix) always pass, matching the `ScopeAllows` carve-out. With scoping, the flagship query's true best match ranks **0** instead of past-200.

3. **Container nodes hijack the exact-name tier** (`store_fts.go`). Tier 0 short-circuits to exact-name hits before FTS. Two `.csproj` module nodes named `Extensions` — in-scope, exactly named — eclipsed every `*Extensions` class, and since downstream shaping drops container kinds, the page came back empty while looking "handled" to the wipeout rescue. Fix: container/binding kinds (module, package, file, import, local, param, generic param, closure, builtin) no longer short-circuit; they remain reachable through FTS ranking. Exact-name hits on real symbols keep the fast path.

4. **Warm-restarted daemons silently degrade to substring search** (`internal/query/engine.go`). The engine gates the backend on `Count() > 0` — but `Count` is a delta of this process's Add/Remove calls (its own doc warns it is not a corpus size), so a daemon restarted over a populated disk FTS reads 0 and every search takes the substring fallback until the first file edit. Easy to miss when dogfooding (active development bumps the counter constantly); guaranteed for read-mostly consumers. Fix: `backendHasCorpus` consults the authoritative `DocCount` (forwarded through `Swappable` / `HybridBackend`) before falling back.

## Real-world results (reference .NET solution, ~8,000 indexed files, two-repo workspace)

- `Extensions`: 0 → **71 hits**, the four same-named module-variant classes first, immediately after a cold daemon restart with no file activity.
- `Controller` / `Repository` suffix queries return the expected controllers/repositories.
- Exact-name, camelCase-prefix, and cross-repo Go queries unchanged (regression battery on both repos).
- Each mechanism was bisected live against the real store (a copy driven through store → bundles → engine layers) before its fix.

A fifth commit hardens the four fixes against an adversarial review pass: the corpus gate memoises its positive answer (the `DocCount` fallback is a real `COUNT(*)`; a corpus never un-populates within a backend's lifetime) and releases the `Swappable` read-lock before querying; the engine's scoped over-fetch cap no longer clamps below an explicit caller limit (which silently flattened the escalation's deeper iterations into byte-identical re-queries); a scoped query that legitimately finds nothing returns a non-nil empty page instead of falling back to the doomed unscoped flood fetch; the escalation skips the content corpus (those rows live only in `content_fts`), caps its deepest fetch, and honours context cancellation; an all-`false` `RepoAllow` map can no longer select the scoped path with an empty (= unscoped) allow list; and the tier-0 contract is pinned by tests in both directions — exact symbol-kind hits keep the fast path, container hits ride along for `kind:`-filtered callers rather than being droppable.

## Known gaps (documented, not silent)

- Param / generic-param nodes are still FTS-indexed and dilute BM25 heads (the KindLocal exclusion rationale applies to them verbatim); excluding them needs an index rebuild, so it is left for a follow-up.
- The escalation depths are bounded (×25, absolute cap 2000); a flood deeper than that falls back to the previous empty-page behaviour rather than an unbounded scan.
- Repo-grain narrowing rides the FTS query; project-grain (sub-repo) scope still filters post-fetch over the bounded head.
- The escalation refetch re-runs the code-shaped channel only; a `corpus:docs` wipeout relies on prose ranking within it rather than re-running the widened doc channel.

## Testing

- [x] New: 3 handler tests (flood rescue, no-flag-on-hit, true-miss no-refetch), 9 scoped-store tests (cross-repo flood, exact-name scope, unowned carve-out, nil-scope equivalence, multi-repo allow list, scoped bundles, container-kind gate, container merge-along, exact-type-wins pin), 6 engine tests (scoped-path selection, unscoped preservation, corpus gate warm/empty, DocCount contract corners, gate memoisation, cap-piercing depth).
- [x] `go test` on mcp / query / search / store_sqlite packages — failure set on the Windows dev box is byte-identical to clean `origin/main` (pre-existing platform failures; verified by name-level diff).

## Checklist

- [x] Code follows existing patterns (capability-interface threading mirrors `SymbolBundleSearcher`; SQL shape mirrors the per-repo wipe)

### DIFF
--- a/internal/graph/store.go
+++ b/internal/graph/store.go
@@ -724,6 +724,16 @@ type SymbolBundleSearcher interface {
 	SearchSymbolBundles(query string, limit int) ([]SymbolBundle, error)
 }
 
+// ScopedSymbolBundleSearcher is the repo-narrowed variant of
+// SymbolBundleSearcher: repoAllow filters inside the FTS query, not
+// over a bounded fetched head — a post-fetch filter starves when
+// out-of-scope rows outrank every in-scope one deeper than any
+// over-fetch. Unowned rows (empty repo prefix) always pass, the
+// ScopeAllows carve-out.
+type ScopedSymbolBundleSearcher interface {
+	SearchSymbolBundlesRepoScoped(query string, repoAllow []string, limit int) ([]SymbolBundle, error)
+}
+
 // VectorItem is the payload BulkUpsertEmbeddings takes per node:
 // the node's ID and its embedding vector. Length of Vec must
 // match the dim the corresponding BuildVectorIndex call declared
--- a/internal/graph/store_sqlite/fts_repo_scope_test.go
+++ b/internal/graph/store_sqlite/fts_repo_scope_test.go
@@ -1,0 +1,141 @@
+package store_sqlite
+
+import (
+	"fmt"
+	"path/filepath"
+	"testing"
+
+	"github.com/zzet/gortex/internal/graph"
+)
+
+// floodedScopeStore seeds two repos: "noise" holds noiseCount symbols
+// whose token bag is exactly the query token (they win BM25 outright),
+// "app" holds one WidgetExtensions. This is the cross-repo flood shape:
+// every unscoped head slot goes to noise rows, so any post-fetch scope
+// filter starves no matter how far the caller over-fetches.
+func floodedScopeStore(t *testing.T, noiseCount int) *Store {
+	t.Helper()
+	s, err := Open(filepath.Join(t.TempDir(), "scope.sqlite"))
+	if err != nil {
+		t.Fatalf("Open: %v", err)
+	}
+	t.Cleanup(func() { _ = s.Close() })
+
+	appID := "app/widget.go::WidgetExtensions"
+	s.AddNode(&graph.Node{
+		ID: appID, Kind: graph.KindType, Name: "WidgetExtensions",
+		FilePath: "app/widget.go", Language: "go", RepoPrefix: "app",
+	})
+	if err := s.UpsertSymbolFTS(appID, "widget extensions"); err != nil {
+		t.Fatalf("UpsertSymbolFTS(app): %v", err)
+	}
+
+	for i := 0; i < noiseCount; i++ {
+		id := fmt.Sprintf("noise/f%d.go::Extensions", i)
+		s.AddNode(&graph.Node{
+			ID: id, Kind: graph.KindFunction, Name: "Extensions",
+			FilePath: fmt.Sprintf("noise/f%d.go", i), Language: "go", RepoPrefix: "noise",
+		})
+		if err := s.UpsertSymbolFTS(id, "extensions"); err != nil {
+			t.Fatalf("UpsertSymbolFTS(noise %d): %v", i, err)
+		}
+	}
+	return s
+}
+
+// TestSearchSymbolsRepoScoped_CrossRepoFloodDoesNotStarve: with 250
+// out-of-scope rows outranking the one in-scope symbol, the scoped
+// search must return that symbol — filtering inside the FTS query,
+// not over a bounded head.
+func TestSearchSymbolsRepoScoped_CrossRepoFloodDoesNotStarve(t *testing.T) {
+	s := floodedScopeStore(t, 250)
+
+	hits, err := s.SearchSymbolsRepoScoped("Extensions", []string{"app"}, 20)
+	if err != nil {
+		t.Fatalf("SearchSymbolsRepoScoped: %v", err)
+	}
+	if len(hits) != 1 || hits[0].NodeID != "app/widget.go::WidgetExtensions" {
+		t.Fatalf("scoped search = %+v, want exactly the app symbol", hits)
+	}
+}
+
+// TestSearchSymbolsRepoScoped_ExactNameTierRespectsScope: the exact-name
+// tier-0 short-circuit must not leak out-of-scope nodes; when every
+// exact-name hit is out of scope it falls through to the scoped FTS.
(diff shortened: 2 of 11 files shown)


zzet (changes requested):
Hey @pbednarcik! The overall design is good - especially filtering repo scope inside SQLite - but there is one confirmed pagination bug.

1. Escalation can stop before the requested cursor page is reachable. The loop treats any surviving result as success and immediately breaks (rescue loop)
https://github.com/zzet/gortex/blob/b6da50f30c5fd695996176f805dd4bb28b6aa8e3/internal/mcp/tools_core.go#L1691-L1727 Pagination happens later, so five rescued results with offset 20 produce an empty page and no next cursor (pagination) https://github.com/zzet/gortex/blob/b6da50f30c5fd695996176f805dd4bb28b6aa8e3/internal/mcp/tools_core.go#L1892-L1911

I reproduced this with 445 higher-ranked doc hits followed by 30 valid code hits: the 5× fetch finds five code results and stops; page 2 is empty, although the 25× fetch would expose all 30. Escalation should continue until offset+limit is reachable or the backend is demonstrably exhausted.

2. The scoped-empty sentinel does not prevent an unscoped retry. The adapter deliberately returns a non-nil empty slice to mean “scoped search answered zero” (sentinel contract) https://github.com/zzet/gortex/blob/b6da50f30c5fd695996176f805dd4bb28b6aa8e3/internal/search/symbolsearcher_backend.go#L103-L122 However, bundleHandled becomes true only for non-empty results (handling) https://github.com/zzet/gortex/blob/b6da50f30c5fd695996176f805dd4bb28b6aa8e3/internal/query/engine.go#L722-L765 so a genuine scoped miss still enters the unscoped search path https://github.com/zzet/gortex/blob/b6da50f30c5fd695996176f805dd4bb28b6aa8e3/internal/query/engine.go#L797-L835 That recreates the cross-repo flood cost the PR is intended to remove. Track “scoped query answered” separately from “returned candidates.”

3. Performance note: the 2,000 limit can materialize 4,000 bundles because the engine requests limit*2, and both 5× and 25× iterations can collapse to the same capped query. Avoid duplicate effective depths.

Overall - strong approach, but fix the cursor-window condition before merging is needed.

pbednarcik:
Thanks for the review — and for the repro on the pagination case, it pinned the shape exactly. All three addressed in 3d0afe7d, each with a failing-test-first repro:

1. **Escalation now targets the cursor window.** The loop no longer treats any surviving result as success: a partial rescue keeps the best set so far and keeps escalating until `len(kept) >= offset+limit` or the backend is demonstrably exhausted (short raw page). Your scenario is pinned as `TestSearchSymbols_EscalationReachesCursorWindow` — 170 doc rows + 30 code symbols, `offset 20 / limit 10`: the ×5 depth (175) rescues five and used to stop there with an empty page 2; the loop now continues to ×25 and the page holds the remaining ten.

2. **A scoped zero is now an answer.** `scopedAnswered` (non-nil result, empty included) is tracked separately from "returned candidates", and flips `bundleHandled` on its own — a genuine scoped miss no longer falls through to the unscoped channel fetch. Pinned by `TestGatherSymbolCandidates_ScopedZeroDoesNotFloodFetch` with call counters on both the unscoped-bundle and `Search` paths. One incidental find while writing it: the test fake originally flattened the non-nil-empty sentinel back to nil via `append(nil, ...)` — the fake now preserves nil-vs-empty explicitly, so the sentinel contract is pinned from both sides.

3. **Duplicate capped depths are skipped.** Once the 2,000 cap clamps a multiplier, the next iteration would collapse to an identical query — the loop now breaks instead of paying it twice. `TestSearchSymbols_EscalationSkipsDuplicateDepth` records the backend-visible limits and asserts a single deep call (previously two — and the recording also shows your ×2 observation: the 2,000-row fetch reaches the backend as `limit*2 = 4000`). With the dedupe, the worst case per wipeout is one such materialisation; if you'd rather the cap account for the engine's doubling (i.e. bound the *materialised* rows at 2,000), happy to halve it.

Touched packages are green; the full-suite failure set on my Windows box remains name-identical to clean `origin/main`.

zzet (approved):

PR #432, opened 2026-08-01
csharp: resolve extension-method calls by namespace visibility

pbednarcik:
## Summary

`get_callers` on a C# extension method returned **zero edges** for the most common shape .NET DI configuration takes: every module defines its own `ContainerExtensions.ValidateRegistrations(this IContainer)` and calls it from its `Startup` via extension syntax. Extraction already stamps `extension` / `this_param_type`, and `tryBindCSharpExtension` already existed — but its precision rules refuse every multi-candidate case (typed tie, or same name in several namespaces), and the multi-module pattern *is* a multi-candidate case by construction.

The missing tiebreaker is C#'s actual scoping rule: an extension method is callable only when its namespace is an enclosing namespace of the call site or imported by a using directive. This PR threads that visibility into the extension binder — reusing the `csharpFileNamespaceSet` machinery the type narrowing already has — and pairs it with a receiver-eligibility model so visibility never substitutes for applicability. Six commits, each behavior failing-test-first:

1. **Visibility narrowing** (`csharp_extension.go`). Multi-candidate ties narrow to the call site's visible namespaces — the calling method's own enclosing-namespace chain first (deepest wins, matching member-lookup order), then imports. A typed tie that narrows to one binds at the typed tier; an untyped set that narrows to one binds at the inferred tier; two-visible still refuses.

2. **`global using` and `using static`** (`csharp.go` + `csharp_ns_narrow.go` + `csharp_extension.go` + `cross_pkg_guard.go`). The extractor stamps both under their own file-node meta keys; the resolver propagates global usings across the declaring file's **compilation unit** (nearest-ancestor csproj directory — project-wide inheritance across sibling subtrees, zero leakage into nested projects with their own csproj) and admits a `using static` class's extensions directly. One deliberate side effect to be aware of: the propagated globals enrich the same `imported` set the **pre-existing type narrowing** reads, so a type-reference tie can now also break on a sibling file's `global using` — that is the correct visibility semantics, but it reaches beyond extension calls. This also needed a cross-package-guard keep-rule: extension visibility is namespace-scoped, not import-scoped — a bind justified by a sibling file's `global using` (or the enclosing chain) leaves no import edge on the calling file, and the reachability closure was reverting exactly the correct binds.

3. **Hardening** (first adversarial review round; every item repro'd first): the first cut of the contradiction veto (receiver evidence matching no candidate refuses on an exact key match; refined into the tiered rule of commit 5); Windows path IDs (the store joins the repo prefix with `/` but keeps OS-native `\` below — without splitting on both, every `global using` keyed to the repo root and leaked repo-wide, invisible to forward-slash fixtures); the calling method's own `scope_ns` chain for the enclosing tier (a sibling `namespace` block in the same file must not out-rank the call site's own); mixed-generation stores keep their `EdgeImports` fallback; the global-usings index gets pass lifetime instead of a per-page O(graph) rebuild under a contended lock; a guard revert clears the stale `resolution` tag; the three usings keys joined `metaStringSliceKeys` with the legacy `[]any` shape pinned.

4. **Extractor version bump** (`csharp` 2 → 3). The stamps are extraction-time: without the salt, nothing re-extracts unchanged `.cs` files automatically after an upgrade and the feature stays dormant — found live, on a store carrying every fix and zero new stamps. Note the salt's reach honestly: stores with `index.merkle` enabled restage `.cs` automatically; default-config stores surface the staleness through the MCP freshness rider (the salt's only default-config consumer) and pick the stamps up on their next full reindex.

5. **Receiver eligibility** (second, independent review round). A typed receiver is matched against `this_param_type` in tiers, and only a *provable* contradiction refuses:
   - verbatim match on suffix-trimmed forms, with BCL aliases folded (`String`/`System.Int32` ↔ `string`/`int`);
   - a namespace-qualified receiver matches a bare this-param only when that bare name actually **denotes** the receiver's type from the extension's own file (see commit 6) — otherwise `Data.Inner` would falsely match an unrelated `Vendor.Inner` extension;
   - no name match → walk the receiver's declared base/interface chain: reaching a candidate's this-param is a typed match (`Crate : IBox` → `Foo(this IBox)`);
   - `Foo<T>(this T)` and `this object` match any receiver (exempt from the veto, no typed-tier confidence);
   - an unresolved edge in the receiver's hierarchy (external base) makes an "unrelated" verdict unreliable — the untyped rules (visibility narrowing, then the unique-name bind) run instead, over the candidates that unknown part could satisfy;
   - only a receiver that is provably unrelated (or absent from the repo) to every candidate refuses — which is what stops a visible same-name extension from swallowing a BCL instance call (`xs.Add(1)` vs `Add(this Widget)`).
   Primitive-typed locals (`int n; n.Foo()`) carry the evidence via a separate, **method-scoped** `receiver_builtin` stamp (primitives stay out of `receiver_type` so the receiver-gate passes keep their contract). The indexer's incremental reuse key reads the same stamp, so an edit that changes a local's builtin type re-resolves instead of re-pinning the stale bind.

6. **Denotation anchoring** (third adversarial review round; every item repro'd by a failing test first). Bare type names now resolve by C#'s actual lookup order from the position that wrote them: the innermost enclosing namespace declaring a same-name type **claims** the name (shadowing) before any using directive may supply it. That one rule anchors the qualified-receiver match (commit 5's visibility check alone let `using Data;` in the extension's file outweigh the extension's own `Vendor.Inner`), both ends of the inheritance walk (the seed set is now only what the receiver can denote at the call site — an unrelated same-name type's hierarchy no longer donates a match — and a reached type satisfies a candidate only when the candidate's this-param denotes it), and unresolved base stubs, which the walk chases by the same lookup so in-pass edge ordering cannot fake an incomplete hierarchy. Around it, the same round fixed: the local type environment is **method-scoped for user types** too, matching the builtin stamps (a sibling method's `Widget s` was typing this method's `string s`); the incomplete-hierarchy waiver excludes candidates the external part provably cannot satisfy (in-repo or sealed-builtin this-params) and competes with — rather than losing to — the generic catch-alls; typed binds require the same visibility the guard keep-rule re-checks, killing a bind/revert cycle (guard reverts also clear the stale confidence label now); restubbed member calls (eviction flattens `x.Pack()` to a bare unresolved name) route back through the extension rule via their surviving Meta — receiver stamps, a new extraction-time `member_call` marker, or the old resolution tag — instead of being re-pinned by a repo-wide locality guess that bypasses the veto; and the receiver gate's extension exemption now belongs to the *target*, not the edge's tag, so a stale tag cannot shield a misattributed rebind. The hierarchy walk memoizes per-type ancestor closures and the type-name lookup for the pass — one BFS per unique receiver type instead of one per call site.

## Real-world results (reference .NET solution, ~8,000 files, cold-built store)

- The flagship shape: four same-named `ValidateRegistrations(this IContainer)` extensions across four module namespaces, each called from its module's `Startup` via a property receiver (no type evidence) — `get_callers` went 0 → 1 per variant, **each Startup bound to its own module's copy**, via the enclosing-namespace tier.
- **1,031 extension-method call edges** resolved across the workspace from scratch (`resolution: extension_method`), with builtin-receiver stamps feeding the typed tier.
- Regression battery unchanged on this store: suffix-convention searches, exact-name lookups, and the namespace-narrowing results for types (the type narrowing's *inputs* do gain the propagated globals — see commit 2).

## Known gaps (documented, not silent)

- Generic type arguments are invisible to eligibility: extraction strips them on both sides, so a `List<int>` receiver can typed-match `Foo(this List<string>)`. In compiling code the shape requires a same-name extension differing only in type arguments while the real target is external — rare, but a wrong edge when it happens; carrying argument fidelity through extraction is a follow-up.
- Block-scoped `using` directives are flattened to file level, so an inner scope's import cannot out-rank an outer enclosing namespace's extension the way real C# lookup does — a wrong-edge residue confined to that (rare) overlap; a per-scope using model is out of scope here.
- `global using static` is stamped but not project-propagated (miss-only: the declaring file sees it, siblings don't).
- The local type environment is method-scoped, not block-scoped: two same-named locals of different types in *sibling blocks of one method* still last-wins (the cross-method bleed is fixed).
- A user-declared class named like a BCL alias (`class String`) still folds to the keyword and can typed-match `this string`.
- Eligibility walks *declared, in-graph* hierarchy only: a receiver whose relation to the this-param runs through a type the index doesn't hold resolves via the incomplete-hierarchy fallback at the inferred tier — restricted to candidates the unknown part could satisfy — or, when the hierarchy looks complete but isn't, loses the edge (missing, never wrong).
- Visibility evidence is graph-derived and treated as complete once any of it exists; the zero-visibility fallback covers pre-stamp graphs, not partially-stale ones.
- Editing only a `Usings.cs` re-resolves dependent files on the next full pass, not incrementally (frontier design).

## Testing

- [x] New: 36 resolver tests (visibility tiers, typed/untyped ties, contradiction veto + its exemptions — interface reach, generic catch-all, aliases, incomplete-hierarchy fallback, provably-unrelated refusal — builtin + qualified receivers, enclosing-namespace shadowing of bare this-params, anchored hierarchy-walk seeds and reached types, waiver scope incl. builtin/in-repo exclusions and the catch-all tie, typed-bind visibility with no bind/revert churn, restubbed-member-call routing (typed refusal + visibility rebind), stale-tag gate exemption, visibility-cache clearing, suffix-trim fixpoint, false canonical-match refusal, sibling-block attribution, global-using propagation, two no-leak shapes incl. Windows path IDs, csproj-unit scoping, using-static, guard keep-rule gates, cross-directory enclosing keep, file-scoped namespaces, `[]any` meta shapes, mixed-generation stamp gate), 8 extractor tests (usings-scope stamps, builtin receiver stamp + method scoping, user-type local method scoping, member-call marker, generic this-param stamp, builtin suffix-order parity), 1 indexer reuse-key test, plus an extractor-version staleness assertion.
- [x] `go test` on resolver / parser / query / store_sqlite / indexer — the failure set on the Windows dev box is name-identical to clean `origin/main` (pre-existing platform failures; verified by stash-diff at multiple points).

## Checklist

- [x] Code follows existing patterns (visibility tiering mirrors `csharpNarrowByNamespace`; the hierarchy walk mirrors the receiver gate's reachability conservatism; the version bump uses the extractor-salt registry as designed)


zzet (changes requested):
Hey @pbednarcik! Thank you for addressesing a real C# resolution gap and the implementation effort is impressive. The PR has excellent motivation, a strong real-world fixture, substantial adversarial testing, good Windows-path handling, compatible SQLite metadata decoding, careful cache locking, and useful restubbed-call hardening. The test-to-production-code ratio is also healthy. But it introduces several high-confidence wrong edges plus a watcher-path performance regression.

### Blocking correctness issues

#### Extensions can override inherited instance methods

For:
```csharp
     class Base { public void Foo() {} }
     class Derived : Base {}
     static class E { public static void Foo(this Base x) {} }

     Derived d = new();
     d.Foo();
```
Gortex binds `E.Foo`; C# must bind `Base.Foo`. The new hierarchy-matching branch https://github.com/zzet/gortex/blob/8adb9605c44bc9093512ee7028aa169e5188cd01/internal/resolver/csharp_extension.go#L118-L154 never checks the receiver hierarchy for an applicable instance member before binding an extension.

The C# specification https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#128103-extension-method-invocations says extension lookup begins only when normal member lookup finds no applicable method.

#### Receiver type shape is erased

Receiver extraction https://github.com/zzet/gortex/blob/8adb9605c44bc9093512ee7028aa169e5188cd01/internal/parser/languages/csharp.go#L1462-L1484 and candidate normalization https://github.com/zzet/gortex/blob/8adb9605c44bc9093512ee7028aa169e5188cd01/internal/resolver/csharp_extension.go#L643-L697 discard array, nullable, and generic structure.

I reproduced this valid case:

```csharp
     // App namespace
     Foo(this string value)

     // imported Lib namespace
     Foo(this string[] value)

     string[] xs;
     xs.Foo();
```

Gortex selects the scalar Foo(this string) instead of the array overload.

Related problems:
  - Data.Inner and Vendor.Inner can match through the bare string "Inner".
  - Constrained Foo<T>(this T) where T : ITagged is treated as universally applicable.
  - List<int> and List<string> become indistinguishable.

#### Namespace lookup order does not follow C# scope rules

The resolver puts all enclosing namespaces in one tier and all imports in another, always preferring any enclosing candidate. C# instead searches scope-by-scope from the innermost namespace: declarations and imports at that scope are considered before moving outward.

I reproduced a call inside `A.B` where using `X`; correctly makes `X.Foo` win, but Gortex selects an extension declared in the outer namespace `A`. The problem is split between the file-wide using collection https://github.com/zzet/gortex/blob/8adb9605c44bc9093512ee7028aa169e5188cd01/internal/parser/languages/csharp.go#L1105-L1160 and flattened visibility narrowing https://github.com/zzet/gortex/blob/8adb9605c44bc9093512ee7028aa169e5188cd01/internal/resolver/csharp_extension.go#L565-L609

A using inside namespace `A` is also incorrectly made visible to calls inside a sibling namespace `B` in the same file.

#### Incremental reuse preserves stale extension bindings

The reuse key  https://github.com/zzet/gortex/blob/8adb9605c44bc9093512ee7028aa169e5188cd01/internal/indexer/incremental_resolve.go#L30-L34 includes caller, edge kind, receiver name, and method name—but no visibility fingerprint.

Changing:

```csharp
using LibA;
```
to:
```csharp
using LibB;
```
can therefore restore the old resolved target  https://github.com/zzet/gortex/blob/8adb9605c44bc9093512ee7028aa169e5188cd01/internal/indexer/incremental_resolve.go#L175-L199 before the resolver runs. Editing a sibling global using file similarly does not enqueue dependent callers for re-resolution.

#### One changed C# file can add another workspace-wide scan

The first visibility lookup in each pass builds the global-using index by iterating every KindFile node in the store under the resolver and C# cache locks. The cache is then cleared at every incremental resolve, so it cannot amortize across saves. Global-using index construction  https://github.com/zzet/gortex/blob/8adb9605c44bc9093512ee7028aa169e5188cd01/internal/resolver/csharp_ns_narrow.go#L134-L174

On SQLite, NodesByKind materializes all matching full node rows—including metadata—before yielding. This directly adds more of the scan/meta-decoding workload that dominated the daemon memory burst we previously profiled.

### Additional merge concerns

  - global using static is detected but remains file-local.
  - The C# extractor version bump only causes automatic restaging when Merkle mode is enabled; default persisted stores remain on v2 until a manual/full reindex.
  - Local-owner discovery adds repeated O(locals × functions) scans during extraction.
  - The existing benchmark passes because its no-pending path exits before triggering the new global file scan.

----
 
I would keep the PR open and request a focused revision covering:

  - Instance-member precedence before extension lookup.
  - Lossless receiver type identity and shape.
  - Scope-aware using metadata and exact scope-by-scope lookup.
  - Revalidation or exclusion of extension edges from incremental reuse.
  - Compilation-unit invalidation for global-using changes.
  - A persistent, project-scoped global-using index.
  - An upgrade path for default non-Merkle stores.
  - Regression tests for the four misbindings above plus a benchmark that actually enters visibility lookup.

pbednarcik:
Thank you for the thorough review and the reproducible cases — all four misbindings are confirmed on my side. I'll revise per your list: instance-member precedence first, then lossless receiver shape, scope-by-scope using resolution, reuse invalidation, and the persistent global-using index. Each of your repros becomes a regression test.

To make revisions like this faster to verify, I've also started a small purpose-built C# fixture repository: every construct is hand-counted, with a scripted battery asserting exact node/edge counts and specific binding outcomes — including known-gap cells that flip visibly when a fix lands. It indexes in seconds, so it gives a much tighter feedback loop than the large production solution I validate against (minutes per cold reindex). Your four cases will be encoded there as well, alongside the unit tests.

Will update here when the revision is ready.

pbednarcik:
All five points are addressed, each pinned first by a test reproducing your example against the old behavior:

**1. Instance-member precedence** (0f854ac3): a same-name member anywhere in the receiver's declared hierarchy refuses the extension bind — `d.Foo()` with `Base.Foo()` declared never binds `E.Foo(this Base)`. In-graph proof only; builtin receivers stay exempt.

**2. Receiver/param type shape** (7cba8678): extraction stamps full shape additively (`receiver_shape`, `this_param_shape`, `method_type_params`, `this_param_constraints`); the existing stamps are untouched. `string[]` no longer fits `this string`, `List<int>` no longer fits `this List<string>`, and `where T : ITagged` excludes a receiver only on full in-repo proof — an incomplete hierarchy keeps the candidate rather than guessing.

**3. Scope-by-scope using visibility** (562fd0ed): usings are stamped with their declaring namespace scope, and lookup walks the caller's namespace levels innermost-out, merging each level's declarations with that level's own usings. Both of your scope cases are locked as tests.

**4. Incremental reuse across visibility changes** (96b66b31): a changed using stamp drops the reuse snapshot and the prior-unresolved skip on all three incremental paths (the metadata-only refresh falls back to structural). A `global using` edit — or deleting that file, or a csproj — restubs and re-resolves the compilation unit's dependents.

**5. Per-save global-using scan** (0a41e128): the index is seeded by one workspace scan and maintained by reconciling exactly the files a scoped resolve names; wholesale invalidation moved to the ResolveAll entry. A new benchmark that actually enters the visibility lookup: 31.3ms → 25.8ms per save on a 100k-file graph.

Smaller items (ffb25652): `global using static` is compilation-scoped through the same unit index; a bumped extractor version now restages its language on non-Merkle stores too (one full-tree pass, then re-stamped); per-local owner attribution is a sorted interval lookup instead of a linear scan per query.

**One more, found while validating end-to-end** (bec99e2e): the fixture's refusal probes went green on the cold pass, then flipped to bound minutes later — the cross-repo pass's name-only fallback was binding the deliberately-unresolved member stubs by name, with no confidence and no origin. C# member calls no longer take that fallback: the main binder's refusal is a verdict (the exact-receiver cross-repo tiers stay available). On the production codebase that no-provenance signature is now absent and the refusal probes stay refused across later passes. Two adjacent no-provenance populations remain out of scope — unqualified calls through the function-call fallback, and same-class name binds on invisible receivers — both want receiver anchoring from field/property types, which is the next piece of work either way.

Follow-up commits round out the range (e336ca9f, 01449de1, db7e46d8), each test-first: constructors carry the same scope stamp as methods; the `?` shape veto applies only to builtin value types (`Crate?` binds `this Crate`; `int?` still refuses `this int`); a `var` initializer's shape comes from the outermost creation, not a nested initializer's `new`; the point-4 re-pricing also covers the batched deletion path and the deferred-save fallback; and the verdict gate covers restubbed member calls (they travel as bare names), keys on exactly `csharp` (a razor caller never enters the member gates, so it keeps its fallback), and fails closed on a missing caller node.

Known simplifications left open, for the record: the member-claims gate is name-based, not applicability-based; adding a csproj doesn't re-price binds made under the old unit boundaries (deleting one does); type references aren't re-priced on a global-using swap; argument-position generic variance is treated as invariance; and multi-repo mode constructs the master resolver per catch-up pass, so the persistent-index win applies per resolver lifetime there.

Suites are at their pre-existing baselines with no new failures. Your six repro cases run as spec tests in `internal/resolver/csharp_extension_spec_test.go` and as store-level fixture cells, including the three refusal probes that must stay unresolved. Re-validated cold on the production codebase: 1,031 extension binds — parity with the count before this round, now with every bind through the precedence/shape/scope gates; the per-module DI-configuration battery is line-identical; the fixture battery passes twice, minutes apart, with no verdict drift.

zzet:
@pbednarcik can you please take a look at linter outcome?

pbednarcik:
@zzet - should be fixed now.

internal/indexer/indexer.go.3599427492050421765
```diff
@@ -0,0 +1,8637 @@
+package indexer
```
zzet: @pbednarcik I guess, it's the temporal file which should be removed, isn't it?

internal/resolver/resolver.go.8052787374988375801
```diff
@@ -0,0 +1,4928 @@
+package resolver
```
zzet: @pbednarcik looks like the temporal file too.

internal/resolver/csharp_extension.go
```diff
@@ -56,36 +80,1123 @@ func (r *Resolver) tryBindCSharpExtension(e *graph.Edge, methodName, receiverTyp
...
+		}
+	}
+	return false
+}
+
+// csharpMemberNamesOf memoizes a type's declared member-name set per
+// pass — GetInEdges returns every reference to a popular receiver type,
+// and the claims gate runs once per candidate-bearing call edge.
+func (r *Resolver) csharpMemberNamesOf(typeID string) map[string]struct{} {
+	if names, ok := r.csharpMemberNamesByType[typeID]; ok {
```
zzet: There reading and writing the map without lock (mutex), which is reason for race. It's reachable from the parallel path: `resolveEdge` runs on `runtime.NumCPU()` goroutines (resolver.go:883), and both `resolveMethodCall` (:4103) and `resolveFunctionCall` (:3527) reach the memo via `tryBindCSharpExtension` → `csharpInstanceMemberClaims`. I confirmed with a -race probe:
```
WARNING: DATA RACE
  Write at 0x00c00068c5a0 by goroutine 28:
    ...csharpMemberNamesOf() csharp_extension.go:411
  Previous read at 0x00c00068c5a0 by goroutine 27:
    ...csharpMemberNamesOf() csharp_extension.go:396
```

In production that's a potential fatal error: concurrent map read and map write — an unrecoverable daemon crash — on any multi-worker C# pass with ≥2 typed-receiver extension calls. It's nondeterministic, which is why the 8k-file validation run and CI both got away with it. Fix mirrors `csharpTypeNodesNamed` exactly: `RLock` the read, `Lock` the write.

internal/graph/store_sqlite/meta_json.go
```diff
@@ -126,6 +126,7 @@ var metaInt64Keys = map[string]bool{
 var metaStringSliceKeys = map[string]bool{
 	"path_param_names": true, "query_params": true, "status_codes": true,
 	"notes": true, "methods": true, "arg_names": true, "repos": true,
+	"usings": true, "global_usings": true, "using_static": true,
```
zzet, line 129: Worth adding `scoped_usings` and `global_using_static` to the list

zzet (changes requested):
@pbednarcik I found a few small issues (two are important to address)

internal/indexer/indexer.go.3599427492050421765
pbednarcik: Yes, good catch.

internal/resolver/resolver.go.8052787374988375801
pbednarcik: Yes, same as above

internal/graph/store_sqlite/meta_json.go
pbednarcik: Definitely.

pbednarcik:
All four addressed (d1a978d0, cf0bc9f1):

- **Data race**: confirmed — the member-claims memo was the one map in the file not under `csharpNSMu`; my own comment even said "same lifetime and lock" while taking neither. Now mirrors `csharpTypeNodesNamed` exactly (RLock read / Lock write; the published set is never mutated after). Locked in by a parallel-workers test that fails under `-race` on the old code, and the full C# resolver set now runs `-race`-clean.
- **`scoped_usings` / `global_using_static`**: added to `metaStringSliceKeys` — without them the JSON decode path handed the visibility rules `[]any` and the file's usings silently vanished. A round-trip test now pins all five C# visibility keys as `[]string`.
- **Stray files**: leftovers from some local stash juggling while reordering the fix stack — removed, and the pattern is in my local excludes now so nothing like that can sneak into a commit again 🤞

internal/resolver/csharp_extension.go
pbednarcik: Fixed.

zzet (approved):

zzet:
@pbednarcik great job! Thank you!

PR #446, opened 2026-08-03
Layer .gitignore from the git root down to a tracked subdirectory

### LINKED ISSUE #419: .gitignore excludes are silently skipped when the tracked root is a subdirectory of the git repository
## Summary

`EffectiveExclude` reads `.gitignore` only at the tracked root itself (`statGitignore(repoPath)` in `internal/config/manager.go`). Git applies ignore files from the repository root downward, so in the common monorepo layout — git root at `repo/`, gortex tracking `repo/projects/App` — the entire gitignore layer silently disappears. Nothing warns; the index just quietly fills with everything the ignore file was supposed to keep out.

For .NET solutions this is compounded by `internal/excludes/builtin.go`: `Builtin` covers JS (`node_modules/`), Python (`.venv/`, `__pycache__/`), JVM (`.gradle/`, `target/`), iOS (`Pods/`) — but has no Visual Studio / MSBuild entries at all. When the gitignore layer is missing, nothing catches `.vs/`, `bin/`, or `obj/`.

## Observed on a real solution

Production .NET solution, ~10,400 indexed files, tracked root one level below the git root (whose `.gitignore` has standard `.vs/` and `[Oo]bj/` entries — verified with `git check-ignore -v`):

- `.vs/<Solution>/v17/HierarchyCache.v1.txt` and the `v18` copy indexed as first-class doc nodes
- 50+ `obj/**/project.assets.json` indexed (hit the result cap)
- generated `obj/{Debug,Release}/net8.0/*.AssemblyInfo.cs` indexed as real C# source — duplicate generated symbols per build configuration alongside the hand-written ones
- `obj/**/*.csproj.FileListAbsolute.txt` indexed as doc nodes

Beyond wasted index size, this junk actively breaks search recall — the ranking side of that is addressed in PR #416, but the two-line version: those `.vs`/`obj` text files are stuffed with source file paths, so they outrank real classes on exactly the tokens .NET naming conventions use. Keeping them out of the index in the first place is the other half of the defense.

## Repro (minimal)

```
git init repo && cd repo
echo "junk/" > .gitignore
mkdir -p projects/app/junk
echo "class Real {}" > projects/app/Real.cs
echo "noise" > projects/app/junk/noise.txt
gortex track projects/app   # track the subdirectory, not the git root
# index → junk/noise.txt is indexed despite being git-ignored
```

## Suggested fix

1. Discover the git root by walking up from the tracked root to the nearest `.git`, and layer every `.gitignore` on the path from the git root down to the tracked root (patterns re-anchored relative to the tracked root, matching git semantics).
2. Independently, add the standard VS/MSBuild set to `excludes.Builtin` as defense-in-depth, same spirit as the existing `Pods/`/`.gradle/` entries: `.vs/`, `bin/`, `obj/`, `TestResults/`. (`bin/` and `obj/` are effectively never first-party source directories in .NET repos; a re-include via `!pattern` remains available for the exceptions.)

Happy to send a PR for either or both halves if that's welcome.

zzet:
Fixes #419.

## Validation

Reproduced first, as a failing test: with the git root at `repo/` and the tracked root at `repo/projects/app`, `ConfigManager.EffectiveExclude` returned the builtin baseline and nothing else — every pattern in `repo/.gitignore` was absent. `statGitignore(repoPath)` only ever looked at the tracked root itself, so for any repo tracked below its git root the whole ignore layer silently disappeared.

## 1. Layer the whole `.gitignore` chain (`internal/config`)

The chain now climbs to the nearest enclosing `.git` — a directory in a clone, a *file* in a linked worktree or submodule — and layers every `.gitignore` from there down to the tracked root, outermost first so a deeper file still overrides a shallower one (go-gitignore is last-match-wins, so a `!re-include` next to your code beats the repository-wide rule).

Ancestor patterns are re-anchored onto the tracked root, mirroring git's own anchored/unanchored split:

| ancestor pattern (git root, tracked root `projects/app`) | becomes |
|---|---|
| `junk/`, `*.log` — unanchored, matches at any depth | carried over verbatim |
| `/projects/app/generated/` | `/generated/` |
| `/projects/other/` — sibling subtree | dropped |
| `/*/app/packed/` — wildcard segment matching the real name | `/packed/` |
| `projects/**/artifacts/` | `**/artifacts/` |
| `/projects/app/` — swallows the tracked root | dropped |

That last row is a deliberate call: the user pointed Gortex at that directory explicitly, and honouring an ancestor ignore of it would empty the index of the very repo they asked for.

`**` matches zero or more segments, so a pattern can reach into the tracked root at more than one depth. `anchoredRemainders` resolves this with a reachability sweep over `(pattern segment, sub-path segment)` — `**` modelled as a self-loop — which keeps a pattern packed with `**` polynomial instead of exponentially branching, and drops a remainder that a `**`-led sibling already subsumes.

Outside a git repository nothing changes: only the tracked root's own file applies, since ancestor files outside a repository are not something git would honour either.

### Keeping the hot path hot

`EffectiveExclude` runs on a firehose of calls, and the upward `.git` walk is an `os.Lstat` per level — for a tracked root outside any repository, all the way to `/`. Discovering the chain on every call cost 6.7× and 14× the allocations:

| layout | before | naive chain | shipped |
|---|---|---|---|
| tracked root == git root | 2.0 µs / 3 allocs | 9.5 µs / 7 allocs | 2.3 µs / 4 allocs |
| two levels below git root | — | 30.8 µs / 25 allocs | 6.5 µs / 12 allocs |
| no repository above (walk to `/`) | 2.0 µs / 3 allocs | 37.0 µs / 42 allocs | 2.3 µs / 4 allocs |

The chain's *shape* — which directories git would read a `.gitignore` from — is memoized per tracked root, since it moves only when a `.git` appears or disappears above it. Every call still re-stats each member, so an edit anywhere in the chain is picked up immediately; `publishWorkspaceConfig` forgets the shape whenever a repo's path is published (track, reindex, config refresh), which is where a newly created git root gets noticed. The parse cache is now keyed by the `.gitignore`'s own directory, so sibling tracked roots under one git root share their common ancestors' parse.

## 2. Visual Studio / MSBuild artifacts in `excludes.Builtin`

Defense-in-depth for the repo whose ignore layer is missing entirely: `.vs/`, `TestResults/`, and what MSBuild actually writes under `obj/` — the per-configuration trees holding regenerated C# and the NuGet restore metadata.

`obj/` and `bin/` are **not** excluded wholesale, which is where this departs from the issue's suggestion. Both are first-party source outside .NET — Go's own toolchain ships `cmd/internal/obj`, Rails commits `bin/rails`, npm packages commit `bin/cli.js` — and a silent blanket drop of real source is the same failure this list exists to prevent, just pointed the other way. Naming what MSBuild writes covers every artifact the issue reports while leaving a first-party `obj` package indexed; a .NET repo with a normal `.gitignore` gets `bin/` from part 1 regardless.

## Testing

- Chain behaviour: monorepo repro, the .NET layout, re-anchoring, sibling-drop, tracked-root-swallow, `**`, wildcard segments, deeper-overrides-ancestor, intermediate files, no-git-root, nested repo, `.git`-as-a-file, `respect_gitignore: false`, sibling tracked roots.
- Cache: ancestor edit invalidates and re-reads exactly the changed file, a newly added intermediate `.gitignore` reshapes the chain, republishing rediscovers a git root that appeared mid-session.
- Unit table over `reanchorPattern`, plus a `**`-saturation case pinning the sweep's bound.
- End-to-end through the real index walk: a repo tracked below its git root indexes its source and keeps `junk/` and `obj/` out of the graph.
- `excludes.Builtin`: the MSBuild artifacts match; `cmd/internal/obj/link.go`, `bin/rails`, `bin/cli.js` do not.
- `go build ./...`, `go vet ./...`, `golangci-lint run ./internal/... ./cmd/...` clean; `go test -race` green on `internal/config`, `internal/excludes`, and `internal/indexer`. A full-tree `go test -race ./...` is still running at the time of opening — I will report the result here.

## Known gaps

- `.git/info/exclude` and `core.excludesFile` are still not read — the chain covers `.gitignore` files only.
- Per-directory `.gitignore` files *below* the tracked root remain unread (`.gortexignore` / `.ignore` / `.rgignore` already are, via `excludes.Hierarchical`).
- `gortex config exclude list` prints the builtin/global/repo/workspace layers but has never shown the `.gitignore` layer; unchanged here, but it means the CLI still under-reports the effective list.
- A whitelist-style root `.gitignore` (`*` plus explicit re-includes) now applies to a tracked subdirectory the way git applies it, which can exclude more than before. That is the intended semantics; `respect_gitignore: false` or an `include:` entry opts out.

### DIFF
--- a/docs/multi-repo.md
+++ b/docs/multi-repo.md
@@ -21,7 +21,9 @@ Two-tier config hierarchy:
 - **Global config** (`~/.gortex/config.yaml`) — projects, repo lists, active project, reference tags
 - **Workspace config** (`.gortex.yaml` per repo) — guards, excludes, local overrides
 
-Excludes are layered — builtin → repo's own `.gitignore` → global → per-repo entry → workspace — with gitignore semantics. The repo's `.gitignore` is respected by default so you don't have to re-declare entries already curated for git; opt out per-workspace with `respect_gitignore: false` in `.gortex.yaml`. Use `!pattern` in a later layer to re-include something an earlier layer excluded. Beyond `.gitignore`, the index walk also honors per-directory `.gortexignore` files (Gortex's own ignore file, a sibling to `.gitignore`) and ripgrep's `.ignore` / `.rgignore` — each scoped to the directory that contains it.
+Excludes are layered — builtin → the repo's `.gitignore` chain → global → per-repo entry → workspace — with gitignore semantics. `.gitignore` is respected by default so you don't have to re-declare entries already curated for git; opt out per-workspace with `respect_gitignore: false` in `.gortex.yaml`. Use `!pattern` in a later layer to re-include something an earlier layer excluded. Beyond `.gitignore`, the index walk also honors per-directory `.gortexignore` files (Gortex's own ignore file, a sibling to `.gitignore`) and ripgrep's `.ignore` / `.rgignore` — each scoped to the directory that contains it.
+
+When a tracked root sits below its git root — the monorepo case, `gortex track repo/projects/App` with the repository at `repo/` — every `.gitignore` from the git root down to the tracked root applies, as it would for git, with ancestor patterns re-anchored onto the tracked root. A deeper file overrides a shallower one, so a `!pattern` next to your code still wins over the repository-wide rule. Ancestor patterns that describe a sibling subtree are dropped, and so is one that would ignore the tracked root itself: you asked Gortex to index that directory explicitly.
 
 ```yaml
 # ~/.gortex/config.yaml
--- a/internal/config/gitignore.go
+++ b/internal/config/gitignore.go
@@ -8,27 +8,30 @@ import (
 	"unicode/utf8"
 )
 
-// loadRepoGitignore reads the `.gitignore` file at the repo root and
-// returns its entries as gitignore-syntax patterns ready to feed into
-// the excludes matcher. Blank lines and `#` comments are stripped.
+// loadRepoGitignore reads the `.gitignore` file in dir and returns its
+// entries as gitignore-syntax patterns ready to feed into the excludes
+// matcher. Blank lines and `#` comments are stripped.
 //
-// We deliberately do NOT walk the per-directory hierarchy here:
-//   - The repo-root file already covers ~95% of the value (build
-//     output, dependency caches, generated code).
-//   - Per-directory `.gitignore` semantics (path anchoring relative to
-//     the file's location) would require either nested matchers or a
-//     pattern-rewrite pass; both add complexity for marginal coverage.
+// Callers pass one directory at a time. gitignoreChain assembles the
+// upward chain — every `.gitignore` from the enclosing git root down to
+// the tracked root — and re-anchors ancestor patterns; this function only
+// parses one file.
+//
+// We deliberately do NOT walk *downward* into per-directory `.gitignore`
+// files here:
+//   - The files at and above the tracked root already cover ~95% of the
+//     value (build output, dependency caches, generated code).
 //   - Users with sub-directory ignores can list them in `.gortex.yaml`
-//     `excludes` until/unless we add hierarchy support.
+//     `excludes` until/unless we add downward hierarchy support.
 //
 // Returns nil when the file is absent or unreadable — gitignore reading
 // is a convenience, never a hard requirement, so a missing or
 // permission-denied file silently no-ops.
-func loadRepoGitignore(repoPath string) []string {
-	if repoPath == "" {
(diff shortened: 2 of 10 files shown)


zzet:
Full-tree `go test -race ./...` finished green — no failures.

Two commits landed after that run started (a `slices.Contains` cleanup in `gitignore_chain.go` and the end-to-end indexer test), so `internal/config`, `internal/excludes`, and `internal/indexer` were re-run with `-race` against the final tree: all three pass.

PR #466, opened 2026-08-05
csharp: extract this./base.-qualified member calls

pbednarcik:
`this.Foo()` and `base.Foo()` currently emit no call edge at all — both member-call query patterns capture the receiver with `expression: (_)`, but `this` and `base` are **anonymous tokens** in tree-sitter-c-sharp, and a named-node wildcard never matches them. Every qualified self/base call in a C# codebase silently vanishes from the graph (noticeable on `base.OnActionExecuting(...)`-style overrides, which are everywhere in ASP.NET code).

**Fix** (extraction only, resolver untouched):

- Two new query alternations match the anonymous token literally (`(member_access_expression "this" name: (identifier))` and the `base` twin).
- The receiver type is resolved at capture time from the enclosing declaration — `this` → the innermost type's name, `base` → its declared base class (discrimination mirrors `emitCSharpBaseList`: first base-list entry that is a ctor-base or not I-prefixed; structs and interface-only bases stamp nothing and the call still emits as a plain `member_call`).
- The stamped `receiver_type` then rides the existing receiver-typed member binding path, so `this.Foo()` binds the hiding member and `base.Foo()` the base member with no new resolver machinery.

**Tests**: extractor spec (edges + `receiver_type` stamps for both qualifiers, `new`-hiding case) and an end-to-end resolver spec (`this.Foo()` → hiding member, `base.Foo()` → base member). Both watched red before the fix. `internal/parser/languages` fully green; `internal/resolver` and `internal/indexer` at their pre-existing Windows baselines with zero new failures.

The second commit bumps the `csharp` extractor version 3→4 so existing stores restage `.cs` files and grow the new edges. Validated live on a counted C# fixture repo via exactly that path — binary swap, no store wipe, salt-triggered restage — with both call shapes binding their compiler-correct targets (`this.X()` → the `new`-hiding member, `base.X()` → the base member) at 0.95 with `receiver_type` stamped.

### DIFF
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -32,7 +32,7 @@ var extractorVersions = map[string]int{
 	//   "go": 2,
 	"c":      generatedParserProjectionPolicyVersion, // generated parser projection covers all strictly detected table sizes
 	"php":    2,                                      // class/interface inheritance now emits typed structural edges
-	"csharp": 3,                                      // scoped/global using + receiver/param shape stamps for extension binding (was: null-conditional call edges; caseless reference forms)
+	"csharp": 4,                                      // this./base.-qualified call edges (was: scoped/global using + receiver/param shape stamps; null-conditional call edges)
 }
 
 // extractorSaltExtLang maps a lower-case file extension to the language
--- a/internal/parser/languages/csharp.go
+++ b/internal/parser/languages/csharp.go
@@ -85,6 +85,16 @@ const qCSharpAll = `
       (member_binding_expression
         name: (identifier) @callm.method))) @callm.expr
 
+  (invocation_expression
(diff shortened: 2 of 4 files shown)


zzet (approved):
Two adjacent gaps found (inspired by PR, not directly related to PR)

- `this?.Foo()` / `base?.Foo()` emits nothing — the conditional-access pattern has the identical condition: (_) blindness. Rare in practice (this is never null), but it's the same bug left half-fixed.

- C# records emit no member nodes at all. `public record Tower(int N) { public int Chime() {...} }` produces a type node and its extends edge, and nothing else — every method in every record is invisible to the graph, so no call edges leave them. Records are idiomatic C# 9+, so this is a substantially bigger gap than the one this PR closes.

PR #473, opened 2026-08-05
Decommission the in-memory graph backend; SQLite becomes the only store

zzet:
## What

Retires the in-memory graph **backend** and makes the SQLite store the only persistence layer. `*graph.Graph` survives in exactly two demoted roles: the indexer's cold-index staging shadow, and the fixture for the existing test corpus — a new fence test (`internal/graph/newfence_test.go`, AST-based) pins `graph.New()` to the sanctioned staging sites so it cannot drift back into service as a backend.

Why the staging shadow stays: measured on this repository (3,994 files, 1.56M edges), a cold index into sqlite takes **6m34s with the shadow** and **did not finish in 35 minutes without it** (killed mid-resolver; per-call disk reads). On rails (4.5k files): 43s vs 99s. The shadow is load-bearing; the backend option was not.

### Removed
- `--backend memory` selection in `serverstack.OpenBackend` — the memory spellings now fail fast with guidance (`--backend sqlite --backend-path <path>` for an ephemeral store). An empty backend name still defaults to sqlite; silently falling back would have written an "ephemeral" store into `~/.gortex/store/store.sqlite`.
- Both graph-snapshot subsystems (dead code on the sqlite default — every entry point was gated on `.(*graph.Graph)`): the daemon's gob+gzip snapshot (`cmd/gortex/daemon_snapshot.go`, ~1.3k LOC with its consumers) and the one-shot commit-keyed `FileStore` snapshot in `cmd/gortex/mcp.go`, plus the graph half of `internal/persistence`.
- The boot shape-degradation guard — it only ever armed from the gob snapshot load, so it has been a no-op on the sqlite default all along. A store-sourced replacement can be built later from the persisted repo index state.
- The last production `.(*graph.Graph)` behaviour forks (daemon warmup, warm-restart routing in `multi.go`, `explain_change_impact`).

### Converted to real stores
- `gortex mcp` without a daemon now runs on a private per-process temp sqlite store (guarded so it can never touch the daemon's shared store; stale temp stores are reaped at startup). **Accepted trade-off: without the snapshot cache, every daemonless one-shot invocation cold-indexes the repository.** `--no-cache` becomes a hidden no-op shim like the other retired MCP flags.
- Eval harnesses and the bench trio run against temp sqlite stores; the recall harness's lexical row now measures the store-native FTS the daemon actually serves.
- `gortex init` intake dry-run no longer constructs a store at all.
- `pkg/gortex` (**breaking**): `New(opts ...Option) (*Engine, error)`, `Close() error` required, `WithStorePath` added. The break is deliberate — a compile error instead of a silently leaked sqlite handle + WAL goroutine per engine.

### Fixed on the way (pre-existing sqlite bugs this removal would have exposed)
- The direct (non-shadow) cold path never wrote `symbol_fts` — every large-repo cold index shipped with degraded symbol search. Now populated on both the direct and streaming paths, with a regression test.
- Seven sqlite cursor helpers returned silently truncated results on mid-scan driver errors (`rows.Err()` unchecked / swallowed `Query` errors).
- Attribute-only `ReindexEdges` entries were dropped by sqlite but persisted by the memory store; converged on persisting, with the receipt preload aligned to match.

### Conformance
The store contract now specifies what was previously implicit, green on both implementations: the persist round-trip, detached-vs-aliasing read semantics (declared per backend), adjacency ordering (sqlite's `ORDER BY` is plan-locked as index-served, including the degraded plan during bulk load), reentrancy from iterator callbacks, and mutation receipts.

## Compatibility notes

- `docs/versioning.md` scopes SemVer to the MCP tool API and the CLI. This PR rejects a **value** of a kept flag (not a flag removal), shims the one removed flag, and keeps every MCP tool signature. Two response-payload details changed: `explain_change_impact` no longer emits the community-`couplings` detail (the dedicated coupling analysis covers it; the zero-impact warning/caveat annotations are preserved and now armed on every backend), and the `snapshot_loaded` readiness phase (name kept) no longer carries a `snapshot_repos` count.
- `BENCHMARK.md`: the fixture row was re-measured against sqlite; the token-efficiency and daemon-latency tables are marked as measured on the retired backend, pending re-measurement.

## Follow-ups (deliberately out of scope)
- Partition mixed reindex batches so attribute-only entries stop disqualifying the resolved-conversion UPDATE fast path.
- Liveness guard for terminal-stamp clears racing file eviction.
- Delete the now-caller-less vector snapshot API (`SetSkipVectorBuild` / `ImportVectorIndex` / `ExportVectorIndex`).
- `rows.Err()` sweep of the remaining cursor family (analysis/vector/purge helpers).
- A bulk-staged direct cold path — the prerequisite for ever deleting the staging type itself.

## Validation
- `go test -race -count=1 ./...` green (142 packages), before and after the review-fix commits. The one flake it surfaced — the perf fixture's latency budgets, still calibrated for the retired backend — was recalibrated against measured sqlite figures with ~13-18x headroom.
- `golangci-lint run` — 0 issues.
- Dual-backend conformance suite, the staging fence (proven to catch aliased-import evasion), and new regression tests each proven to fail without their fix.
- Three independent adversarial review passes over the full diff; every confirmed finding fixed or listed above as a follow-up.


cmd/gortex/mcp.go
```diff
@@ -81,21 +84,85 @@ func init() {
...
+func reapStaleEmbeddedStores(logger *zap.Logger) {
+	matches, err := filepath.Glob(filepath.Join(os.TempDir(), embeddedStoreDirPattern))
+	if err != nil {
+		logger.Debug("mcp: could not scan for stale embedded stores", zap.Error(err))
+		return
+	}
+	cutoff := time.Now().Add(-staleEmbeddedStoreTTL)
+	for _, dir := range matches {
+		info, statErr := os.Stat(dir)
+		if statErr != nil || !info.IsDir() || info.ModTime().After(cutoff) {
```
peterkc, line 158: **[P1] Protect live embedded stores before reaping**

Directory age does not prove that a store is orphaned. SQLite activity updates files inside this directory without necessarily updating the directory mtime, so another `gortex mcp` launch can remove a store that is still open after 48 hours.

I reproduced this by opening the SQLite store, aging its directory, and running the reaper; the live directory was deleted.

Please hold an advisory lock for each embedded store's lifetime and reap only after acquiring that lock non-blockingly. A regression test should keep an aged store open and verify that the reaper leaves it usable.

internal/indexer/indexer.go
```diff
@@ -765,6 +765,101 @@ func ftsTokensFor(n *graph.Node, projectName string) string {
...
+
+	items := make([]graph.SymbolFTSItem, 0, symbolFTSDirectChunkRows)
+	var pending uint64
+	written := 0
+	var flushErr error
+	flush := func() bool {
+		if len(items) == 0 {
+			return true
+		}
+		if err := batcher.BatchUpsertSymbolFTS(items); err != nil {
```
peterkc: **[P1] Make the FTS replacement atomic**

`ResetSymbolFTS` commits before the independently committed append batches. If a later batch fails, the old complete corpus has already gone and the new corpus remains partially visible.

With 2,500 seeded documents and an injected failure on batch two, the function returned an error but left only 2,048 documents in the authoritative corpus. Search then has durable false negatives.

Please stage the replacement under a generation and publish it atomically, or keep reset and append operations in one rollback-capable transaction. The regression test should assert that a failed rebuild preserves every document from the previous corpus.

internal/graph/store_sqlite/store.go
```diff
@@ -2382,9 +2418,17 @@ func (s *Store) EdgesWithUnresolvedTarget() iter.Seq[*graph.Edge] {
 // deadlocking on the MaxOpenConns=1 pool. Companion to the existing
 // queryEdges helper that takes a *sql.Stmt; this one takes a raw
 // SQL string so the predicate iterators can pass inline queries.
+//
+// Statement-level and cursor-level failures go through panicOnFatal: a
+// teardown-race read still degrades to what was materialised, but a real
+// storage failure is raised instead of being handed back as an empty or short
+// slice the caller cannot tell from a genuinely small result. A single row
+// that will not decode stays non-fatal and is skipped — one unreadable row
```
peterkc: **[P2] Surface row decode failures as storage failures**

`rows.Err()` catches driver iteration failures, but `scanEdgeCursor` and the node/light variants return decode errors directly. These loops still skip those rows and return a plausible truncated result.

I corrupted one edge metadata blob and `queryEdgesSQL` returned normally with the edge missing. That conflicts with the new contract that a short result must not look successful.

Please route scan/decode errors through `panicOnFatal` and stop the scan. Tests should cover corrupt edge metadata, node metadata, and a light-projection scalar.

cmd/gortex/repos_cmd.go
```diff
@@ -150,8 +133,8 @@ func runRepos(cmd *cobra.Command, _ []string) error {
 // (the SQLite repo_index_state table) keyed by repo prefix. It opens the
 // backend store read-only so it is safe to run while a daemon holds the
 // same store. Any failure (no store yet, unreadable cache) degrades to an
-// empty map so `gortex repos` falls back to the snapshot store rather than
-// erroring out.
+// empty map so `gortex repos` reports every repo as never indexed rather
```
peterkc: **[P2] Do not translate corrupt authoritative state into 'never indexed'**

Now that the snapshot fallback is removed, every `ReadRepoIndexStates` error becomes a successful response saying that every repository was never indexed. Corruption, permission failures, and schema errors are therefore indistinguishable from an absent store.

I confirmed this with invalid SQLite bytes: `runRepos` returned success and reported the repository as unindexed.

Please tolerate only a genuinely absent store or table and propagate other errors. A corrupt-store regression test should require `runRepos` to fail.

cmd/gortex/repos_cmd.go
```diff
@@ -150,8 +133,8 @@ func runRepos(cmd *cobra.Command, _ []string) error {
...
+// than erroring out.
 func loadRepoIndexStates() map[string]graph.RepoIndexState {
 	path := reposBackendPath
```
peterkc: **[P2] Read the backend path the daemon actually owns**

`daemon start --backend-path` can place the authoritative store anywhere, but `gortex repos` always reads `reposBackendPath` or the platform default. There is no production route that forwards or discovers the daemon's custom path.

I seeded a fresh `repo_index_state` row only in the daemon's custom store and `repos` still reported the repository as unindexed.

Please persist or query the active daemon path, or expose and document a matching `repos --backend-path` option. The regression test should exercise that production routing path.

cmd/gortex/daemon.go
```diff
@@ -242,27 +241,20 @@ func runDaemonStart(cmd *cobra.Command, _ []string) error {
...
-		// restart reads everything it needs from the on-disk store —
-		// no gob+gzip round-trip required.
+		// Nothing else has to be serialized here: per-file mtimes live in
+		// the FileMtime sidecar table, contract records ride on
+		// KindContract.Meta, and the vector index is persisted by the
+		// backend itself. Warm restart reads everything it needs straight
+		// from the on-disk store.
 		// Run the shared stack's teardown chain — flushes the savings
 		// store and closes the backend handle (checkpointing the sqlite
 		// WAL) so the daemon shuts down cleanly.
```
peterkc: **[P2] Run shared teardown on signal exits**

The SQLite close/checkpoint chain is installed only in `controller.Shutdown`. The daemon server's signal handler calls `Server.Shutdown` directly, after which `Serve` returns without closing `state.shared`.

As a result, SIGINT and SIGTERM bypass watcher shutdown, side-store flushing, and the final WAL checkpoint. With SQLite now the only runtime backend, this affects every signal-managed daemon.

Please make teardown idempotent and run it for every exit, preferably through a defer installed immediately after state construction. A regression test should drive the signal path and verify that watcher stop, shared close, and checkpoint each run exactly once.

peterkc (commented):
I reviewed the complete SQLite-only change at the current head. The overall direction is coherent, but I found two P1 and four P2 correctness or lifecycle gaps.

Five findings have focused failing reproductions. The signal-shutdown finding was confirmed through execution-path inspection and independent review. I recommend resolving these before merge.

The existing focused tests, race tests, vet, diff check, and GitHub checks remain green. Details and suggested regression tests are inline.

cmd/gortex/mcp.go
zzet: Fixed in 61203d47. Each embedded store now holds a flock advisory lock on <dir>/store.lock for its process lifetime; the reaper removes a candidate only after non-blockingly acquiring that lock, with age kept as a cheap pre-filter. A store that cannot take its lock fails allocation rather than running unprotected. Regression test TestReapStaleEmbeddedStoresSpareLockedStore holds the lock on an aged dir and asserts it survives while an unlocked aged sibling is reaped — verified failing against the unfixed reaper.

internal/indexer/indexer.go
zzet: Fixed in 5a3190a1. Added Store.ReplaceSymbolFTS (graph.SymbolFTSRepoReplacer): wipe + all streamed chunks in a single rollback-capable transaction, producer-driven so memory stays bounded; populateSymbolFTS now uses it. TestPopulateSymbolFTSKeepsPriorCorpusWhenRebuildFails injects a mid-rebuild failure and asserts the prior corpus is fully intact — against the old reset-then-append code it fails with exactly your 2048-docs symptom. The shadow drain keeps its interleaved shape deliberately: it runs immediately after EvictRepo, so there is no prior corpus to protect there (stated in the commit body).

internal/graph/store_sqlite/store.go
zzet: Fixed in 52891666. queryEdgesSQL, queryNodesSQL and queryEdgesLightSQL now route decode errors through panicOnFatal and stop the scan (the rest of the cursor family was already fail-closed). The skip-tolerant contract encoded by TestGetNodesByQualNamesFailsClosedOnDecodeQueryAndClosedStoreErrors is retired and the test updated to expect the raise. New tests cover corrupt edge metadata, corrupt node metadata and a corrupt light-projection scalar — all verified failing before the fix.

cmd/gortex/repos_cmd.go
zzet: Fixed in 36b90f00. ReadRepoIndexStates now tolerates only a missing store file or missing repo_index_state table; every other error propagates and runRepos fails with the store path. TestRunRepos_CorruptIndexStoreIsNotAnUnindexedOne feeds invalid sqlite bytes and requires the command to fail, and also pins that a genuinely absent file still reports unindexed.

cmd/gortex/repos_cmd.go
zzet: Fixed in 64d0c5fc. The daemon now records its resolved backend path in a daemon.state.json beside the pidfile (written after Listen, removed on shutdown, ignored when its PID is dead), and gortex repos resolves explicit --backend-path flag > running daemon's recorded path > platform default — the flag is now registered and the precedence documented in the command help. TestRunRepos_FollowsTheDaemonRecordedBackendPath exercises the production route with no flag set.

cmd/gortex/daemon.go
zzet: Fixed in 4d4f683d. installDaemonTeardown builds a sync.Once-guarded teardown (watcher stop, shared close/checkpoint) installed both as the controller hook and as a defer in runDaemonStart, so the signal path and the control-socket path converge on the same idempotent chain and later callers see the run's error. TestDaemonTeardownRunsOnASignalledExit drives a real SIGTERM and asserts watcher stop + shared close each ran exactly once; sibling tests cover both paths firing and error propagation.

zzet:
Before/after timing A/B (single machine darwin/arm64, single sequential runs, isolated HOME per arm; main = 6a2905e1, branch = 12394922):

| measurement | main | branch |
|---|---|---|
| daemon warm start, rails (ready / enriched) | 1.29s / 2.69s | 1.24s / 2.71s |
| daemon cold start, rails (ready / enriched) | 30.4s / 95.7s | 21.8s / 79.1s |
| cold index (TestBackendBench, sqlite): 313-file repo | 931ms | 924ms |
| cold index: rails (4,530 files) | 51.2s | 45.3s |
| cold index: gortex source (4,061 files, 1.59M edges) | 7m18.8s | 5m29.9s |

Warm start is the number that matters day-to-day and it is dead even; store sizes are identical. Cold-index runs put the branch at parity or better everywhere — I would not claim the branch is faster from single runs (run order favors the second arm's filesystem cache), only that there is no regression. Node/edge counts match across arms per corpus (gortex corpus differs by 29 edges of 1.59M, from enrichment pick order).

The one accepted regression stands as designed: daemonless `gortex mcp` re-launches pay a full cold index (previously snapshot replay) — for a repo of the 300-file class that is under a second of indexing; rails-class repos pay tens of seconds per launch, which is the trade documented in the PR description.

PR #478, opened 2026-08-06
csharp: extract record members

pbednarcik:
Follow-up to the #466 review's second finding: members declared inside a `record` body were not extracted at all. The cause is one gate — the member emitters resolve ownership through a parent walk whose allowed-container list is `class_declaration` / `struct_declaration` / `interface_declaration`, so every method, constructor, field, and property inside a `record_declaration` hit the "unrecognised container" skip. The record's type node existed; its body was invisible.

The damage is worse than a missing edge. With no member node to bind, a call into a record member falls through to the fallback tiers and lands on a same-named member of an unrelated class — reproduced live at confidence 0.7 with `receiver_type` evidence present. A confident wrong answer, where the rest of the pipeline had done everything right.

**Fix**: add `record_declaration` to the four owner lists. `record struct` is the same node type in the grammar (the `struct` keyword is a child, as `csharpDeclAllowsBaseClass` already handles), so both flavors ride the same change. Everything downstream comes along for free — the function-range lookup is built from emitted nodes, so tenv typing, awaited-receiver unwrapping, and async spawn edges now work inside record methods with no further changes.

**Tests**: an extractor spec pinning member nodes for all four member kinds (method with `receiver` meta, property, field, `<init>`) plus a record-struct method and the call edge out of a record method; and a resolver spec asserting a record-typed receiver binds the record's member over a same-named decoy on an unrelated class at ≥ 0.9 — pre-fix that spec reproduces the 0.7 decoy mis-bind exactly. All tests were written first and failed against the old behaviour before the fix landed. `internal/parser/languages` fully green; `internal/resolver` and `internal/indexer` at their pre-existing Windows baselines with zero new failures.

The commit bumps the `csharp` extractor version 5→6 so existing stores restage `.cs` files. Validated live on a counted C# fixture repo after a cold rebuild: a record method's call edge out, a call into a record method, and a call into a record-struct method all bind their compiler-correct targets at 0.95 with the decoy excluded — three cells that were pinned as known-gap flipped on the first settled run.

### DIFF
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -32,7 +32,7 @@ var extractorVersions = map[string]int{
 	//   "go": 2,
 	"c":      generatedParserProjectionPolicyVersion, // generated parser projection covers all strictly detected table sizes
 	"php":    2,                                      // class/interface inheritance now emits typed structural edges
-	"csharp": 5,                                      // awaited-receiver Task<T> unwrap + return_shape (was: this./base.-qualified call edges)
+	"csharp": 6,                                      // record members extracted (was: awaited-receiver Task<T> unwrap + return_shape)
 }
 
 // extractorSaltExtLang maps a lower-case file extension to the language
--- a/internal/parser/languages/csharp.go
+++ b/internal/parser/languages/csharp.go
@@ -1018,7 +1018,7 @@ func (e *CSharpExtractor) emitMethod(m parser.QueryResult, filePath, fileID stri
 	def := m.Captures["method.def"]
 	startLine1 := def.StartLine + 1
 
-	owner := csharpDirectMemberOwner(def.Node, src, "class_declaration", "struct_declaration", "interface_declaration")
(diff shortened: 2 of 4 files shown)


zzet (approved):

zzet:
One gap worth a follow-up

Positional record parameters are still not extracted. `public record Medal(int Id, string Motto)` emits the `Medal` type node and nothing else — no `Medal.Id`, no `Medal.Motto`. Those are the defining features of records and the most commonly used members, so "record members extracted" is broader than what actually ships.

pbednarcik:
Yep, aware of it. I left it out of #478 on purpose. That PR fixed one specific gate: the owner walk didn't recognize `record_declaration`, so members that actually exist as declaration nodes got skipped. Positional properties are a different thing — there's no declaration node to walk, they're synthesized, so the extractor has to fabricate the property nodes from the parameter list (plus small stuff like dedupe against explicit redeclarations). Didn't want to mix that into what was basically a one-line fix. Follow-up PR is next on my list. But good catch

Issue #479, opened 2026-08-06
hooks: every tool call pays ~280 ms of binary startup — proposal: thin hook client + untracked-cwd fast path


pbednarcik:
## The cost

Agent harnesses fire `PreToolUse` on every tool call, and the hook command is the full `gortex` binary. Measured over one real working session: 223 PreToolUse + 14 PostToolUse fires ≈ **74 seconds of serialized wait** — PreToolUse hooks block the tool call they precede. This is the mechanical signature of "the agent feels laggy" with Gortex hooks installed.

Where the time goes (Windows 11, warm cache; every number below includes ~31 ms of shell-spawn measurement overhead — compare rows, not absolutes):

| Invocation | Binary size | Wall time |
|---|---|---|
| `gortex hook` fed a payload that fails JSON parse on line 1 (no daemon dial, no work) | ~380 MB | **278–284 ms** |
| Scratch binary importing only `internal/hooks`, same payload | ~291 MB | 76–103 ms |
| Hello-world Go binary (spawn floor for this method) | 2.4 MB | ~31 ms |

So ~250 ms of the per-call cost is pure process startup of the everything-binary — before the hook logic runs at all. The worst case observed in that session was 3.0 s (a cold start eating the entire hook timeout budget).

## Why the hook can't currently build small

`internal/hooks` has modest direct imports, but its daemon transport is `internal/daemon` — a package that is client *and* server in one: 72 internal deps including all 31 tree-sitter grammars. `internal/mcp` (used for wire types) is heavier still (126 deps). Anything that dials the daemon today links the whole engine.

The wire protocol itself needs none of that: `internal/daemon/client.go` and `proto.go` are already pure stdlib (unix socket, newline-delimited JSON handshake + control frames). They're heavy only by package association.

## The second problem: untracked-cwd sessions pay full price for wrong advice

In a session whose cwd is not under any tracked root, the hook still costs the full ~280 ms per call — and the enrichment it emits mandates MCP tools that answer `repository not tracked` in that session. Deny plus an unusable alternative is the worst combination; silence would be strictly better and nearly free.

## Proposal (sketched, in adoptable phases)

1. **Extract a leaf transport package** — move `client.go` / `proto.go` / socket-path discovery out of `internal/daemon` into e.g. `internal/daemon/client` with zero engine imports (mechanical; the files already qualify). Server keeps re-exporting for compatibility.
[...]

PR #480, opened 2026-08-06
hooks: read the tracked-repo registry from config, not a daemon round trip

zzet:
Addresses the measured cause of #479's per-call hook latency — though not the one the issue proposes to fix.

## What I measured

Tracing the real `PreToolUse` path (macOS 15 / Apple Silicon, warm daemon, 44-repo registry) gives this per-call anatomy:

| phase | cost |
|---|---|
| exec + Go runtime init (400 packages) | ~30 ms |
| stdin read + parse + model hint | ~0.1 ms |
| **`ControlStatus` RPC — only to learn tracked roots** | **76 ms → 1210 ms → 2000 ms timeout** |
| dial + handshake for the tool call | **0.3 ms** |
| `get_file_summary` tools/call | 38–48 ms |

The handshake #479 proposes to make lightweight is **0.3 ms**. The issue's PoC saw a fresh client's first dial cost ~1 s and attributed it to per-session handshake bookkeeping; that second is this status call sitting in front of it.

## The bug behind the latency

`runPreToolUse` → `repoRootForFile` → `trackedRepoForPath` → `cachedTrackedRepos` → a full `daemon status` RPC, on every fire, to answer a membership test. There is a cache, but its TTL is process-scoped and every hook fire is a fresh process, so it only ever helps a `postGlob` loop inside one invocation.

On timeout the registry comes back empty, and an empty registry reads downstream as "no repo owns this path" — so the hook silently downgrades an indexed-file decision to generic guidance. Same payload, same file, back to back:

```
slow → 2016 ms → "[Gortex] Use explore first..."               (generic)
fast →  118 ms → "[Gortex] BLOCKED: ... (96 symbols indexed)"  (correct)
```

The hook gets slower and wronger under exactly the load that makes it slow.

## The change

`~/.gortex/config.yaml` is already the membership source of truth — a status row for a repo the daemon holds no index for is itself synthesised from that file. The hot path reads it directly: `repoRootForFile` and the scope gate need `Path`, `hookRepoScope` needs `Path` and `Prefix`, and `config.ResolvePrefix` is a pure function of the entry.

Live counters stay with the daemon. Their only consumer is `SessionStart`, which keeps its own status call and fires once per session rather than once per tool call.

Fallback is deliberate in both directions: an absent or unparseable config falls back to the daemon, so an unestablished registry degrades to today's behaviour rather than to a confident empty one. A config that parses with no repos is authoritative — that genuinely means nothing is tracked, and re-dialling there would put the round trip back on exactly the untracked sessions that should cost nothing.

## Result

24 interleaved paired runs (alternating binaries so daemon load drift hits both equally), tracked file, warm daemon:

| | before | after |
|---|---|---|
| median | 181.8 ms | **91.6 ms** |
| p90 | 2063.0 ms | **199.3 ms** |
| max | 2090.5 ms | **497.5 ms** |
| answers | 6/24 generic, 18/24 correct | **24/24 correct** |

The tail is what reads as "laggy", and it drops ~10×. The wrong-answer path is gone.

`internal/hooks`, `internal/config`, `cmd/gortex` green; `golangci-lint` clean on the changed package.

## What I did not change

I tested the obvious follow-on — narrowing the hook's `full` tool surface to `core` for the per-call probe — and it made no measurable difference (still 36–38 ms), so I dropped it rather than carry a change that only risks re-opening the #356 regression where a restricted preset silently emptied the Stop/PreCompact briefings.

That leaves `get_file_summary` at ~40 ms as the next target, and it wants a real answer rather than a guess: the hook uses a general-purpose file summariser to obtain a single symbol count. A narrow probe RPC returning `{tracked, indexed, symbol_count}` is the shape worth measuring next.

Two notes for #479's remaining phases, both now cheaper to judge with the status call gone:

- **Phase 1/2 as specified** target a 0.3 ms handshake. They may still be worth doing for structure, but not for latency.
- **Phase 3** (the thin binary) is worth ~30 ms here and ~250 ms on the Windows 11 numbers in the issue — a much easier sell there than on Apple Silicon, and now measurable against a baseline that is not dominated by this round trip.

### DIFF
--- a/internal/hooks/session_paths.go
+++ b/internal/hooks/session_paths.go
@@ -17,10 +17,15 @@ type sessionScope struct {
 	CWD       string
 }
 
-// hookTrackedReposFn is the daemon's tracked-repo list, injectable so tests
+// hookTrackedReposFn is the tracked-repo list, injectable so tests
 // never dial a real daemon.
-var hookTrackedReposFn = cachedTrackedRepos
+// Reads the local registry (see tracked_repos_config.go) and only falls back
+// to a daemon round trip when that file cannot be established.
+var hookTrackedReposFn = configTrackedRepos
 
+// cachedTrackedRepos is the daemon-backed registry: the fallback for
+// configTrackedRepos, and still the direct source for callers that need the
+// live per-repo counters the config cannot supply.
 func cachedTrackedRepos() []daemon.TrackedRepoStatus {
 	status, err := cachedDaemonStatus()
 	if err != nil || status == nil {
--- a/internal/hooks/tracked_repos_config.go
+++ b/internal/hooks/tracked_repos_config.go
@@ -1,0 +1,121 @@
+package hooks
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/zzet/gortex/internal/config"
+	"github.com/zzet/gortex/internal/daemon"
+)
+
+// The tracked-repo registry is `~/.gortex/config.yaml` — a daemon status row
+// for a repo the daemon holds no index for is itself synthesised from that
+// file, so the config is the membership source of truth and the status call
+// only adds live counters on top.
+//
+// Reading it directly is what the hot path wants. Asking the daemon instead
+// cost a full ControlStatus round trip per hook fire — every tracked repo
+// marshalled with its file/node/edge counts and memory breakdown — measured at
+// ~76 ms warm, over a second under load, and often enough hitting its 2 s
+// deadline that the hook then answered from an empty list. An empty list reads
+// as "no repo owns this path" downstream, so a timeout did not merely make the
+// hook slow: it silently downgraded an indexed-file decision to generic
+// guidance.
+//
+// Membership is all the hot path needs — repoRootForFile and the scope gate
+// read Path, hookRepoScope reads Path and Prefix. The live counters stay zero
+// here; SessionStart is the only consumer that reads them and it keeps its own
+// status call, firing once per session rather than once per tool call.
+
+var (
+	configReposOnce sync.Once
+	configReposVal  []daemon.TrackedRepoStatus
+	configReposOK   bool
+)
+
+// configTrackedRepos returns the tracked-repo registry read from the user's
+// global config, falling back to the daemon only when that file is absent or
+// unparseable. The result is memoised for the process: a hook invocation is
+// short-lived and the registry cannot change underneath it.
+func configTrackedRepos() []daemon.TrackedRepoStatus {
+	configReposOnce.Do(func() {
+		configReposVal, configReposOK = loadConfigTrackedRepos()
+	})
+	if configReposOK {
+		return configReposVal
+	}
+	// Unreadable config: fall back to the daemon rather than report an empty
+	// registry, which downstream cannot distinguish from "nothing is tracked".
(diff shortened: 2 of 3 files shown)


pbednarcik:
Measured this branch against main on the same Windows box as my #479 numbers (Win11, Ryzen 9 9950X, Optane P5800X, go1.26.5).

Protocol: 24 interleaved pairs, binary order alternating each pair, `hook --mode=enrich` with an identical PreToolUse `Read` payload targeting an indexed file in a tracked repo, 3 discarded warmup fires per binary, daemon warm and fully settled.

| binary | median | p90 | max | min |
|---|---|---|---|---|
| main | 175 ms | 178 ms | 870 ms | 167 ms |
| this PR | 86 ms | 90 ms | 93 ms | 83 ms |

Both binaries returned the full specific guidance on all 24 fires — no deny→generic degradation on a warm daemon — so this halves the per-fire cost here with no answer-quality change. The status RPC was ~90 ms of the ~175 ms total on this box. What remains (~86 ms) is the full-binary spawn, which is the thin-client discussion in #479.

zzet:
@pbednarcik, before jumping into multi-binary distribution, I'd prefer to explore if there is a possibility to reduce the impact via:
1. limitations on when the hook is fired
2. if it's a fallback to the native grep/read tools - the reasoning behind this, which could be a model distrust for some responses, and solving this problem is way more important than the wall-time reduction (doesn't matter how fast you're running if you're running in the wrong direction)

pbednarcik:
Agreed and the numbers above make it clear: with this merged, the thin binary would only buy ~40 ms more per fire. Not worth a second binary.

On (2), I have something concrete you can investigate: in my sessions the fallback is mostly not model distrust. The cwd is often a workspace root — a directory containing tracked repos without being tracked itself. The MCP tools answer repository not tracked: <cwd> there, the server instructions then explicitly tell the model to fall back to native reads, and the hook goes on flagging exactly the calls it caused. Once one MCP call errors, the model sticks with native tools for the rest of the session (the hook is exit-0 advisory, nothing pulls it back). Fixing the workspace-root answer — fan out across the contained repos instead of erroring — would remove the loudest fallback path before the distrust question even starts.

pbednarcik:
Followed up on my workspace-root claim with a source dive — the fallback loop is two layered gates in `cmd/gortex/daemon_mcp.go`:

1. `isCWDTracked` only checks one direction of containment. A cwd inside a tracked root passes (`HasPathPrefix(cwd, root)`), but a cwd that contains tracked roots — a workspace root like `<root>` over `<root>/projects/<repo>` — is classified untracked. That's my daily layout in two different workspaces.
2. The untracked gate in `Dispatch` refuses every `tools/call` before parsing arguments. Once the session is untracked, a `read` whose `target.file` explicitly names a file inside a tracked root is still rejected — and `notTrackedError` reports `sess.CWD`, not the path the call asked about, which is why the error looks so wrong from the agent's side ("repository not tracked: `<workspace-root>`" in response to a request about a tracked file).

What makes it a loop rather than a dead end: the hook layer does have workspace-root awareness — its session orientation says "cwd is a workspace root containing N tracked repos, calls fan out across them" — so the hooks keep steering the model to MCP while the dispatcher fail-closes the same calls. The model obeys the hook, gets `-32000`, falls back to native reads, and the hook flags those. Every fire in that state is wasted on both sides.

Issue #277, opened 2026-07-09
bug: Windows: gortex init-generated config uses uppercase drive letter, MCP server publishes 0 tools due to case-sensitive path coverage


david-harasta-takkt:
**Describe the bug**
On Windows, the MCP server connects but publishes an empty tool list
(`tools/list` -> `[]`) when the MCP client's working directory and the tracked
repository path differ only by the case of the drive letter (`c:\` vs `C:\`).
The daemon reports the directory as "not covered by any tracked repository", so
the session is INACTIVE and no tools appear in the client's tool picker - even
though the repo is indexed and healthy (`gortex tools list` on the CLI returns
the full catalog, and `gortex daemon status` shows it tracked).

Notably, the mismatching config was generated by gortex itself: `gortex install`
and `gortex init` were run from inside the repo, and the resulting
`~/.gortex/config.yaml` stored the path with an uppercase drive letter
(`C:\...`), while the MCP client (VS Code) launches the server with a lowercase
drive letter (`c:\...`). So a default, tool-generated setup ends up INACTIVE
with zero tools.

**To Reproduce**
1. From inside a repo whose cwd uses a lowercase drive letter
   (`c:\work\git\myrepo`), run `gortex install` and `gortex init`.
   The generated `~/.gortex/config.yaml` stores `path: C:\work\git\myrepo`
   (uppercase drive letter).
2. Start the MCP server as the client does: working directory
   `c:\work\git\myrepo`, command `gortex mcp`.
3. Send `initialize` then `tools/list`.
4. See `tools/list` return `{"tools":[]}`, and the `initialize` response
   `instructions` say:
   "Gortex is connected but INACTIVE for this directory: \"c:\\work\\git\\myrepo\"
   is not covered by any tracked repository ..."
5. The server shows "Discovered 0 tools" in the client log and is absent from
   the tool picker.

Proof it is the drive-letter case (identical daemon and repo, only case differs):
- cwd `C:\work\git\myrepo` (uppercase) -> 58 tools published
- cwd `c:\work\git\myrepo` (lowercase) -> 0 tools published

**Expected behavior**
On Windows, the path-coverage comparison should be case-insensitive (at least
for the drive letter, ideally the whole path, since NTFS is case-insensitive).
A cwd of `c:\...` should match a tracked `C:\...` repo, and the full tool
catalog should be published. Ideally `gortex install` / `gortex init` and the
MCP client should also agree on drive-letter case so a default setup is never
INACTIVE.

**Environment:**
[...]

PR #484, opened 2026-08-07
daemon: resolve workspace-root cwds to their contained repos' workspace

pbednarcik:
Fixes the fallback loop root-caused in https://github.com/zzet/gortex/pull/480#issuecomment-5214702976: a session whose cwd *contains* tracked repos (a workspace root like `<root>` over `<root>/projects/<repo>`) was classified untracked by both gates — `isCWDTracked` and `ScopeForCWD` only tested one direction of containment — so every `tools/call` got `-32000 repo_not_tracked` while the hook layer kept steering the agent to MCP.

Two coordinated changes:

- **`ScopeForCWD` gains a reverse-containment fallback.** When cwd is inside no tracked repo, repos rooted *under* it are collected: exactly one repo → its full scope (home repo included, locality ranking intact); several repos sharing one effective workspace → workspace-bound session with no home repo; repos spanning different workspaces → still fail closed, since a single-workspace session scope can't express the union and widening silently would cross the isolation boundary.
- **`cwdReachable` delegates to `ScopeForCWD`** instead of duplicating containment logic, so the dispatcher gate and the session scope can never disagree again: the gate admits a session exactly when the scope can bind it. `isCWDTracked` keeps its strict "inside a repo" meaning (its parent-must-not-count pin still passes).

Tests: three-layout scope test (single nested repo / shared-workspace pair / mixed-workspace pair), a case-variant workspace-root pin (the #277 folding scenario one level up), and a dispatcher test covering both the pass and the mixed-workspace refusal. Full indexer/mcp/cmd suites show failure sets identical to main on Windows.

Validated live on my box in both shapes — a root above a single repo and a root above two repos sharing a config-assigned workspace slug — with cross-workspace isolation confirmed intact and my C# fixture battery unchanged.

Judgment calls I want to flag rather than hide:

1. With a single tracked repo, *any* ancestor cwd (even a drive root) now binds to it. Consistent with single-repo mode's allow-everything, but say the word if you'd rather see a depth bound.
2. A parent spanning multiple workspaces still gets `repo_not_tracked` while the hook orientation advertises fan-out for it — narrower inconsistency than before, not zero. Multi-workspace sessions would be the real fix and felt out of scope here.
3. Multi-repo binding leaves `projectID` empty even when the repos share a project.

### DIFF
--- a/cmd/gortex/daemon_integration_test.go
+++ b/cmd/gortex/daemon_integration_test.go
@@ -223,6 +223,17 @@ func TestDaemon_EndToEnd_WorkspaceRootCWDServed(t *testing.T) {
 	}
 	require.NoError(t, json.Unmarshal(reply, &resp))
 	require.NotNil(t, resp.Result, "workspace-root cwd must get a tool result: %s", string(reply))
+
+	// Scope-sensitive twin: graph_stats never consults the session
+	// scope, so a binding that admits the session but blanks its scope
+	// would pass the assertion above. A symbol search runs through the
+	// scoped engine — the tracked repo's own symbol must come back.
+	frame = []byte(`{"jsonrpc":"2.0","id":8,"method":"tools/call","params":{"name":"search_symbols","arguments":{"query":"main"}}}`)
+	require.NoError(t, client.WriteMCPFrame(frame))
+	reply, err = client.ReadMCPFrame()
+	require.NoError(t, err)
+	require.Contains(t, string(reply), "main.go",
+		"workspace-root session's scope must reach the contained repo's symbols: %s", string(reply))
 }
 
 // TestDaemon_EndToEnd_TrackAddsRepoLive proves track-while-running is
--- a/cmd/gortex/daemon_integration_test.go
+++ b/cmd/gortex/daemon_integration_test.go
@@ -163,8 +163,12 @@ func TestDaemon_EndToEnd_GraphStatsOverMCPProxy(t *testing.T) {
 // dispatcher. An agent opening in an untracked directory must see a
 // usable error on every tool call, not a silent zero result.
 func TestDaemon_EndToEnd_UntrackedCWDRejectedViaProxy(t *testing.T) {
-	socket, trackedRoot := spinUpDaemon(t)
-	untracked := filepath.Dir(trackedRoot) // parent of tracked root is NOT tracked
+	socket, _ := spinUpDaemon(t)
+	// A directory unrelated to the tracked root in either containment
+	// direction. (The tracked root's PARENT no longer qualifies — a
+	// workspace-root cwd above tracked repos is served, see
+	// TestDaemon_EndToEnd_WorkspaceRootCWDServed.)
+	untracked := t.TempDir()
 
 	client, err := daemon.DialTo(socket, daemon.Handshake{
 		Mode: daemon.ModeMCP,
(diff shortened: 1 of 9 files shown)


zzet:
Hey @pbednarcik, thanks — this is a carefully scoped fix.

The scope of change is tricky and has some hidden pitfalls. Two changes make sense to add before merge, both small.

### 1. The new empty-prefix bound state leaks content search across workspaces

The multi-repo branch of `scopeForWorkspaceRootLocked` returns `(sharedWS, "", "", true)` — the first time a *bound* session can carry an empty `repoPrefix`. That empty prefix flows into the content-search channel unfiltered:

- `sessionScope` caches `scopeRepoPrefix = ""` for the bound session;
- `mergeContentChannel` (`internal/mcp/search_corpus.go`) hands it to `SearchContent`, and `store_content_fts.go` adds its `repo_prefix` predicate only when the prefix is non-empty — so the FTS query spans **all repos in all workspaces**;
- the merged hits are materialized via raw `graph.GetNode`, and `filterNodes` is a no-op when `RepoAllow` is nil (the default-intent path), so nothing re-applies the workspace boundary afterwards.

Repro shape: daemon tracks `r1` + `r2` (workspace `shared`) plus `z` (workspace `other`) with content-indexed docs. A session dialing from the root above `r1`/`r2` binds to `shared`, and `search_symbols corpus:"all"` returns `z`'s doc chunks. Pre-PR this cwd was rejected at the dispatcher, so the path was unreachable — the PR makes it live.

Suggested fix: when a bound session has an empty repo prefix, derive the allow-set from `ReposInWorkspace(workspaceID)` (populate `RepoAllow` before the content channel runs, or workspace-filter the merged nodes). A regression test driving the shared-workspace binding through a scope-sensitive call would pin it.

### 2. `TestScopeForCWD_CaseMismatchedWorkspaceRoot` never exercises a case mismatch

`caseVariantOf(filepath.Dir(dir))` case-toggles the basename of `t.TempDir()`, which is Go's `%03d` sequence directory (`001`) — digits only, so `swapCaseASCII` returns it unchanged and the "variant" is byte-identical to the real parent. The test currently passes even with case folding stripped from the reverse-containment path. Interposing a lettered directory (e.g. index the repo under `<tmp>/Work/myrepo` and toggle `Work`) makes it pin what the name claims.

### Coverage note (non-blocking)

The two new gate tests assert only the absence of `repo_not_tracked`, via `graph_stats` — which doesn't consult the session scope. A regression that admits a workspace-root session but blanks its scope would keep both tests green. One assertion through a scope-sensitive tool (or directly on the bound scope) would close that. Relatedly, the `(ws, "", "", true)` state is only asserted at the `ScopeForCWD` unit level today — the dispatcher and e2e tests cover the single-repo pass and the mixed-workspace refusal, never the shared-workspace binding.

### On the flagged judgment calls

1. Ancestor-at-any-depth binding for a lone workspace: fine as-is — consistent with single-repo mode's allow-everything, and the mixed-workspace fail-closed keeps it contained. No depth bound needed now.
2. Mixed-workspace parents vs. hook orientation: agreed it's pre-existing and out of scope here; this PR strictly narrows the inconsistency.
3. Empty `projectID` on multi-repo binding: acceptable, with one downstream wrinkle worth a small guard —`activeProjectName` falls through to the server-wide `s.activeProject` when `proj == ""`, so a shared-workspace-root session's initialize instructions can advertise a project outside its bound workspace. Returning nothing for the bound-empty case would avoid steering agents toward a project their scope can't see.

pbednarcik:
Thanks — both were real. Fixed in e7b2d319:

**1. Content-search leak**: `mergeContentChannel` now clamps its hits to the bound workspace's repo set whenever the session has no home repo — the allow-set comes from `ReposInWorkspace(workspaceID)`, and a bound session whose workspace enumerates nothing fails closed (merges nothing). Regression test builds your exact repro shape (r1+r2 sharing a workspace slug, z with a content body in another workspace) and pins both directions: the bound workspace-root session must not see z's content, an unbound session keeps the whole-graph channel. Wrote the test first — it reproduced the leak verbatim before the clamp.

**2. Case-mismatch test**: you're right, it toggled `t.TempDir`'s digits-only basename and exercised nothing. The repo now sits under a lettered intermediate directory (`Work/myrepo`), and the test guards `require.NotEqual` on the variant so it can't silently go vacuous again.

**Coverage note**: the shared-workspace binding is now asserted at the mcp layer (the new test checks `bound == true`, empty home repo, and a scope-sensitive channel), and the e2e workspace-root test gained a scope-sensitive twin — a `search_symbols` call that must return the contained repo's symbol, which a session admitted with a blanked scope would fail.

**`activeProjectName`**: took the guard as suggested — a resolved cwd now owns the answer even when its project is empty, so a shared-workspace-root session's initialize instructions no longer advertise the server-wide default. Covered by a test with the fallback behavior for unresolved cwds pinned alongside.

Suites: affected packages' failure sets are byte-identical to the branch base on my Windows box (pre-existing platform failures only, stash-verified); the e2e twin rides the existing macOS-only family.

zzet (approved):

PR #485, opened 2026-08-07
daemon: add a control probe that indexing cannot delay

zzet:
Root-cause fix for the hook timeout cluster in #479. One daemon-side change; it is the cause behind four separate hook bugs.

## The measurement

Five **serial** `daemon status` calls, no concurrent load, 44-repo workspace — identical 14,468-byte payload every time:

```
11531 ms · 8225 ms · 344 ms · 2041 ms · 156 ms
```

Same payload, 74× spread. That is not the work, it is the wait. The daemon log shows incremental derived passes running 28 seconds at a stretch, and `Status` takes `c.mu` — which is held for the entire duration of a track / reload / enrichment.

I had assumed the hooks were causing this themselves (63k connections). They are not: these calls were serial, one second apart, with no hooks running.

## This hazard is already known in this file

```go
// multiWatcher is an atomic pointer, not a mu-guarded field: the daemon's
// teardown hook reads it, and reading it under mu is what kept `daemon
// stop` queued behind a running track / reload / enrichment.
```

```go
// Guarded by its own mutex, deliberately NOT by mu. Shutdown is the one
// control RPC that has to work when the daemon is at its least healthy,
// and mu is held for the entire duration of a track / reload / enrichment
// — minutes on a large workspace.
```

`Shutdown` and `multiWatcher` were given that treatment. `Status` never was, and it sits on the critical path of every agent hook.

## The consequence is a wrong answer, not slowness

Callers on sub-second budgets time out against a perfectly healthy daemon and render their unreachable fallback. Measured over 233 real fires, the SessionStart briefing emits

> ⚠️ Gortex daemon status query failed — continuing with rule-only enforcement.

in **81.5%** of sessions. Same root cause behind UserPromptSubmit (82.6% at its 800 ms budget), PostToolUse (48.5% at 2 s), and PreCompact (100% at 5 s, which has never emitted anything).

## What `ControlProbe` does

Answers "up, ready, and what do you track" with none of the aggregation `Status` performs — no `c.mu`, no graph handle, no `runtime.ReadMemStats` (which stops the world), no per-repo `stat`. It reads two atomics and the config registry, which is the membership source of truth anyway: a status row for a repo the daemon holds no index for is itself synthesised from that file.

`ProbeResponse` is a distinct type rather than a subset-typed `StatusResponse` deliberately. Sharing the type would leave every aggregate field present but zero, and a caller cannot tell a real zero from an unpopulated one — that ambiguity is exactly what made a timed-out status read as "nothing is tracked".

## Tests

Five, under `-race`, including a deliberate pair:

| test | result |
|---|---|
| `TestProbeAnswersDuringLongMutation` — `c.mu` held | **0.00 s** |
| `TestStatusBlocksDuringLongMutation` — same mutex | **0.31 s, blocked** |

The second test is the guard on the first: it asserts `Status` *does* block, so if `Status` is ever fixed properly this fails and `Probe`'s justification gets revisited rather than silently kept. The rest cover project-nested repos, context cancellation, and a nil config manager (liveness must not depend on a registry being present).

## Scope

Deliberately daemon-side only. Callers still use `Status`; wiring them to `Probe` is a follow-up so the transport lands and can be reviewed on its own. The follow-up has one real design question worth separating: SessionStart renders node/edge counts, which a probe cannot supply, so it needs an explicit answer for "healthy but counts unknown" rather than silently rendering zeros.

## Related

- #480 (merged) — removed this same `ControlStatus` round trip from the PreToolUse path; p90 2063 → 199 ms, and 6/24 wrong answers → 0. That fix worked by *avoiding* the contended control plane; this one fixes the plane.
- #483 — guard that every hook-dispatched tool still resolves.

### DIFF
--- a/cmd/gortex/daemon_controller.go
+++ b/cmd/gortex/daemon_controller.go
@@ -687,6 +687,72 @@ func resolveSearchBackend(b search.Backend) searchBackendInfo {
 // Status gathers per-repo stats and basic process metrics. Daemon-level
 // fields (PID, uptime, socket, session count) are filled in by the
 // daemon itself before the response goes out.
+// Probe answers "is the daemon up, is it ready, and what does it track"
+// without taking c.mu and without reading the store.
+//
+// This is the whole point of the call. Status takes c.mu, and c.mu is held
+// for the entire duration of a track / reload / enrichment — the same
+// reasoning that already keeps multiWatcher and onShutdown off that mutex.
+// Measured on a 44-repo workspace, serial Status calls with no concurrent
+// load returned the identical payload in 156 ms to 11.5 s depending only on
+// what the indexer was doing. Callers on a sub-second budget therefore
+// concluded the daemon was unreachable while it was perfectly healthy.
+//
+// So: no c.mu, no graph handle, no runtime.ReadMemStats, and no per-repo
+// stat. ready/enriched are already atomics, and the tracked-repo registry is
+// the config — which is the membership source of truth anyway, since a
+// status row for a repo the daemon holds no index for is itself synthesised
+// from it.
+//
+// Anything added here that reads the store or takes a lock reintroduces
+// exactly the stall this exists to avoid; TestProbeAnswersDuringLongMutation
+// pins that.
+func (c *realController) Probe(ctx context.Context) (daemon.ProbeResponse, error) {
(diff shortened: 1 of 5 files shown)


zzet:

PR #487, opened 2026-08-07
hooks: stop reporting a busy daemon as a broken one at session start

zzet:
Stacked on #485 (needs `ControlProbe`). Review that one first.

## The bug

The SessionStart briefing is appended to the session system prompt and visible on every turn — by its own comment, *"the strongest 'rule restated' surface we have."* Measured over 233 real fires, **81.5% of sessions** got this instead:

> ⚠️ Gortex daemon status query failed — continuing with rule-only enforcement.

The daemon was healthy every time. `Status` queues behind the controller mutex, which a track / reload / enrichment holds for minutes, so the 2 s budget expired while the daemon was merely mid-index. **A budget overrun carries no information about health**, but the briefing treated it as proof of ill health — and then told the agent enforcement was degraded.

## The fix

`ControlProbe` (#485) cannot queue behind indexing, so it distinguishes *busy* from *down*. The status path falls back to it: a genuinely unreachable daemon still emits the hard warning; an overrun no longer does.

## The part worth reviewing

The probe returns identity but no aggregates — it computes none, deliberately. Rendering those absent counters as zeros would trade one wrong briefing for another: **"0 nodes" is a claim about the graph**, and an indexed repo reported as empty is its own kind of lie.

So `CountsUnknown` marks a probe-sourced response and the renderers say the totals were not queried instead. I'd rather this be explicit in the type than inferred from a zero — the inability to tell a real zero from an unasked one is the same failure that made a timed-out status read as "nothing is tracked".

Worth noting: I initially patched two renderers and a test caught a third still printing `0 nodes` for a tracked repo. That's the test doing its job, and it's why the assertions are written against rendered output rather than internal state.

## Tests

| test | asserts |
|---|---|
| `BusyDaemonIsNotReportedAsBroken` | no "status query failed" / "unreachable" when the daemon is merely busy |
| `UnknownCountsAreNotRenderedAsZero` | no "0 nodes"; says "not queried" |
| `UnreachableDaemonStillWarns` | this narrows what counts as unreachable — it does not remove the state |
| `HealthyStatusStillReportsCounts` | invisible on a quiet daemon; real counters render as before |

`internal/hooks`, `internal/daemon`, `cmd/gortex` green; `golangci-lint` clean.

## Remaining in this cluster

Same root cause, not yet wired to the probe — each needs its own decision, so they aren't bundled here:

- **PreCompact** — 13/13 fires emit nothing; its socket call starves for the full 5 s, then falls through to a vestigial HTTP fallback pointed at the removed `:8765` surface (nothing listens there).
- **UserPromptSubmit** — 82.6% at its 800 ms budget, **3 emits in 311 fires**. Worth deciding whether it earns its place at all before optimising it.
- **PostToolUse** — 48.5% at 2 s, and separately its installed matcher (`mcp__gortex__*` only) disagrees with its handler (which enriches Grep/Glob/Read), so that enrichment path is unreachable.

### DIFF
--- a/internal/daemon/proto.go
+++ b/internal/daemon/proto.go
@@ -284,6 +284,15 @@ type StatusResponse struct {
 	// the daemon — while SearchBackend above may simultaneously report
 	// the symbol index as disk-resident. Nil when no indexer is wired.
 	TrigramCache *TrigramCacheStats `json:"trigram_cache,omitempty"`
+	// CountsUnknown marks a response assembled without the aggregate pass —
+	// the per-repo and whole-store counters are zero because they were never
+	// computed, not because the graph is empty. Set when a caller fell back
+	// to ControlProbe after ControlStatus exceeded its budget.
+	//
+	// Without this a consumer cannot tell "no nodes" from "did not ask", and
+	// rendering an unasked zero as a real one is the same class of mistake as
+	// reading a timed-out status as "nothing is tracked".
+	CountsUnknown bool `json:"counts_unknown,omitempty"`
 	// PProfAddr is set when the daemon has opened an HTTP pprof
 	// listener (via the GORTEX_DAEMON_PPROF_ADDR env var). Empty
 	// string means pprof is not enabled on this daemon.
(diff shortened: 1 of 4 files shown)


zzet:

PR #499, opened 2026-08-08
fix(init): isolate integration-test XDG paths

peterkc:
## Summary

Init tests could inherit real config paths from the developer's machine. A test
could then write temporary repositories into the user's Gortex config.

This change sends all test config, state, runtime, and daemon files to temporary
directories. A regression test proves the inherited paths remain untouched.

```text
Before: init test -> inherited user paths
After:  init test -> temporary test paths
```

## Changes

- Add one helper that isolates all init test paths.
- Use it in every `runInit` integration test.
- Verify repeated init stays isolated and does not modify inherited paths.

## Testing

- [ ] All tests pass (`go test -race ./...`) - the full race suite was not run.
- [x] `go test -race ./cmd/gortex -run '^TestInit' -count=1` - 22 tests passed.
- [x] `go test ./cmd/gortex -run '^TestInitCreatesProjectMarker$' -count=1` - passed.
- [x] `go test ./...` - 12,077 tests passed across 158 packages.
- [x] New tests added for new functionality.
- [x] Benchmarks not run because the change is test isolation only.
- [x] The production config SHA-256 was unchanged after the qualified full suite.

## Checklist

- [x] Code follows existing patterns in the codebase.
- [x] No unnecessary abstractions added.
- [x] Language extractor metadata is not applicable.
- [x] Member-edge behavior is not applicable.

### DIFF
--- a/cmd/gortex/init_integration_test.go
+++ b/cmd/gortex/init_integration_test.go
@@ -48,7 +48,7 @@ func TestInitDryRunJSONReportShape(t *testing.T) {
 
 	repo := t.TempDir()
 	home := t.TempDir()
-	t.Setenv("HOME", home)
+	isolateInitTestEnv(t, home)
 
 	initYes = true
 	initDryRun = true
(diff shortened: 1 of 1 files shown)


zzet:
@peterkc Is any other work is expected or it accedentally stays as draft?

PR #500, opened 2026-08-08
fix(graph): exclude topics from repo-prefix diagnostics

peterkc:
## Summary

Valid topic nodes were incorrectly reported as repository-prefix errors. Topics
use global names, so they should not be checked like repository-owned nodes.

This change skips that check for topics in both the in-memory and SQLite stores.
Regression tests keep both implementations consistent.

```text
Repository node -> check repository prefix
Topic node      -> skip this check
```

## Changes

- Exclude topics from repository-prefix checks.
- Apply the same rule in the in-memory and SQLite stores.
- Add shared regression coverage for topic nodes.

## Testing

- [ ] All tests pass (`go test -race ./...`) - the full race suite was not run.
- [x] `go test ./internal/graph/...` - 989 tests passed.
- [x] `go test -race ./internal/graph ./internal/graph/store_sqlite` - 908 tests passed.
- [x] `go vet ./internal/graph/...` - passed.
- [x] Isolated `go test ./...` - 11,962 tests passed across 180 packages.
- [x] New tests added for new functionality.
- [x] Benchmarks not run because the predicate adds one constant kind exclusion.

## Checklist

- [x] Code follows existing patterns in the codebase.
- [x] No unnecessary abstractions added.
- [x] Language extractor metadata is not applicable.
- [x] Member-edge behavior is not applicable.

### DIFF
--- a/internal/graph/prefix_diagnostics.go
+++ b/internal/graph/prefix_diagnostics.go
@@ -18,9 +18,9 @@ import (
 // The audit applies only to repository source nodes. Some synthetic nodes
 // retain a FilePath as attribution rather than as an on-disk source path:
 // semantic externals use `external::…`, synthesized external calls use
-// `external-call::…`, and contracts use identity namespaces that deliberately
-// do not mirror their source path. Those populations are outside the ownership
-// invariant even when FilePath is non-empty.
+// `external-call::…`, and contracts plus topics use identity namespaces that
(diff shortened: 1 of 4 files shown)


zzet:
@peterkc Is any other work expected, or does it accidentally stay as a draft?

zzet (approved):

PR #504, opened 2026-08-08
Lower the compact target selector on every analyze surface

### LINKED ISSUE #449: bug: analyze kind=impact ignores its target and returns a repo-wide hotspot ranking (target absent from its own result)
## Summary

`analyze(kind:"impact", target:{symbol:...})` does not answer "what is the blast radius of this symbol". It returns a repo-wide hotspot ranking that is **identical in shape regardless of target**, does not contain the target, and mixes in stdlib and non-product files. There is no caveat saying the target was dropped, so the caller gets a confident answer to a different question.

Same fail-open class as #436: a wrong answer that is indistinguishable from a right one.

## Environment

- daemon `v0.62.0+99d745c`, healthy (not warming, not stale)
- corpus: single tracked Go repo, **2408 files / 124,619 nodes / 769,213 edges**
- measured 2026-08-04

## Exact query

```
analyze(kind:"impact",
        target:{symbol:"<repo>/services\domain-core\internal\repository\hosting.sql.go::Queries.MarkHostingAccountActive"})
```

## Observed

100 rows. The target symbol **does not appear at all**. The top rows are unrelated to hosting:

```
SepayReconcileRunner.Run     internal/billing/sepay_reconcile.go     score 33.14
accountFlags.Set             cmd/dev-seed-accounts/main.go           score 30.82
rows                         tests/scripts/test_local_gate_plan.py   score 30.19
```

The ranking also includes Go stdlib nodes and files from a reference-only Next.js tree that cannot be affected by a Go method:

```
<repo>::stdlib::fmt::Errorf                     fan_in 2878
<repo>::stdlib::io::WriteString                 fan_in 4690
app\dashboard\ssl\[id]\page.tsx::Label
components\ui\badge.tsx::Badge
```

The returned fields (`pagerank`, `centrality`, `hotspots`-style `score`, `fan_in`, `complexity`) are global graph metrics. Nothing in the row set is relative to the target.

## Known-positive control — the graph *does* have the answer

The same symbol, same daemon, same moment:

```
relations(operation:"callers", target:{symbol:"…::Queries.MarkHostingAccountActive"})
→ 11 call edges across 5 functions in 3 files
```

Independently corroborated: this repo pins that exact expectation (11 edges / 5 functions / 3 files) in a committed document, and I read back four of the reported `file:line` positions — each really contains the call.

So the neighborhood is present and correct in the graph. `impact` simply does not consult it for the target it was given.

## Expected

One of:

1. rank by the target's actual dependency closure (callers / transitive dependents), or
2. if a target-scoped impact is not implemented for this selector, **fail closed** — return an explicit `target_not_applied` / `unsupported_target` caveat instead of a global ranking.

Either is fine. Silently substituting a global hotspot list is the part that bites, because the output looks authoritative.

## Impact on users

Blast-radius questions are exactly the "before I edit this, what breaks?" case. A global hotspot list answers that question wrongly while looking precise, and stdlib/non-product rows make it look thorough. In our repo the tooling policy now forbids using `analyze impact` as a target-specific blast radius at all, and requires closing every such claim with a second tool.

Related: #436 (contract tier returns 0 with no caveat), #433 (zero-edge classification wording).

zzet:
Fixes #449.

## What was broken

`analyze(kind:"impact", target:{symbol:"…"})` returned a repo-wide hotspot ranking instead of the target's blast radius. The decisive control from the reopened report held: `target:{symbol:"GORTEX_NEGATIVE_CONTROL_NO_SUCH_SYMBOL_9f68c2a1"}` returned the same confident 100-row ranking rather than the documented symbol-not-found error, so the selector was never reaching `resolveImpactTarget`.

It was reproducible on every tool preset, including `facade-v1`, which rules out the surface-negotiation theory in the issue thread. `applyFacadeTarget` and `normalizeFacadeArguments` were both correct in isolation — the selector was destroyed upstream of them, by two independent layers.

**1. The argument reconciler renamed the selector away.** `aliasCanonicals` in `internal/mcp/param_alias.go` maps `target` → `id` to tolerate a hallucinated parameter name. `reconcileToolParams` runs inside `wrapToolHandlerMode`, outside the facade wrapper, on every call and every preset. `analyze` declares a real string `id`, so the entire `target` **object** was moved into it; `stringArg` then read `""` and the handler ran exactly as though no target had been sent. The same rewrite moved a read-shaped `context` envelope into `edit`'s `content` — the text an edit writes.

**2. Compact analyze calls never reached the dispatcher.** `wrapLegacyFacade` routed a reused legacy name to `handleFacade` only when the call carried `operation` or the session had negotiated `facade-v1`. **`analyze` discriminates on `kind`**, so a well-formed compact analyze call never carries an `operation` and never got its envelope lowered outside a facade-v1 session.

This also explains the asymmetry in the report: `options:{id}` worked because `options` is not an alias candidate and is merged by the facade, while `target` was renamed before any facade code ran.

## What this changes

- An alias candidate must declare a JSON Schema type that can hold the value, so no envelope container is ever renamed into an unrelated scalar.
- A reused facade name lowers any call that speaks the compact envelope, not only one carrying `operation`. A malformed selector (a scalar `target`) is refused rather than dropped.
- **Fail closed on a selector the operation cannot consume.** `analyze(kind:"routes", target:{symbol})` — the smaller symptom in the report — is now a structured `unsupported_target` refusal naming the kinds that do accept one. The rule generalises: 21 analyze kinds dispatch to a different legacy tool (`clones` → `find_clones`, which has no `id`), the `to` container is advertised on every `relations` and `trace` operation but only lowered by `flow`/`path`/`taint`, and `change(operation:"detect")` accepted a target it has no field for. All were accepting selectors and discarding them.
- The lowering and the check read the target field name from one function, so a selector cannot be accepted by one and dropped by the other.
- `target:{symbols:[…]}` lowered to a JSON array while the impact reader only split on commas — a batch selector matched nothing and returned an empty ranking with no error. Symbol-id readers now accept every encoding the surface produces.
- The parameter reconciler no longer accepts a caller's filter as its own negation. "include" and "exclude" are two substitutions apart, the same budget a genuine typo gets, so `include_tests` landed in `exclude_tests`, `exclude_speculative` in `include_speculative`, and `max_score` in `min_score` — each returning a well-formed result set close to the complement of the requested one.
- `analyze` advertises `target` in `tools/list`, so the schema a `core`-preset client receives matches what `capabilities` documents.
- `def_use`'s missing-target error names the public selector instead of only the legacy `id` / `ids`.

## Tests

The report is right that no existing test could see this: `TestAnalyzeImpact_FacadeTargetReachesTheHandler` calls the lowering and the handler directly, and passed throughout. The new tests drive real JSON-RPC frames through `MCPServer().HandleMessage` — the same path the daemon uses for MCP and CLI `tools/call` — and would have failed before this change:

- `TestAnalyzeTargetReachesHandlerOnEverySurface` — target-scoped result plus the negative control, under `core`/`defer`, `full`/`hide`, and `facade-v1`.
- `TestAnalyzeAdvertisedTargetMatchesAcceptedTarget` — anti-drift over the whole analyze catalogue: a kind whose published request shape carries a target must accept one, and a kind whose shape omits it must refuse one.
- `TestAnalyzeUnsupportedTargetFailsClosed`, `TestInertSelectorsFailClosedAcrossFacades`, `TestAnalyzeImpactAcceptsSymbolBatchSelector`.
- `TestReconcileArgKeys_NeverMovesContainersIntoScalars`, `TestReconcileArgKeys_NeverInvertsMeaning`, `TestReconcileToolParams_PreservesFacadeEnvelope`.

`go test ./...` and `golangci-lint run ./...` are clean.

## Not changed

Several analyze kinds (`annotation_users`, `field_writers`, `external_calls`, `ref_facts`, …) read `id` as a row filter, but neither their schema annotation nor `capabilities` publishes it, so `target:{symbol}` is refused there rather than silently honoured. That refusal is loud and points at the accepting kinds; the legacy `id` field keeps working. Publishing those filters as selectors is a separate change.


tiendungdev:
Thanks for digging past what I could see from outside — and for saying plainly that the surface-negotiation theory was wrong. **Retracting it**: I inferred it from the `def_use` error speaking legacy vocabulary (`id`/`ids`) plus the coarse `freeObject("target")` schema my client received, and I read that as evidence about which surface had been bound. Your reproduction on `facade-v1` rules that out. The reconciler renaming `target` onto a declared string `id` explains every observation I had, including the `options:{id}` asymmetry, without needing preset negotiation at all — and it explains it better, because it predicts the same failure on surfaces I never tested.

The part worth keeping from my side is only the boundary: not `applyFacadeTarget`, not the impact handler. I should have stopped there rather than naming a layer I could not read.

Two things in your write-up land harder for us than the original bug:

**The `context` → `edit.content` move.** That is a write path. Our agent profile mandates mutating through `edit`, so an envelope key silently becoming the text an edit writes is a different severity class from a wrong ranking. We have added a local fail-closed rule (no `context` envelope on `edit`, read back the file after any tool-driven edit) until the fix is in a tagged build.

**Filters accepted as their own negation.** `include_tests` → `exclude_tests` and `max_score` → `min_score` are the ones that worry me most, because the result is well-formed and plausible. A wrong ranking gets noticed; a result set that is the complement of the requested one gets pasted into a report. We are treating any `analyze` result that used those flags as untrusted for now.

Both are the same failure shape as the original: a helpful correction produces a valid request that answers a different question. The fix making an alias candidate declare a type that can hold the value is the right cut — it removes the class, not the instance.

**Offer:** the corpus that surfaced all three symptoms is a two-repo Windows workspace, 4,853 files / 249,926 nodes / 1,311,204 edges, where one tracked repo has a built contract tier and the other does not, and it reproduces `unsupported_target` on `analyze routes`, the nested-selector caller gap, and the impact target drop. If a real-world verification pass on `#504` before or after merge is useful, I am happy to run your exact controls on it and post results — including the negative controls, and including anything that does *not* reproduce.

One small thing I could not check from here: `TestAnalyzeAdvertisedTargetMatchesAcceptedTarget` sounds like it closes the drift I hit from the other direction too — `capabilities` advertising a shape the runtime refuses. Does it also cover the `routes` case where the published schema has no `target` at all, or only the kinds whose shape does carry one?

zzet:
Follow-up on the parity gate's coverage, since it is a fair thing to ask of a test that claims to close drift.

`TestAnalyzeAdvertisedTargetMatchesAcceptedTarget` asserts **both** directions, and the no-target direction is the one this issue tripped over:

```go
if publishesTarget {
    require.Falsef(t, refused, "capabilities publishes target for analyze(kind=%s) but the dispatcher refuses it", ...)
    continue
}
require.Truef(t, refused, "analyze(kind=%s) does not publish a target but accepted one — it would be silently ignored", ...)
```

So `routes` is covered by the second branch, alongside 71 other kinds that publish no target.

It did have a boundary, though: it filtered to kinds the analyze dispatcher handles itself, skipping the 21 that dispatch to a different legacy tool. Measured buckets before removing that filter:

| bucket | count |
|---|---|
| publishes target, accepts it | 2 (`impact`, `def_use`) |
| publishes nothing, refuses | 72 (incl. `routes`) |
| **aliased — skipped by the gate** | **21** |

Removing the filter found three kinds accepting a target nobody advertises: `communities`, `processes`, `why`. Only `why`'s is a symbol — `get_communities` and `get_processes` take an `id` naming a community or a process, so a symbol lowered into it matches nothing and returns an empty answer. Declaring the field is not evidence the field means what the selector meant.

Pushed in 7be71ee: those two refuse a symbol target, `why` publishes and accepts one, and the gate now runs over the whole catalogue with `require.Greater(t, aliased, 10)` so the filter cannot quietly come back. Full suite and lint clean.

tiendungdev:
Thanks — the both-directions assertion answers it, and the `aliased` bucket is a good catch to have made while the gate was open. `require.Greater(t, aliased, 10)` is the right shape for that: the filter can't come back quietly.

Two things from running this on our corpus, one of which contradicts part of the description. Measured on the released `v0.63.1+4f2d169`, Windows, two-repo workspace, 4,853 files / 249,926 nodes / 1,311,204 edges.

**`communities` with a symbol target returns the full list, not an empty one.**

```
analyze(kind:"communities")                                  -> 1366 communities
analyze(kind:"communities", target:{symbol:"…::setupOutboxDispatcher"})
                                                             -> identical 1366, byte for byte
analyze(kind:"processes")                    vs  same + target:{symbol}
                                                             -> identical 50-row list
```

Positive control that the `id` channel itself is fine when given the right entity:

```
analyze(kind:"communities", options:{id:"community-0"})
  -> cohesion 0.75, hub fillCredentials, exactly 5 members
```

I read your "matches nothing and returns an empty answer" as describing the state *after* lowering works — once a symbol id actually reaches `get_communities`, it matches no community. On the released build the reconciler destroys `target` before it can be lowered into `id`, so no filter is applied at all and the caller gets everything. If that reading is right, the two symptoms bracket the fix rather than contradict it, and `7be71ee` closes both. Worth knowing which one a bug report is describing, though — an unfiltered result is the more deceptive of the two, since it reads as a successful broad answer rather than as nothing found.

**A correction to something I nearly told you.** I was about to report that this defect is exposed by your own shipped skills, since `gortex-architecture-review` and `gortex-onboarding` both build on `get_communities` / `get_processes`. I read the skill files instead of their descriptions before sending it, and that claim is wrong: `gortex-onboarding/SKILL.md` instructs `get_communities({id: <community-id>})` and `gortex-architecture-review` never pairs those kinds with a symbol selector. **Both use the correct entity type.** The real exposure is narrower — an agent improvising `target:{symbol}` against the generic free-form `target` that `analyze` advertises in `tools/list`. Which is exactly what your parity gate now prevents from being silently honoured.

One small observation while measuring, not a bug: `analyze(kind:"communities", options:{repo:…})` widens to the whole workspace but **says so** —

```
scope.note: kind 'communities' is clamped to the session workspace but is not
            repo/project-narrowed in v1; the requested narrowing was widened
```

That disclosure is the difference between a caveat and a wrong answer, and it is the behaviour the rest of this issue has been about. Please keep it when `communities` eventually does get repo narrowing.

Happy to re-run any of the above on `#504` once it merges, including whichever controls you would want to see fail.

zzet:
Your reading is right, and my justification was wrong — thanks for measuring it.

I claimed `communities` with a symbol target "matches nothing and reads as an empty answer". Measured on this branch, it does neither:

```
analyze(kind:"communities")                            -> full list, no error
analyze(kind:"communities", options:{id:"main.go::helper"})
                                                       -> ERROR "community not found: main.go::helper"
analyze(kind:"processes",   options:{id:"main.go::helper"})
                                                       -> ERROR "process not found: main.go::helper"
```

`handleGetCommunities` returns `NewToolResultError("community not found: " + id)` on an unmatched id — loud, never empty. So the three states are:

| build | `target:{symbol}` on `communities` |
|---|---|
| `v0.63.1` released | selector destroyed pre-lowering → **unfiltered full list, no error** (what you measured) |
| lowering fixed, refusal absent | lowered into `id` → **`community not found: <symbol>`** |
| this branch | structured refusal naming the entity mismatch |

So your bracket is exactly right, with one amendment: the middle state was never the silent one I described. That also means the refusal does not earn its keep by preventing silence — the lowering fix alone removes the unfiltered answer. It earns it on two narrower grounds: `capabilities` never advertised a target for these kinds, so accepting one re-opens the advertise/accept drift this gate exists to close; and `community not found: pkg/foo.go::Bar` blames the id channel for what is really a selector-semantics mistake.

Pushed in 9db7501:

- the code comment now states the real behaviour rather than the invented one;
- the refusal stops giving every caller the same reason. `routes` (no field at all) keeps "it would be ignored"; `communities` / `processes` get "selects by id, which does not name a symbol — target.symbol would be resolved against the wrong entity", plus `entity_mismatch: "id"` in the structured data. Writing that test immediately caught a bug in my own discriminator: the analyze dispatcher declares `id` once for all 70+ kinds behind it, so "the legacy tool declares the field" wrongly classified `routes` as an entity mismatch. Only a kind dispatching to its own legacy tool can be in that second case.

And noted on the skills — appreciated the correction before it became a bug report. Agreed on `scope.note`; nothing in this PR touches the widening disclosure, and the parity gate is the same principle applied to selectors, so it should stay when `communities` gets repo narrowing.

Full suite and lint clean. Re-runs on your corpus welcome once this merges — the control I would most like to see is `target:{symbol}` on a few of the 17 aliased kinds that should now refuse, since my fixture only exercises `clones` end to end.

tiendungdev:
Baseline for the control you asked for, taken now on the **released** build so you have the "before" half while the branch is still moving. I will re-run the same set on `#504` after it merges.

Build `v0.63.1+4f2d169`, Windows, two-repo workspace: 4,853 files / 249,926 nodes / 1,311,204 edges. Method: identical call with and without `target:{symbol:"…::setupOutboxDispatcher"}` (a real, indexed, unambiguous Go function), outputs compared.

| kind | group | with vs without `target:{symbol}` |
|---|---|---|
| `impact` | publishes a target | 100-row repo-wide ranking, identical |
| `routes` | publishes none | 153 rows, identical |
| `communities` | aliased | 1366, identical |
| `processes` | aliased | 50, identical |
| `clones` | aliased | 385 clusters / 1402 pairs, identical |
| `untested` | aliased | identical |

No kind errored. So on the released build the selector is dropped across all three groups, and the aliased group is not special — it is the same reconciler rename, one layer earlier than where the entity-mismatch question even arises.

`untested` is the most useful row of the six, because its answer is not a row count:

```
analyze(kind:"untested")                        coverage_ratio: 0.594105480868666
                                                total_candidates: 13538  total_uncovered: 5495
analyze(kind:"untested", target:{symbol:…})     coverage_ratio: 0.594105480868666
                                                total_candidates: 13538  total_uncovered: 5495
```

An identical float to the last digit rules out the reading that a filter applied and happened to return a similar-sized page — nothing was filtered at all. Worth borrowing for a fixture: a scalar-bearing kind fails this way more legibly than a list-bearing one, where truncation can disguise it.

On scope: I derived your 17 independently before measuring, to be sure we mean the same set — `facade_registry.go` lists 21 aliased kinds; minus `communities` / `processes` / `why`, which you handle separately, and `contracts`, whose legacy name is unchanged, leaves 17. Agreed set.

Since your fixture only drives `clones` end to end: `clones`, `communities`, `processes` and `untested` all currently drop the selector, so at least four of the aliased kinds exercise the path, not one. Happy to run the remaining thirteen after merge, or a specific subset if some are more interesting than others — `co_change` and `replay` I left alone on purpose, since one mines git and the other replays an episode.

One adjacent observation, not part of this PR and not a complaint: `analyze(kind:"communities", options:{repo:…})` widens to the whole workspace and discloses it in `scope.note`. That disclosure is the reason I could tell a widened answer from a narrowed one while measuring — it is doing the same job your parity gate does, one level up.

zzet:
The baseline is exactly the right shape, and the `untested` row earned its keep twice — once as evidence, once as a test design. But the set is not agreed: we derived different 17s, and yours would mis-predict two of the controls you are about to run.

**Measured on the branch, not derived.** Driving `target:{symbol}` at every aliased kind through real frames:

```
ALIASED TOTAL: 21
ACCEPTS  (2): co_change, why
REFUSES (19): agent_config architecture churn citation clones communities contracts
              coupling extraction health inspection_catalog inspections knowledge_gaps
              lint processes recent_changes replay surprising_connections untested
```

Two differences from `21 − 3 − 1`:

- **`contracts` refuses.** Its legacy name being unchanged is not what the code keys on — the discriminator is `spec.Legacy != "analyze"`, and `analyze.contracts` maps to the `contracts` tool, so it is aliased like the rest. Subtracting it drops a kind that should be in your control set.
- **`co_change` accepts, and always did.** It is the one aliased kind `capabilities` already published a target for (`find_co_changing_symbols` takes `symbol_id`). Including it in a "should refuse" set would produce a false failure.

So post-merge the expectation is 19 refuse / 2 accept, and `communities` / `processes` sit in the refusing 19 rather than outside the set. You are right that the aliased group was not special on the released build — the reconciler rename got there first, which is why all six of your rows are identical. The aliased group only becomes distinguishable once lowering works, which is the state your re-run will measure.

**Borrowed your fixture idea, generalised.** You are right that a scalar is more legible than a row count, and my fixture had the exact weakness you describe from the other direction: repo-wide and targeted both returned two rows, so `total` could not have caught "accepted then not applied". Pushed in b537ba4 — the fixture now contains a symbol outside the target's dependent closure, and the test asserts the two calls disagree on both totals and membership:

```go
require.Less(t, targetedTotal, wideTotal,
    "an identical total means the target was accepted and then not applied")
require.NotContains(t, ids(targeted), "main.go::isolated")
```

That is your `coverage_ratio` control as a regression test: identical scalars fail it.

On the remaining thirteen — `co_change` and `replay` are the right two to leave alone, and I would not spend your time on the other eleven individually. The highest-value controls after merge are the two that changed classification (`communities`, `processes` — expect a refusal naming `entity_mismatch: "id"`, not a full list), `contracts` (expect a refusal, since our sets disagreed there), and `co_change` (expect it to *work*, not refuse). If those four land as predicted, the remaining fifteen share one code path with `clones` and `untested`.

Full suite and lint clean.

tiendungdev:
Ran the controls on merged `main` (`2cfeb76e`, includes `9db7501`), built locally, on a throwaway two-repo workspace — no private data this time, just a synthetic fixture.

**All four land as you predicted.**

```
analyze(kind:"communities", target:{symbol:…})
  -> invalid_argument / unsupported_target
     "selects by id, which does not name a symbol — target.symbol would be
      resolved against the wrong entity, so the call is refused"
     entity_mismatch: "id"   accepted_by: [co_change, def_use, impact, why]

analyze(kind:"processes",  target:{symbol:…})   -> same shape, entity_mismatch: "id"

analyze(kind:"contracts",  target:{symbol:…})
  -> invalid_argument / unsupported_target
     "has no target.symbol to act on — it would be ignored, so the call is
      refused rather than answered as if it had not been sent"
     (no entity_mismatch key)

analyze(kind:"co_change",  target:{symbol:…})
  -> WORKS: {"symbol_id":"…::alphaHandler","target_file":"…/main.go","total":0}
```

The discriminator split you added in `9db7501` is visible from the outside: `contracts` gets "would be ignored" with **no** `entity_mismatch`, `communities`/`processes` get the entity-mismatch wording **with** the key. That is the distinction that would have been wrong under the "legacy tool declares the field" test, so the fix is observable, not just internal.

`accepted_by` in the refusal is a nice touch — it answers "then what does take one" without a second call.

**The original #449 defect is fixed**, including the control that exposed it:

```
analyze(kind:"impact", target:{symbol:"…::alphaHandler"})
  -> "scope":"target_closure", target.dependents: 6      (was: repo-wide ranking)

analyze(kind:"impact", target:{symbol:"NO_SUCH_SYMBOL_9f68c2a1"})
  -> symbol_not_found: "impact target … is not an indexed symbol in this scope"
                                                          (was: same repo-wide
                                                           ranking, no error)
```

That negative control is the one I care about most: a wrong target now fails loudly instead of returning a confident answer to a different question.

**Nothing failed to reproduce.** I did not run the other thirteen aliased kinds, per your note that they share a code path with `clones` and `untested`.

---

One request, if it is reasonable: **could you cut a release with these in it?**

`v0.63.1` is still the newest tag, so everything above only exists on `main`. On my side the practical effect is that the mitigations I put in place stay in place — my team's rules currently say not to pass a `context` envelope to `edit` and not to trust an `analyze` result that used `include_*`/`exclude_*`/`min_score`/`max_score`, and my own policy is to lift those only after the fix ships in a **tagged** build and the controls are re-run against it. I would rather re-run against a tag than pin a hand-built binary.

No urgency implied — you may well be batching. I am only asking because #504 fixes a class where a wrong call returns a well-formed answer, and that is unpleasant to leave in the newest release for people who do not know to look for it.

(Separately: #507 is up for the routes repo-narrow zero from #506 — same fixture, the shared route-contract node deciding repo visibility for handlers that are not its owner.)

PR #536, opened 2026-08-11
fix(csharp): resolve calls with explicit type arguments and same-name extension overloads

### LINKED ISSUE #534: csharp: extension-method calls resolve only within the declaring file — every cross-file call is dropped
**Version:** `v0.63.2+6a2905e` (windows/amd64, SHA-256 `511855e4367d6e3357621768f2d9fd10e45b367808f734cae1a1bb5d987edea8`, verified against the release `checksums.txt`)

Following up on the C# extension-binding work in #432 / #519 / #224. That work does resolve extension calls — but only when the call site and the extension declaration live in the **same file**. Any extension call made from a different file produces **no `CALLS` edge at all** — not an `Unresolved` stub, nothing. The calling method simply has zero outgoing call edges.

This matters because the dominant real-world shape for extension methods is a library declaring them in one file and consumers calling them from another. `gortex query callers` on such a method returns `total_edges: 0` with the `likely_unused` caveat, which is easy to read as "dead code."

## Minimal reproducer

Four files, no NuGet, no build required.

<details>
<summary><code>Lib/Extensions/ServiceCollectionExtensions.cs</code></summary>

```csharp
using System;

namespace Lib.Extensions
{
    public interface IServiceCollection { }

    public class DecoratorOptions { public int RetryCount { get; set; } }

    public static class ServiceCollectionExtensions
    {
        public static IServiceCollection AddFooDecorator<TInterface, TImpl>(
            this IServiceCollection services)
            where TInterface : class
            where TImpl : class, TInterface
        {
            services.AddFooDecoratorInfrastructure();   // line 25 — RESOLVES (same file)
            return services;
        }

        public static IServiceCollection AddFooDecorator<TInterface, TImpl>(
            this IServiceCollection services,
            Action<DecoratorOptions> configure)
            where TInterface : class
            where TImpl : class, TInterface
        {
            services.AddFooDecoratorInfrastructure();   // line 41 — RESOLVES (same file)
            return services;
        }

        public static IServiceCollection AddFooDecoratorInfrastructure(
            this IServiceCollection services) => services;
    }
}
```
</details>

Three call sites, each in a different file, all calling `AddFooDecorator`:

| Case | File | Project | Namespace |
|---|---|---|---|
| 1 | `Consumer/Extensions/ServiceCollectionExtensions.cs` | different | different (`using Lib.Extensions;`) |
| 2 | `Lib/Extensions/SameNamespaceCaller.cs` | same | **same** (`Lib.Extensions`, no `using` needed) |
| 3 | `Lib/Other/DifferentNamespaceCaller.cs` | same | different (`using Lib.Extensions;`) |

```csharp
// Case 2 — same project, same namespace, different file
namespace Lib.Extensions
{
    public interface IDelta { }
    public class Delta : IDelta { }

    public static class SameNamespaceCaller
    {
        public static IServiceCollection Register(this IServiceCollection services)
        {
            services.AddFooDecorator<IDelta, Delta>();
            return services;
        }
    }
}
```

## Observed

Outgoing calls from each of the three callers — **all zero**:

```
$ gortex query calls ".../Consumer\Extensions\ServiceCollectionExtensions.cs::ConsumerRegistration.AddConsumerServices" --format json
  "total_edges": 0
$ gortex query calls ".../Lib\Extensions\SameNamespaceCaller.cs::SameNamespaceCaller.Register" --format json
  "total_edges": 0
$ gortex query calls ".../Lib\Other\DifferentNamespaceCaller.cs::DifferentNamespaceCaller.Register" --format json
  "total_edges": 0
```

Correspondingly, `query callers` on the extension method reports it unused:

```json
{
  "caveat": {
    "class": "likely_unused",
    "message": "no incoming call or reference edges, but the symbol is indexed ... this is evidence of no callers, not proof: confirm with a text search for the symbol name before removing it."
  },
  "edges": [],
  "total_edges": 0,
  "total_nodes": 1
}
```

`query usages` on the same symbol also returns `total_edges: 0`.

The **intra-file control** in the same index resolves fine:

```json
{
  "edges": [
    { "from": "...ServiceCollectionExtensions.AddFooDecorator",     "to": "...AddFooDecoratorInfrastructure",
      "line": 25, "kind": "calls", "origin": "ast_inferred", "tier": "heuristic" },
    { "from": "...ServiceCollectionExtensions.AddFooDecorator_L34", "to": "...AddFooDecoratorInfrastructure",
      "line": 41, "kind": "calls", "origin": "ast_inferred", "tier": "heuristic" }
  ]
}
```

## Expected

Cases 1–3 should each produce a `CALLS` edge to `AddFooDecorator`. Case 2 in particular is a plain same-namespace call with no import indirection at all, so a namespace-visibility gate (#432, #519) should already admit it.

Failing that, an `Unresolved` stub would at least be honest — silently emitting no edge makes an unused-code query indistinguishable from a resolution failure.

## Notes

- **Overload sensitivity in a larger codebase.** In a private production repo we see one of two near-identical overloads resolve its intra-file self-call while the sibling overload does not. The minimal fixture above does **not** reproduce that asymmetry — both overloads resolve there — so there is some additional trigger we have not isolated. Mentioning it only as a hint; the cross-file failure above is the reproducible part.
- **Possibly spurious edge.** In the fixture, `AddFooDecoratorInfrastructure` also gains a third inbound edge `from: ...AddFooDecorator_L34` at `line: 34`, which is overload B's *declaration* line rather than a call site. Minor, and possibly an artifact of the `_L<line>` overload disambiguation.

zzet:
Fixes #534.

## What the report got right, and what it turned out to be

The reporter's observation is exact — every cross-file call to their extension method produced `total_edges: 0` — but the cause is not cross-file resolution. Running their fixture through the real extractor shows the three failing call sites never reach the resolver at all:

```
=== EDGES BEFORE RESOLVE ===
ServiceCollectionExtensions.AddFooDecorator_L20 -> unresolved::*.AddFooDecoratorInfrastructure  line=26
ServiceCollectionExtensions.AddFooDecorator     -> unresolved::*.AddFooDecoratorInfrastructure  line=16
                                              (nothing from the three callers)
```

The variable is not the file. It is the **type arguments**. Their intra-file control calls `services.AddFooDecoratorInfrastructure()`; all three failing sites call `services.AddFooDecorator<I, T>()`. Two independent defects sit behind that, and both had to be fixed to close the issue.

## 1. Generic invocations emitted no call edge at all

The C# call-site query pinned every invoked name to `(identifier)`. A call spelling explicit type arguments parses that name as a `generic_name`, so it matched **no pattern** — no edge, not even an unresolved stub. The member-access patterns carried the same constraint, so the call earned no reference edge either, which is why `query usages` also reported zero.

This is not a corner case. `GetRequiredService<T>()`, `AddSingleton<TI, TImpl>()`, `Deserialize<T>()`, `Map<TDest>()` — generic invocation is the dominant call shape in .NET, so a large fraction of every indexed C# repo was invisible. It is also a recurring class here: `rust.go` carries a note that the turbofish version cost **5,301 unrecorded call sites**.

Each affected pattern now accepts either spelling and captures the inner identifier, so the callee name stays bare and every downstream tier is unchanged.

## 2. Same-name extension overloads could never be bound

With edges flowing, the fixture still resolved to nothing. Both `AddFooDecorator` overloads extend `IServiceCollection` and both live in `Lib.Extensions`, so neither the receiver-type tiers nor the namespace-visibility tie-break could separate them, and the binder's ambiguity refusal dropped all three sites. Overloaded extension methods over one interface is the standard .NET DI registration shape.

C# separates such a set by **applicability**, which is exact rather than heuristic: an overload is a candidate only if the call's argument count fits its parameter list and its type-parameter count matches any explicitly spelled type arguments. Neither fact was in the graph. Both are now recorded — argument and type-argument counts on the member-call edge, `param_count` / `param_required` / `param_variadic` on the method node — and the binder narrows on them before the visibility tie-break, matching §12.8.10.3, where each scope level considers only applicable candidates.

All three of the reporter's cases now bind to the one-parameter overload, and the intra-file control still binds.

### Static-form calls, and why they mattered

An extension can also be invoked through its declaring class — `BagExt.Add(bag)` — where the receiver is passed as the *first argument*, so the argument list has to cover every parameter. Comparing such a call against the extension-form window shifts it by one, and with a sibling overload present that does not merely lose the bind: it leaves the **wrong** candidate as the sole survivor, and narrowing then binds it. I hit this while reviewing my own patch, on a probe rather than a test.

The resolver had no way to tell the forms apart, so the extractor now stamps a bare receiver's spelling — but only in the one case nothing else explains it, after the local, parameter and builtin lookups have all missed. That restriction is what makes the signal trustworthy: reaching it means no value in scope carries the name, so a receiver equal to a candidate's declaring class is a type reference rather than a variable shadowing it. Beyond removing the misbinding, this also resolves static-form calls correctly, which previously stayed unresolved on any overloaded set.

### Narrowing only, and never past the visibility veto

Missing evidence, an unreadable stamp, or a filter that would empty the set all leave the candidates untouched, so this can turn a refusal into a bind but never a bind into a different one. Where it *does* narrow, the winner must still clear the visibility veto: the pool-unique rule waives visibility deliberately, but that waiver is scoped to sets applicability never touched, so an overload the call site cannot see is refused rather than promoted by arity. `ArityNeverBeatsVisibility` pins exactly this — I hit it as a real regression in an earlier draft of the patch.

## 3. The reporter's second note, same root cause

Their Note about an edge anchored at a *declaration* line rather than a call site is real, and it is the same `<` blind spot. The function-as-value capture treats a `(` after an identifier as proof the identifier is a callee, and relies on that to skip a declaration's own name. `AddFooDecorator<TInterface, TImpl>(` puts the type parameters where the `(` is looked for, so the method's own name was captured as a function reference at its declaration line, and the gate then bound it to the first same-name sibling — an edge between two overloads at a line where no call exists.

The byte rule cannot simply learn `<`: in JavaScript, TypeScript and every C-family language the capture serves, `f < g` is a comparison. The tree answers exactly and for all of them at once — a node filling its parent's `name` field where that parent also declares `parameters` is a declaration however it is spelled.

## Also fixed along the way

- **Existing indexes**: `extractorVersions["csharp"]` is bumped, so already-indexed `.cs` files re-extract instead of silently keeping the edge-less graph. Razor templates are salted with C# too, since they are extracted by the C# extractor.
- **Parameter positions**: a discard (`Foo(int _, string name)`) emits no param node but still occupies its slot; skipping the slot along with the node renumbered every later parameter, so `name` reported position 0.

## Two grammar traps the tests pin

- The vendored grammar does **not** resolve a `params object[] rest` entry into a `parameter` node — it flattens it into loose siblings. A naive count claims one parameter for a two-parameter method, which would exclude that overload from argument counts it really accepts. A parameter list that cannot be read in full now yields no evidence at all, leaving the method universally applicable.
- A defaulted parameter has no equals-value wrapper node, only a bare `=` token, so optional parameters were invisible to the obvious check.

## Verification

`go test ./internal/...` passes in full; `go test -race` on the three touched packages passes; `golangci-lint` reports 0 issues.

New coverage, all failing before the change:
- the reporter's four-file fixture, asserting all three cases bind and the declaration reports three incoming callers;
- each call shape (typed local, null-conditional, `this`-, `base`-qualified, receiverless) paired with and without type arguments, asserting identical behaviour — the property that was actually violated;
- argument count picking each overload in turn, an optional parameter correctly *keeping* the ambiguity, a `params` list widening it, type-argument count splitting `Map<T>` from `Map<TIn, TOut>`, and arity not overriding visibility;
- extension form and both static forms of an overloaded extension each landing on the right overload;
- a generic declaration minting no function-value reference at its own line.

## Not in this PR

The same class of blind spot is present in **`scala.go`, `golang.go`, `cpp.go` and `swift.go`** — verified by running each real extractor and comparing the plain and generic spellings of the same call, not by reading queries. Go has two distinct mechanisms behind it (`index_expression`, and `type_conversion_expression` when a single type argument meets exactly one value argument), and C++ additionally drops every `ns::`-qualified call. Filed separately rather than bundled here, since each needs its own grammar handling, version bump and fixtures.

### DIFF
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -32,7 +32,7 @@ var extractorVersions = map[string]int{
 	//   "go": 2,
 	"c":      generatedParserProjectionPolicyVersion, // generated parser projection covers all strictly detected table sizes
 	"php":    2,                                      // class/interface inheritance now emits typed structural edges
-	"csharp": 8,                                      // interface base lists emit extends edges (was: record positional property nodes)
+	"csharp": 9,                                      // generic invocations emit call edges (was: interface base lists emit extends edges)
 }
 
 // extractorSaltExtLang maps a lower-case file extension to the language
@@ -41,41 +41,47 @@ var extractorVersions = map[string]int{
 // staleness, the pre-existing behaviour). Extensions are grouped to the
 // extractor that owns them.
 var extractorSaltExtLang = map[string]string{
-	".go":    "go",
-	".py":    "python",
-	".pyi":   "python",
-	".js":    "javascript",
-	".jsx":   "javascript",
-	".mjs":   "javascript",
-	".cjs":   "javascript",
-	".ts":    "typescript",
-	".tsx":   "typescript",
-	".mts":   "typescript",
-	".cts":   "typescript",
-	".java":  "java",
-	".rb":    "ruby",
-	".rs":    "rust",
-	".c":     "c",
-	".h":     "c",
-	".cc":    "cpp",
-	".cpp":   "cpp",
-	".cxx":   "cpp",
-	".hpp":   "cpp",
-	".hh":    "cpp",
-	".cs":    "csharp",
-	".php":   "php",
-	".swift": "swift",
-	".kt":    "kotlin",
-	".kts":   "kotlin",
-	".scala": "scala",
-	".m":     "objc",
-	".mm":    "objcpp",
-	".lua":   "lua",
-	".dart":  "dart",
-	".ex":    "elixir",
-	".exs":   "elixir",
-	".sh":    "bash",
-	".bash":  "bash",
+	".go":   "go",
+	".py":   "python",
+	".pyi":  "python",
+	".js":   "javascript",
+	".jsx":  "javascript",
+	".mjs":  "javascript",
+	".cjs":  "javascript",
+	".ts":   "typescript",
+	".tsx":  "typescript",
+	".mts":  "typescript",
+	".cts":  "typescript",
+	".java": "java",
+	".rb":   "ruby",
+	".rs":   "rust",
+	".c":    "c",
+	".h":    "c",
+	".cc":   "cpp",
+	".cpp":  "cpp",
+	".cxx":  "cpp",
+	".hpp":  "cpp",
+	".hh":   "cpp",
+	".cs":   "csharp",
+	// Razor templates are extracted by the C# extractor over their
+	// embedded code blocks, so a C# extraction fix has to re-flag them
+	// too — otherwise the same stale extraction survives on exactly the
+	// files whose C# is hardest to re-trigger by hand.
+	".razor":  "csharp",
+	".cshtml": "csharp",
+	".php":    "php",
+	".swift":  "swift",
+	".kt":     "kotlin",
+	".kts":    "kotlin",
+	".scala":  "scala",
+	".m":      "objc",
+	".mm":     "objcpp",
+	".lua":    "lua",
+	".dart":   "dart",
+	".ex":     "elixir",
+	".exs":    "elixir",
+	".sh":     "bash",
+	".bash":   "bash",
 }
 
 // ExtractorLangForFile returns the extractor-staleness language key for a
--- a/internal/parser/languages/csharp.go
+++ b/internal/parser/languages/csharp.go
@@ -670,6 +670,16 @@ func (e *CSharpExtractor) extractCSharp(filePath string, src []byte) (*parser.Ex
 				}
 			} else if strings.Contains(c.receiver, ".") || strings.Contains(c.receiver, "(") {
 				stampFactoryChainReceiver(edge, c.receiver, resolveChainType(c.receiver, tenvByOwner[callerID], result))
+			} else if c.receiver != "" {
+				// A bare receiver nothing above could type. Its spelling
+				// is still evidence: reaching here means no local, param
+				// or builtin in scope carries that name, so a receiver
+				// that names a static class is the STATIC form of an
+				// extension call (`BagExt.Add(bag)`) — where the `this`
+				// slot is filled by the first argument, not the
+				// receiver. The extension binder needs that distinction
+				// before it can compare argument counts.
+				edge.Meta = map[string]any{"receiver_name": c.receiver}
 			}
 			// Eviction restubs a member call to a bare unresolved name; the
 			// marker is what lets the resolver still route the rebind through
--- a/internal/parser/languages/csharp.go
+++ b/internal/parser/languages/csharp.go
@@ -71,41 +71,69 @@ const qCSharpAll = `
 
   (using_directive (_) @using.path) @using.def
 
+  ; An invocation that spells explicit type arguments parses its callee
+  ; name as a generic_name, never a bare identifier — so pinning the name
+  ; to (identifier) dropped every generic call site outright, with no edge
+  ; and no unresolved stub. That is the dominant .NET call shape
+  ; (GetRequiredService<T>(), AddSingleton<TI, TImpl>()), and it left
+  ; heavily-called methods looking like dead code. Each alternation
+  ; captures the inner identifier, so the callee name stays bare.
   (invocation_expression
-    function: (identifier) @call.name) @call.expr
+    function: [
+      (identifier) @call.name
+      (generic_name (identifier) @call.name)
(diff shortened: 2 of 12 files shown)


zzet:
Follow-up for the sibling-language findings: #537

zzet:
Sibling-language fixes are now up as #538. The two PRs touch the same `extractorVersions` map, so whichever merges second needs a one-line conflict resolution.

PR #538, opened 2026-08-11
fix(parser): emit call edges for generic calls in Scala, Go, C++ and Swift

### LINKED ISSUE #537: Calls with explicit type arguments are dropped by four more extractors (scala, go, cpp, swift)
While fixing #534 (C# calls with explicit type arguments emitted no call edge, because the query pinned the callee name to `identifier` and a generic call parses it as `generic_name`) I audited every other extractor for the same class of blind spot: **a call-site pattern that constrains the callee to a plain identifier node, so adding type arguments makes the whole call edge disappear.**

Four more extractors are affected. Each verdict below was produced by running the real extractor over a probe fixture and comparing the plain and generic spellings of the same call, so these are observed edge counts, not query readings.

This is a recurring class here — `rust.go` already carries a note that the turbofish version of it cost **5,301 unrecorded call sites** before it was fixed.

### `scala.go` — `generic_function`
`extractCall` (scala.go:615-637) switches on `node.Child(0)` and handles only `identifier` and `field_expression`, falling through to `default: return`. tree-sitter-scala wraps any explicit-type-argument callee in a `generic_function` node.

```
(call_expression function: (generic_function function: (identifier) type_arguments: (...)) arguments: (...))
```

Probe over 6 call sites emitted edges for only `plainCall` and `methodPlain`. `genericCall[String]()`, `genericTwo[String,Int](x)`, `obj.methodGeneric[String](x)` and `Future[Int] { 1 }` produced **no call edge**.

### `golang.go` — `index_expression` *and* `type_conversion_expression`
golang.go:48-55 pins `function:` to `(identifier)` / `(selector_expression)`. Two distinct mechanisms drop calls, and a fix that only handles the first leaves the second broken:

- **single type argument, 0 or 2+ value args** — `F[int]()`, `obj.M[int](a,b)` parse as `function: (index_expression ...)`.
- **single type argument, exactly 1 value arg** — `F[int](a)`, `obj.M[int](a)` parse as `type_conversion_expression` / `generic_type`, not `call_expression` at all.

Multi-type-argument calls (`F[string,int](a,b)`) are **not** affected: they keep `function: (identifier)` with a sibling `type_arguments:` field and emit normally.

### `cpp.go` — `template_function` / `template_method`, plus a non-generic gap
cpp.go:71-76 matches only `function: (identifier)` and `field: (field_identifier)`.

- `foo<T>()` → `function: (template_function name: (identifier) ...)`
- `obj.foo<T>()`, `ptr->foo<T>()`, `this->foo<T>()` → `field: (template_method name: (field_identifier) ...)`

Six probe call sites, only the non-templated `plain()` emitted an edge. The free-function form still emits a spurious `unresolved::fnvalue::` references edge in place of the call.

**Separately (not generics):** a namespace-qualified call `ns::foo()` parses as `function: (qualified_identifier ...)` and matches nothing either, so all `std::` / `ns::` free-function calls are dropped. `rust.go:82-84` already carries the equivalent `scoped_identifier` pattern.

### `swift.go` — generic call reroutes to `constructor_expression`
`qSwiftAll` (swift.go:46-51) has `(call_expression (simple_identifier) @call.name)` and no `constructor_expression` pattern (`rg constructor_expression` has no non-test hits in `internal/`).

Adding type arguments reroutes **any** call — uppercase, lowercase, and member forms — to `constructor_expression`, which kills both the `calls` edge and the `instantiates` edge:

- `UpperPlain()` → `call_expression` → `calls` + `instantiates`
- `Array<Int>()`, `Dictionary<String,Int>()`, `MyType<Int>(x: 1)`, `obj.doThing<Int>()`, `lowerGeneric<Int>()` → `constructor_expression` → **zero edges**

## Verified clean

Probe-confirmed unaffected, with the reason each is immune:

| Extractor | Why it is safe |
|---|---|
| `typescript.go` (.ts + .tsx) | `type_arguments` is a sibling field of `function:`, so the callee stays `(identifier)` / `(member_expression … property:)` |
| `java.go` | explicit type args live in a separate `type_arguments` field of `method_invocation`; `name: (identifier)` is intact |
| `kotlin.go` | type arguments sit inside `call_suffix`, not around the callee |
| `rust.go` | already fixed — three dedicated `generic_function` patterns |
| `dart.go` | the selector walk keys on `argument_part`, and `type_arguments` sit inside it |
| `c.go`, `php.go` | no generic call syntax (PHP dispatches on `ChildByFieldName("name")` anyway) |
| `csharp.go` | fixed in #534 |

## Why this matters

The failure mode is the one that makes #534 expensive rather than merely incomplete: the call emits **no edge at all**, not an unresolved stub. `query callers` then answers `total_edges: 0` with the `likely_unused` caveat, so a resolution gap is indistinguishable from dead code — and for Go, Scala and C++ the affected syntax is ordinary generic code, not a corner.

## Suggested approach

Per language, widen the call pattern to accept the generic callee node and capture the inner identifier — the shape #534 used for C#:

```
function: [
  (identifier) @call.name
  (generic_name (identifier) @call.name)
]
```

Each fix should bump that language's `extractorVersions` entry (`internal/indexer/extractor_version.go`) so already-indexed files re-extract, and should land with a paired test asserting the plain and generic spellings of the same call behave identically — that pairing is what makes the regression impossible to reintroduce quietly.

zzet:
Fixes #537.

Four more extractors share the blind spot found in #534: a call query that pins the callee to a plain identifier node drops **every call spelling explicit type arguments**, because those grammars wrap the callee in a different node. The call emits *no edge at all* — not an unresolved stub — so `query callers` answers `total_edges: 0` with the `likely_unused` caveat, and a resolution failure is indistinguishable from dead code.

This is a recurring class here: `rust.go` carries a note that the turbofish version of it cost **5,301 unrecorded call sites**.

Every shape below was pinned by dumping the real parse tree and confirmed by running the real extractor before and after — no query was fixed on inference. One commit per language, each with its own `extractorVersions` bump so already-indexed files re-extract instead of keeping the edge-less graph.

## Scala — `generic_function`

`extractCall` switched on the callee and handled only `identifier` / `field_expression`, falling through to `default: return`. Any callee with type arguments is wrapped:

```
(call_expression function: (generic_function function: (identifier) type_arguments: (…)) arguments: (…))
```

`f[T]()`, `obj.m[T](x)` and `Future[Int] { … }` all emitted nothing. Unwrapping to the real callee lets the two existing cases apply unchanged.

## Go — two mechanisms, not one

The patterns pinned `function:` to `(identifier)` / `(selector_expression)`. A Go call with **one** type argument matches neither, via two different routes — fixing only the obvious one leaves half the calls missing:

| Written | Parses as |
|---|---|
| `Zero[int]()`, `obj.M[int]()` | `function: (index_expression …)` |
| `OneArg[int](1)`, `obj.N[int](1)` | `type_conversion_expression` — **not a call at all** |

Two or more type arguments were never affected (`F[string, int](a, b)` keeps a plain callee with a sibling `type_arguments` field), which is why the gap looked like it didn't exist.

**Both shapes are genuinely ambiguous in the grammar.** An `index_expression` callee is also how `handlers[key]()` parses, and a conversion to a generic type is spelled exactly like a one-argument generic call. Widening therefore emits an unresolved stub for those too.

That is safe for a reason specific to Go rather than luck: **a package-scope name is a func, a var or a type — never two of them.** A stub named after a variable or a type finds no function candidate and stays unresolved, and an unresolved target is a placeholder reverse walks ignore. `TestResolveGo_IndexAndConversionDoNotBindWrongly` asserts no call edge in that fixture ever lands on a non-callable node, and its sibling asserts a real cross-file generic call still resolves.

## C++ — three of five callee shapes were missing

The query matched a bare `identifier` and a `field_expression` with a `field_identifier`. C++ has five:

- `generic<int>()` → `function: (template_function …)`
- `obj.m<int>()` / `ptr->m<int>()` → `field: (template_method …)`
- `ns::helper()` / `std::move(x)` → `function: (qualified_identifier …)`

The last is **not about templates and is the larger loss**: every namespace-qualified free call in the language was dropped, so a codebase that qualifies consistently recorded almost no free calls at all. `rust.go` already carries the equivalent `scoped_identifier` pattern; C++ now matches it.

## Swift — two losses, the second much larger

Adding type arguments reroutes *any* Swift call away from `call_expression` entirely:

```
constructor_expression
  user_type "obj.doThing<Int>"
    type_identifier "obj"
    type_identifier "doThing"
    type_arguments "<Int>"
  constructor_suffix "()"
```

`lowerGeneric<Int>()`, `obj.doThing<Int>()` and `MyType<Int>(x: 1)` all became `constructor_expression`, which nothing in the query matched. The constructed type's own `type_identifier` children give the dotted path back; the ones nested inside `type_arguments` are the arguments and are excluded.

**The second loss was not in the issue.** The member branch was gated to factory chains only, on the stated belief that the bare-identifier pattern stayed *"authoritative for ordinary `obj.method()`"*. It never was — that pattern needs a `simple_identifier` as a **direct** child of the call, and a member call nests it under a `navigation_expression`. So every plain `obj.method()` and `self.method()` emitted no edge either:

```
  calls Runner.run -> unresolved::free        line=8   ✓ free call
  (nothing)                                   line=9   ✗ h.work()
  (nothing)                                   line=10  ✗ self.free()
  calls Runner.run -> unresolved::*.work      line=11  ✓ Helper().work()  (factory chain)
```

Swift was recording free calls and factory chains and nothing in between. Removing the gate restores what every other language already does, and the generic fix would have been incoherent without it — `o.doThing<Int>()` resolving while `o.doThing()` did not.

## Verified clean, and why

Probed the same way and left alone: **typescript** (.ts + .tsx), **java**, **kotlin**, **rust**, **dart**, **c**, **php**. In TS and Java the type arguments are a *sibling field* of the callee; in Kotlin they live inside `call_suffix`; Dart's sit inside `argument_part`; C and PHP have no generic call syntax.

## Verification

`go test ./internal/...` passes in full; `golangci-lint` reports 0 issues. Each language's test pairs the plain and generic spellings of every call shape and asserts they behave identically — the property that was actually violated, and the one that makes the regression hard to reintroduce quietly.

## Note for merging

Touches the same `extractorVersions` map as #536, so whichever merges second will want a one-line conflict resolution. The two PRs are otherwise independent.

### DIFF
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -34,6 +34,7 @@ var extractorVersions = map[string]int{
 	"php":    2,                                      // class/interface inheritance now emits typed structural edges
 	"csharp": 8,                                      // interface base lists emit extends edges (was: record positional property nodes)
 	"scala":  2,                                      // explicitly instantiated generic calls emit call edges
+	"go":     2,                                      // explicitly instantiated generic calls emit call edges
 }
 
 // extractorSaltExtLang maps a lower-case file extension to the language
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -35,6 +35,7 @@ var extractorVersions = map[string]int{
 	"csharp": 8,                                      // interface base lists emit extends edges (was: record positional property nodes)
 	"scala":  2,                                      // explicitly instantiated generic calls emit call edges
 	"go":     2,                                      // explicitly instantiated generic calls emit call edges
+	"cpp":    2,                                      // templated and namespace-qualified calls emit call edges
 }
 
 // extractorSaltExtLang maps a lower-case file extension to the language
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -33,6 +33,7 @@ var extractorVersions = map[string]int{
 	"c":      generatedParserProjectionPolicyVersion, // generated parser projection covers all strictly detected table sizes
 	"php":    2,                                      // class/interface inheritance now emits typed structural edges
 	"csharp": 8,                                      // interface base lists emit extends edges (was: record positional property nodes)
+	"scala":  2,                                      // explicitly instantiated generic calls emit call edges
 }
 
 // extractorSaltExtLang maps a lower-case file extension to the language
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -36,6 +36,7 @@ var extractorVersions = map[string]int{
 	"scala":  2,                                      // explicitly instantiated generic calls emit call edges
 	"go":     2,                                      // explicitly instantiated generic calls emit call edges
 	"cpp":    2,                                      // templated and namespace-qualified calls emit call edges
+	"swift":  2,                                      // generic calls and ordinary member calls emit call edges
 }
 
 // extractorSaltExtLang maps a lower-case file extension to the language
--- a/internal/indexer/extractor_version_test.go
+++ b/internal/indexer/extractor_version_test.go
@@ -40,12 +40,21 @@ func TestStaleLangsDetection(t *testing.T) {
 			t.Errorf("bad json = %v, want nil", got)
 		}
 		// Against the live extractor versions, an unchanged baseline language
-		// is not stale.
-		if got := ExtractorVersionStaleLangs(`{"go":1}`); len(got) != 0 {
+		// is not stale. Java is the exemplar because its extractor has never
+		// been bumped — a language whose version this suite also asserts
+		// would make the check tautological.
+		if got := ExtractorVersionStaleLangs(`{"java":1}`); len(got) != 0 {
 			t.Errorf("stored at current = %v, want empty", got)
 		}
-		if got := ExtractorVersionStaleLangs(`{"go":1,"php":1}`); !reflect.DeepEqual(got, []string{"php"}) {
+		if got := ExtractorVersionStaleLangs(`{"java":1,"php":1}`); !reflect.DeepEqual(got, []string{"php"}) {
 			t.Errorf("stored PHP structural-edge version = %v, want [php]", got)
+		}
+		// A store extracted before the generic-call fix must re-extract:
+		// until then, every call spelling explicit type arguments is missing
+		// from its graph entirely, and no content change will trigger it.
+		if got := ExtractorVersionStaleLangs(`{"go":1,"scala":1,"cpp":1,"swift":1}`); !reflect.DeepEqual(
+			got, []string{"cpp", "go", "scala", "swift"}) {
+			t.Errorf("stored pre-generic-call version = %v, want all four", got)
 		}
 		// A store extracted before the C# usings-scope/builtin-receiver
 		// stamps must re-extract its .cs files on upgrade — the stamps
--- a/internal/parser/languages/cpp.go
+++ b/internal/parser/languages/cpp.go
@@ -74,6 +74,33 @@ const qCppAll = `
   (call_expression
     function: (field_expression
       field: (field_identifier) @callm.method)) @callm.expr
+
+  ; Explicitly instantiated template calls. foo<T>() wraps its callee in
+  ; a template_function, and obj.foo<T>() / ptr->foo<T>() wrap the member
+  ; in a template_method, so neither pattern above could match and every
+  ; templated call site emitted no edge at all.
+  (call_expression
+    function: (template_function
+      name: (identifier) @call.name)) @call.expr
+
+  (call_expression
+    function: (field_expression
+      field: (template_method
+        name: (field_identifier) @callm.method))) @callm.expr
+
+  ; Namespace-qualified free calls. std::move(x), ns::helper() and their
+  ; templated forms parse the callee as a qualified_identifier, which the
+  ; bare-identifier pattern cannot match — so every :: call was dropped
+  ; too, generics or not. Rust already carries the equivalent
+  ; scoped_identifier pattern; the trailing name is what the call names.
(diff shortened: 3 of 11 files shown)


zzet:
Closing the loop on the one claim in this PR that was an argument rather than a measurement.

Removing the Swift member-call gate was justified by "every other language emits member calls, so Swift was the outlier" — but the gate's own comment existed to stop the graph being *flooded*, and there was no Swift corpus in the repo to check against. There is one off-repo: 78 files of real iOS app code plus vendored CocoaPods sources. Measured at `3c5fd67` (the commit this PR merged onto) and at the merge result:

| corpus | files | call edges | calls/file | member calls | member share |
|---|---:|---:|---:|---:|---:|
| Swift, before | 78 | 1767 | 22.7 | 97 | 5% |
| Swift, after | 78 | 3724 | 47.7 | 2049 | 55% |
| Go `internal/resolver` | 78 | 3978 | 51.0 | 863 | 22% |
| Go `internal/mcp` | 78 | 7955 | 102.0 | 1820 | 23% |

Two things fall out.

**The gate was hiding ~95% of Swift member calls.** Only 97 of 2049 survived it — the factory-chain ones. A 5% member share was never plausible for Cocoa code, where message passing is the dominant idiom; the 55% after the fix is what iOS code actually looks like.

**It is not flooding.** At 47.7 call edges per file Swift now sits *below* both Go corpora (51.0 and 102.0). Node counts are unchanged at 2455 — this recovers edges that were always implied by the source, and lands Swift in the same density band as a language nobody considers noisy.

Method: extract each file with the real `SwiftExtractor` and count `EdgeCalls`, `unresolved::*.` prefix for the member subset. Same file set and same 78-file cap on both sides.

PR #540, opened 2026-08-11
fix(store_sqlite): bound the repo memory estimate and never cache a truncated count

### LINKED ISSUE #539: daemon status exceeds its own 30s budget: AllRepoMemoryEstimates full-scans a 9.4GB store, and the timeout message blames track/reload instead
## Summary

`gortex daemon status` exceeds its own 30 s control budget and reports

```
Error: daemon unresponsive: "status" got no response within 30s (the daemon is up
but not answering — a long track / reload / enrichment may be holding it)
```

**Nothing was holding it.** `status` is itself the slow operation: `handleControl` calls `AllRepoMemoryEstimates`, which does a sequential B-tree walk over a 9.4 GB `store.sqlite`. The message names a cause it never measured, which sent me down the wrong path twice before I profiled it.

## Environment

- `gortex v0.63.2+6a2905e`, Windows 11, daemon pid 1140, **RSS 14.2 GB**
- 5 tracked repos, `~/.gortex/store/store.sqlite` = **9,404,366,848 bytes** (+30 MB WAL)
- Captured live while the daemon was in the failing state, before any restart

## Evidence

**1. It is genuinely busy, not deadlocked.** CPU time sampled 10 s apart on the daemon process:

```
CPU t0 = 5950.02s
CPU t1 = 5986.22s
delta  =   36.20s over 10s wall  ->  ~3.6 cores saturated
```

Zero goroutines were in `semacquire`. This matters because the *same error message* appeared on an earlier build in a genuine lock-hold, where the CPU delta was **0**. One message, two opposite root causes, and it asserts one of them.

**2. The control handler is the thing running.** From `/debug/pprof/goroutine?debug=2`:

```
goroutine 1252 [syscall]:
  modernc.org/sqlite/lib._winRead
  modernc.org/libc.XReadFile
  ...
  github.com/zzet/gortex/internal/graph/store_sqlite.(*Store).AllRepoMemoryEstimates
      internal/graph/store_sqlite/store.go:2223
  github.com/zzet/gortex/internal/daemon.(*Server).handleControl
      internal/daemon/server.go:686
```

**3. A 15 s CPU profile says where the time goes** (`/debug/pprof/profile?seconds=15`, total samples 37.38 s = 249%):

```
    flat  flat%                 cum   cum%
  34.36s 91.92%  runtime.cgocall
   0.21s  0.56%  31.37s 83.92%  modernc.org/sqlite/lib._sqlite3VdbeExec
   0.04s  0.11%  29.81s 79.75%  modernc.org/sqlite/lib._getPageMMap
   0.03s  0.08%  29.41s 78.68%  modernc.org/sqlite/lib._readDbPage
   0.02s 0.054%  26.87s 71.88%  modernc.org/sqlite.(*rows).Next
   0.02s 0.054%  14.38s 38.47%  modernc.org/sqlite/lib._sqlite3BtreeNext
```

~79% of CPU is SQLite page reads, and `_sqlite3BtreeNext` at 38% cumulative says this is a **sequential B-tree walk**, not an index lookup.

## Why this gets worse on its own

If the estimate is a full scan, its cost tracks **store size**, not repo count. So any daemon crosses the 30 s budget eventually just by being used, and tracking more repos only brings the date forward. Mine crossed it at 9.4 GB.

## Two separable asks

1. **Make `status` cheap.** Per-repo byte estimates are a *nice-to-have* column; a status call should not full-scan the store to produce them. Options that would each fix it: read from `dbstat`/precomputed counters, cache the estimate with a TTL, or make the byte columns opt-in behind a flag so plain `status` stays O(1).
2. **Stop the message asserting a cause it did not measure.** "a long track / reload / enrichment may be holding it" is hard-coded. When the timeout fires, the daemon already knows which control kind was in flight — naming that, or simply saying "the `status` handler did not complete within 30 s", would be accurate. As written it actively misdirects: I restarted twice on that advice before profiling.

Happy to test a patch on this corpus — a 9.4 GB store reproduces it every time.

tiendungdev:
Fixes #539.

## What the dump showed

Captured live while `gortex daemon status` was timing out, before any restart:

```
goroutine 1252 [syscall]:
  modernc.org/sqlite/lib._winRead
  store_sqlite.(*Store).AllRepoMemoryEstimates   store.go:2223
  daemon.(*Server).handleControl                 server.go:686
```

`status` was not blocked by a track/reload/enrichment. It **was** the slow work. CPU delta measured 36.2 s over 10 s wall — genuinely busy, zero goroutines in `semacquire`.

## Measured, same driver, read-only, on a 9.4 GB store

```
nodes  ->  6.9s   SCAN nodes USING COVERING INDEX nodes_by_repo
edges  -> 27.8s   SCAN n USING COVERING INDEX nodes_by_repo
                  SEARCH e USING COVERING INDEX edges_by_from_line (from_id=?)
          -----
           34.7s  > the 30s control budget
```

**No index is missing** — both plans are already covering. The cost is that `COUNT(*) … GROUP BY` must touch every node and every edge.

And the existing mitigation could not help: `memEstTTL` is **3 s in front of a ~35 s scan**, so the cache expired about ten times over before it was ever read. Every poll recomputed, holding `memEstMu` throughout.

## Three changes, no schema

1. **Bound it.** Both scans run under a `memEstBudget` (2 s) context, so an advisory display column can no longer spend the caller's whole budget. I verified the driver honours the deadline mid-scan before relying on it.
2. **Never cache a truncated count.** The scans now report completion. A partial count is not *stale*, it is *wrong*, and the TTL would serve it as if measured. The old loop never checked `rows.Err()`, so a cut-short iteration fell straight through to the cache write — adding a deadline without this would have promoted a latent bug to a routine one.
3. **Scale the TTL with measured cost** (`T*10`, floored at the old 3 s, capped at 5 min), so memoisation keeps working as the store grows instead of silently degrading.

Backoff after a truncated recompute applies **only when a cached answer already exists**. With nothing cached there is nothing to protect, so it keeps trying — each attempt is bounded anyway.

## Conventions followed, not invented

`QueryContext` and the `rows.Err() … ctx.Err() == nil` pattern are already used elsewhere in this file (1784, 1826, 1851). `AllRepoMemoryEstimates` was the one place doing neither.

## Test, and an honest note on what it does not cover

`TestAllRepoMemoryEstimates_TruncatedScanIsNotCached` builds a 20k-node fixture, shrinks the budget through an `export_test` hook so the scan is cut mid-flight, then asserts the next call reports the true count.

Sabotage-checked:

| Sabotage | Result |
|---|---|
| remove the `complete` guard | **red**, as intended |
| remove `rows.Err()` on the *nodes* query | **green** |

The second stays green because the expired context makes the following edges query fail at `QueryContext`, so truncation is caught downstream. That per-query check is convention-conformance for real I/O errors, **not** something this test covers — saying so rather than letting the table imply otherwise.

The test also caught a flaw in my own first draft: it backed off unconditionally and returned an empty map where the true count was available.

## Test run

`go test ./internal/graph/store_sqlite/` has five pre-existing failures on this tree (`TestSweepPlanLocks`, `TestAdjacencyPlanLocksDuringBulkLoad`, `TestSweepPlanLockReceiverRebindBatch`, `TestPreparedStatementPlansNeverScanBigTables`, `TestSweepWarnPlanLocks`). I verified they fail identically on unmodified `main` by stashing the change and re-running, so they are not from this PR.

## Not addressed here

Denormalising `repo_prefix` onto `edges` so the second query becomes a single-table `GROUP BY`. That removes the join entirely and is the real long-term fix — but it is a schema change plus migration, and that is your call.

### DIFF
--- a/internal/graph/store_sqlite/export_memest_test.go
+++ b/internal/graph/store_sqlite/export_memest_test.go
@@ -1,0 +1,13 @@
+package store_sqlite
+
+import "time"
+
+// SetMemEstBudgetForTest shrinks the per-recompute budget that
+// AllRepoMemoryEstimates gives its COUNT scans, so a test can force the
+// truncated path deterministically instead of waiting for a slow store.
+// It returns a restore func.
+func SetMemEstBudgetForTest(d time.Duration) func() {
+	prev := memEstBudget
+	memEstBudget = d
+	return func() { memEstBudget = prev }
+}
--- a/internal/graph/store_sqlite/store.go
+++ b/internal/graph/store_sqlite/store.go
@@ -144,6 +144,12 @@ type Store struct {
 	memEstMu  sync.Mutex
 	memEstVal map[string]graph.RepoMemoryEstimate
 	memEstAt  time.Time
+	// memEstCost is what the last *complete* recompute actually took; the
+	// TTL scales with it so a scan that costs seconds is not repeated every
+	// few seconds. memEstRetryAt suppresses retries after a recompute ran
+	// out of budget.
+	memEstCost    time.Duration
+	memEstRetryAt time.Time
 
 	// Bulk-load fast path (graph.BulkLoader). Non-nil only between
 	// BeginBulkLoad and FlushBulk, and only on a first/empty cold index.
@@ -2212,6 +2218,24 @@ func (s *Store) RepoMemoryEstimate(repoPrefix string) graph.RepoMemoryEstimate {
 // status polls from each triggering a full COUNT … GROUP BY scan.
 const memEstTTL = 3 * time.Second
 
+// memEstTTLScale keeps a recompute that cost T from being repeated for
+// roughly T*scale, capped by memEstMaxTTL. A fixed TTL is only correct while
+// the scan is cheap: on a large store the COUNT scan grew past the TTL, so
+// every poll recomputed and the memoisation stopped doing anything.
+const memEstTTLScale = 10
+
+// memEstMaxTTL caps the adaptive TTL so the estimate cannot go stale forever.
+const memEstMaxTTL = 5 * time.Minute
+
+// memEstBackoff suppresses recomputes after one ran out of budget, so a burst
+// of status polls does not walk into the same wall every time.
+const memEstBackoff = 30 * time.Second
+
+// memEstBudget bounds a single recompute. The estimate is advisory (status
+// display), so it must never be able to spend the caller's whole control
+// budget. A var, not a const, so tests can force the truncated path.
+var memEstBudget = 2 * time.Second
+
 func cloneRepoMemEstimates(m map[string]graph.RepoMemoryEstimate) map[string]graph.RepoMemoryEstimate {
 	out := make(map[string]graph.RepoMemoryEstimate, len(m))
 	for k, v := range m {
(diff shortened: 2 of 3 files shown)


tiendungdev:
CI is red on `test (macos-latest, 1.26)`. I do not think it is this PR, but I want to lay out what I checked and — more usefully — what I could **not** check, because the failure looks like a real race rather than infrastructure noise.

## The failure

```
WARNING: DATA RACE
--- FAIL: TestReconcileContractEdges_ConcurrentNoRaceOrTear (1.81s)

Write  g874: resolver.(*Resolver).ResolveAllContext      resolver/resolver.go:994
             <- indexer.indexCtxRaw <- MultiIndexer.TrackRepoCtx

Previous read  g885: graph.(*Graph).EvictEdgesByKinds    graph/graph.go:1415
             <- indexer.(*MultiIndexer).ReconcileContractEdges   multi.go:4075
```

Both goroutines are the test's own: `func1` flips the matched set via `UntrackRepo` + `TrackRepoCtx`, while 12 readers call `ReconcileContractEdges`.

## Why I do not think this PR caused it

- The diff is confined to `internal/graph/store_sqlite/` — one source file plus two new tests in that package.
- **The failing test does not use `store_sqlite` at all** (`grep store_sqlite contract_bridge_test.go` → 0). It runs against the in-memory `Graph`.
- No frame in the race report belongs to `store_sqlite`; the two sides are `internal/resolver` and `internal/graph`.
- `internal/indexer` never calls `AllRepoMemoryEstimates`, the only function this PR changes the behaviour of.
- `test (ubuntu-latest, 1.26)` passed on the same commit, as did `build-windows`, `lint`, `benchmark` and `govulncheck`.
- Locally, `go test -race ./internal/indexer/ -run TestReconcileContractEdges_ConcurrentNoRaceOrTear -count=3` passes on Windows/amd64 with this branch applied.

## What I could not check, stated plainly

I could not reproduce the failure at all on my machine, so I have **not** shown that unmodified `main` races the same way on macOS. My evidence is exclusion — the diff does not touch that path — not a matched control run. I have no macOS runner, and I cannot rerun your jobs.

So please read the above as "this PR is very unlikely to be the cause", not as "proven unrelated".

## The part I would rather not leave implicit

I deliberately did not push an empty commit to re-roll CI. If it went green I would have reported the problem as solved, when all I would have done is win a coin flip.

The race detector does not report races it did not observe. `ReconcileContractEdges` running concurrently with `TrackRepoCtx` appears to have a genuine race on shared `Graph`/resolver state, and 12 writers against 96 reader iterations is exactly the shape that surfaces on one scheduler (macOS/arm64) and hides on another (ubuntu/amd64). If a rerun goes green, that is the race hiding again rather than the bug leaving.

Happy to open a separate issue for it with this stack if you would like it tracked apart from this PR — it is not mine to fix blind, since I cannot reproduce it.

PR #550, opened 2026-08-12
fix(mcp): fail closed on invalid change.detect source fields

### LINKED ISSUE #549: change.detect silently ignores source.repo_path and analyzes the active repository
## Environment

- Gortex: `v0.63.2+6a2905e`
- Surface: MCP facade v1
- Active repository: `domain-manage`
- Another tracked repository: `gortex-pr544`

## Summary

`change.detect` silently accepts an unsupported `source.repo_path`, drops it during dispatch, and analyzes the active repository instead. The response is structurally successful and even echoes the wrong effective repo, so automation can produce a false blast-radius result for an unrelated working tree.

The documented/supported selector `options.repo` works. The bug is that the runtime does not enforce its own facade schema and silently falls back when the unsupported field is supplied.

## Reproduction

With `domain-manage` active and a separate tracked Gortex clone at:

`C:\Users\daoti\AppData\Local\Temp\gortex-pr-d49ee32ae27b44cd94d1ab13c85ac7e0\gortex`

call:

```json
{
  "operation": "detect",
  "source": {
    "repo_path": "C:\\Users\\daoti\\AppData\\Local\\Temp\\gortex-pr-d49ee32ae27b44cd94d1ab13c85ac7e0\\gortex"
  },
  "output": {
    "detail": "full"
  }
}
```

Observed response:

```json
{
  "changed_files": [
    "tests\\scripts\\test_agent_harness_docs.py",
    "AGENTS.md"
  ],
  "repo": "domain-manage",
  "repo_root": "C:\\Users\\daoti\\Downloads\\domain-manage",
  "scope": "unstaged"
}
```

Those files belonged to the active `domain-manage` tree, not the requested Gortex clone.

The supported form resolves the intended tracked repo correctly:

```json
{
  "operation": "detect",
  "options": {
    "repo": "C:\\Users\\daoti\\AppData\\Local\\Temp\\gortex-pr-d49ee32ae27b44cd94d1ab13c85ac7e0\\gortex"
  },
  "source": {
    "scope": "unstaged"
  }
}
```

It returns `repo: "gortex-pr544"` and the correct `repo_root`.

## Schema and code-path evidence

`capabilities(domain:"change", operation:"detect", detail:"schema")` reports:

- `source.additionalProperties: false`, with only `scope` allowed.
- `options.additionalProperties: false`, with `repo` as the repository selector.

However:

- `validateFacadeInput` checks only that `source` is an object; it does not reject unknown nested properties.
- `normalizeFacadeArguments` merges every key from `source` into the legacy argument map.
- `handleDetectChanges` reads only `repo`, never `repo_path`.
- `resolveDiffRoot` therefore receives an empty selector and falls back to the session's active/cwd-bound repository.

This creates a gap between the published JSON schema and live runtime validation.

## Expected behavior

Either of these is safe:

1. Reject `source.repo_path` with a structured `invalid_argument` that points to `options.repo`; or
2. Deliberately support it as an alias and normalize it to `repo`.

Silently ignoring a repository selector must never produce a successful analysis of another repo.

## Acceptance criteria

- [ ] Runtime validation enforces nested `additionalProperties:false` for facade containers, at least for safety-sensitive selectors.
- [ ] Unsupported `source.repo_path` returns a structured error naming the field and the supported `options.repo` form, or is explicitly normalized as a documented alias.
- [ ] `change.detect` never falls back to the active repo when any repository-selector-like field was supplied but not consumed.
- [ ] Tests cover active repo A + requested tracked repo B, an untracked path, an unknown nested source field, and the supported `options.repo` path/prefix forms.
- [ ] Audit other diff-backed facades such as `review.run` for the same silent-container-field drop.

tiendungdev:
## Summary

- fail closed when `change.detect` receives unknown fields in its `source` container instead of silently forwarding and dropping them
- return a structured `invalid_argument` for `source.repo_path` with an explicit `options.repo` migration hint
- preserve the supported `source.scope` plus `options.repo` / `options.base_ref` lowering to the legacy change detector
- add focused regression coverage for the wrong-active-repository fallback and generic unknown source fields

## Scope

This is intentionally limited to `change.detect`, where an ignored repository selector can produce a successful analysis of the wrong working tree. It does not introduce a facade-wide nested-schema validator or change other operations' compatibility aliases.

## Verification

- `go test ./internal/mcp -run '^TestChangeDetectSource(FailClosed|Allows)' -count=1`
- `go test -race ./internal/mcp -run '^TestChangeDetectSource(FailClosed|Allows)' -count=1`
- `go vet ./internal/mcp`
- `go build ./cmd/gortex`
- `gofmt -d internal/mcp/facade_change_detect_source_test.go`
- `git show --check HEAD`

Fixes #549

### DIFF
--- a/internal/mcp/facade_change_detect_source_test.go
+++ b/internal/mcp/facade_change_detect_source_test.go
@@ -1,0 +1,80 @@
+package mcp
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+func TestChangeDetectSourceFailClosedBeforeActiveRepoFallback(t *testing.T) {
+	spec := facadeOperationSpec{Facade: "change", Operation: "detect", Legacy: "detect_changes"}
+	result := validateFacadeInput(spec, map[string]any{
+		"source": map[string]any{
+			"scope":     "unstaged",
+			"repo_path": `C:\work\repo-b`,
+		},
+	})
+	if result == nil || !result.IsError {
+		t.Fatalf("validateFacadeInput() = %#v, want structured tool error", result)
+	}
+	payload, err := json.Marshal(result)
+	if err != nil {
+		t.Fatalf("marshal result: %v", err)
+	}
+	for _, want := range []string{
+		`\"error_code\":\"invalid_argument\"`,
+		`source.repo_path`,
+		`unsupported_field`,
+		`options.repo`,
+	} {
+		if !strings.Contains(string(payload), want) {
+			t.Errorf("error payload %s does not contain %q", payload, want)
+		}
+	}
+}
+
+func TestChangeDetectSourceFailClosedForUnknownField(t *testing.T) {
+	spec := facadeOperationSpec{Facade: "change", Operation: "detect", Legacy: "detect_changes"}
+	result := validateFacadeInput(spec, map[string]any{
+		"source": map[string]any{"paths": []string{"README.md"}},
+	})
+	if result == nil || !result.IsError {
+		t.Fatalf("validateFacadeInput() = %#v, want structured tool error", result)
+	}
+	payload, err := json.Marshal(result)
+	if err != nil {
+		t.Fatalf("marshal result: %v", err)
+	}
+	for _, want := range []string{`source.paths`, `\"allowed_fields\":[\"scope\"]`} {
+		if !strings.Contains(string(payload), want) {
+			t.Errorf("error payload %s does not contain %q", payload, want)
+		}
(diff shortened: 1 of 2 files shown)


tiendungdev:
Superseded by #553, which preserves the same fail-closed change.detect behavior while also auditing review.run, covering tracked path/prefix and absolute untracked-path resolution, and retaining existing facade compatibility aliases. All checks on #553 are green.

Issue #545, opened 2026-08-12
Expose disk-verified SHA-256 and provenance in file reads and mutation receipts


tiendungdev:
**Is your feature request related to a problem?**

Yes. Gortex mutation responses may expose fields such as `new_sha`, and `read(operation:"file")` supports an ETag/`if_none_match` workflow, but the current public MCP contract does not expose a clearly defined, disk-verified SHA-256 evidence mode.

Verified against `v0.63.2+6a2905e`, the `read(operation:"file")` schema provides content selection, compression, secret handling, offsets, limits, and `if_none_match`, but no documented option that guarantees:

- SHA-256 was computed from the complete physical file on disk;
- the digest covers the same byte buffer that was read or returned;
- whether a hash came from disk, an overlay, an index snapshot, or proposed content;
- whether the response content was truncated while the hash still covered the full file;
- the resolved path and file type used for the calculation;
- whether the graph had re-indexed the resulting content.

For evidence-sensitive work, an agent therefore has to ignore `new_sha` as authoritative physical evidence and run a separate OS command such as `Get-FileHash` or `sha256sum`, followed by readback. This is slower, platform-specific, and creates a time-of-check/time-of-use gap.

**Describe the solution you'd like**

Add an explicit physical-evidence mode to file reads:

```json
{
  "operation": "file",
  "target": {
    "file": "tmp/proposal/manifest.txt"
  },
  "options": {
    "physical_evidence": true,
    "digest": "sha256"
  }
}
```

Suggested response fields:

```json
{
  "resolved_path": "...",
  "file_kind": "regular",
  "git_state": "ignored",
  "byte_count": 1234,
  "content_sha256": "...",
  "hash_algorithm": "sha256",
  "hash_scope": "full_file",
  "content_source": "disk",
  "disk_verified": true,
  "content_truncated": false,
  "same_buffer_as_content": true,
  "read_at": "...",
  "index_state": "current"
}
```

When response content is bounded or compressed, the contract should distinguish the returned representation from the digest scope, for example:

```json
{
  "content_truncated": true,
  "hash_scope": "full_file",
  "same_buffer_as_content": false
}
```

Mutation operations should return the same evidence vocabulary:

```json
{
  "before_sha256": "...",
  "after_sha256": "...",
  "content_source": "disk",
  "disk_verified": true,
  "graph_reindexed": true,
  "indexed_sha256": "..."
}
```

[...]

PR #554, opened 2026-08-12
feat(mcp): add physical file read evidence

tiendungdev:
## Summary

- add opt-in `physical_evidence` and `digest` parameters to `read_file` and the `read.file` facade schema
- compute SHA-256 from a stable full-file disk buffer, independently of line windows, compression, redaction, response caps, binary encoding, or editor overlays
- return explicit provenance (`resolved_path`, `byte_count`, `content_sha256`, hash scope/source, content source, symlink state, observation time, and same-buffer status)
- keep conditional-read ETags stable by excluding the observation timestamp while including the verified evidence fields

## Scope

This is the read-side delivery slice of #545. Mutation receipts remain a separate follow-up, so this PR references rather than closes the issue.

Refs #545

## Verification

- `go test ./internal/mcp -run '^TestReadFilePhysicalEvidence' -count=1`
- `go test -race ./internal/mcp -run '^TestReadFilePhysicalEvidence' -count=1`
- focused facade/window regression tests
- `go vet ./internal/mcp`
- `go build ./cmd/gortex`
- `git diff --cached --check`

`go test ./internal/mcp -count=1` was also attempted on Windows but exceeded the 5-minute local timeout without emitting a test failure. The focused, race, vet, and build checks above completed successfully.


zzet (changes requested):
CI is red, and it reproduces locally.

TestReadFilePhysicalEvidenceHashesFullDiskBuffer fails on macOS:

  expected: "/var/folders/.../001/evidence.txt"
  actual  : "/private/var/folders/.../001/evidence.txt"

The test asserts resolved_path == filepath.Join(dir, ...) while the implementation returns the symlink-resolved path. The Ubuntu job shows fail in gh pr checks, but its Test step was cancelled by matrix fail-fast — macOS is the only genuine failure. Everything else is clean: only that one test fails in the whole internal/mcp package, and golangci-lint is green.

The one-line test fix (EvalSymlinks the expectation) is trivial, but it surfaces a real design bug:

- symlink_resolved doesn't mean what it says. It's EvalSymlinks(abs) != abs, so it's true whenever any ancestor component is a link — every file under /tmp or /var on macOS, any repo under a symlinked home. The issue asked to "report link vs target"; this reports "something in the path was a link", which a caller can't use to decide whether the file itself is a symlink. os.Lstat(absPath).Mode()&os.ModeSymlink is the fact being asked for.

The most subtle new code has zero coverage. samePhysicalFileVersion plus the double-read / re-stat / re-EvalSymlinks drift detector is the safety argument for the feature, and nothing exercises it. Issue #545 explicitly lists "same-size content drift, symlinks/junctions, and files changing during observation" as acceptance criteria. The six tests cover happy path, window, binary, empty, ETag stability, arg validation, and facade publication — all the easy parts.

It reads the file twice, against an explicit constraint in the issue ("Hash from one stable buffer/file handle where possible so the bytes are not read twice"). readPhysicalFileEvidence does ReadAll → seek → ReadAll and holds both buffers: 2× peak memory and 2× I/O per evidence read, uncapped. The stat-triple already covers the ordinary case, and the byte-compare only closes a window that reopens the moment the function returns.

content_truncated gets its meaning overwritten. In the base path it means "capped at max_chars", paired with a max_chars field and a matching omission note. Under physical_evidence it's reassigned to "returned bytes differ from hashed bytes in any way" — windowing, binary, redaction, and elision all set it, with no max_chars companion (the PR's own binary test asserts this). same_buffer_as_content already carries that meaning; the existing field shouldn't be repurposed.

tiendungdev:
Addressed the requested redesign in f8eaa393:

- physical evidence now reads and hashes one buffer from one open handle (no rewind/second ReadAll, so no 2x I/O or peak buffer);
- metadata, file identity, and resolved-path checks remain around that read and reject same-size drift or path replacement;
- symlink_resolved now reports whether the requested file itself is a symlink via Lstat, rather than whether any ancestor resolves differently;
- content_truncated is again reserved for max_chars truncation; same_buffer_as_content remains the divergence signal;
- the macOS expectation now compares against EvalSymlinks;
- added coverage for max_chars semantics, same-size drift, path replacement, direct symlink, and ancestor symlink behavior.

Local verification:
- go test ./internal/mcp -run TestReadFilePhysicalEvidence -count=1
- go test -race ./internal/mcp -run TestReadFilePhysicalEvidence -count=1
- go vet ./internal/mcp
- go build ./internal/mcp
- git diff --check

All passed.

internal/mcp/tools_fileops.go
```diff
@@ -1132,6 +1212,35 @@ func (s *Server) handleReadFile(ctx context.Context, req mcp.CallToolRequest) (*
...
+	var diskContent []byte
+	var physicalEvidence physicalReadEvidence
+	if physicalEvidenceRequested {
+		var readErr error
+		diskContent, physicalEvidence, readErr = readPhysicalFileEvidence(absPath)
+		if readErr != nil {
+			return mcp.NewToolResultError(readErr.Error()), nil
+		}
+		// Close the symlink-retargeting window around the stable physical read.
+		if guardErr := s.guardSymlinkWithinRepo(absPath); guardErr != nil {
```
zzet, line 1276: **High — bind confinement to the target actually hashed.** This post-read guard re-resolves absPath, not physicalEvidence.resolvedPath. I reproduced the pre-guard seeing an inside target, the helper reading and hashing a stable outside target, then the post-guard succeeding after the link was restored inside. Validate the observed resolved target against the resolved repository roots before returning evidence, ideally using handle-relative/no-follow primitives, and add an out-and-back retarget regression.

internal/mcp/tools_fileops.go
```diff
@@ -1113,6 +1116,83 @@ func capReadFileContent(content []byte, maxChars int, binary bool) ([]byte, bool
...
+	resolvedPath    string
+	contentSHA256   string
+	byteCount       int
+	symlinkResolved bool
+	readAt          time.Time
+}
+
+func samePhysicalFileVersion(a, b os.FileInfo) bool {
+	return a != nil && b != nil && a.Mode().IsRegular() && b.Mode().IsRegular() &&
+		os.SameFile(a, b) && a.Size() == b.Size() && a.ModTime().Equal(b.ModTime())
```
zzet, line 1141: **High — this version check can certify stale bytes.** Rewriting before to same-length after! in the existing afterRead hook and restoring the original mtime makes inode, size, and mtime all compare equal, so the helper succeeds and reports disk_verified=true although the path now contains different bytes. Use a stronger change signal where available or a verification strategy that cannot be bypassed by preserved/coarse timestamps, and add the restored-mtime regression.

internal/mcp/tools_fileops.go
```diff
@@ -1113,6 +1116,83 @@ func capReadFileContent(content []byte, maxChars int, binary bool) ([]byte, bool
...
+func readPhysicalFileEvidenceObserved(absPath string, afterRead func()) ([]byte, physicalReadEvidence, error) {
+	linkInfo, err := os.Lstat(absPath)
+	if err != nil {
+		return nil, physicalReadEvidence{}, fmt.Errorf("could not inspect physical path: %w", err)
+	}
+	resolvedBefore, err := filepath.EvalSymlinks(absPath)
+	if err != nil {
+		return nil, physicalReadEvidence{}, fmt.Errorf("could not resolve physical file: %w", err)
+	}
+	f, err := os.Open(absPath)
```
zzet: **Fail closed before opening special files.** os.Open on a FIFO blocks waiting for a writer; the regular-file check occurs only afterward. My bounded regression hung until it supplied a writer, then received the intended error. Reject non-regular targets before a blocking open and use a race-safe/nonblocking open plus fstat so FIFOs, devices, and sockets cannot stall or cause side effects.

internal/mcp/tools_fileops.go
```diff
@@ -1256,6 +1367,24 @@ func (s *Server) handleReadFile(ctx context.Context, req mcp.CallToolRequest) (*
...
+	if physicalEvidenceRequested {
+		contentAltered := servedFromOverlay || isBinary || bodiesElided || salienceTruncated || windowed || secretsRedacted || contentTruncated
+		contentSource := "disk"
+		if servedFromOverlay {
+			contentSource = "overlay"
+		}
+		result["resolved_path"] = physicalEvidence.resolvedPath
+		result["file_kind"] = "regular"
+		result["byte_count"] = physicalEvidence.byteCount
+		result["content_sha256"] = physicalEvidence.contentSHA256
```
zzet, line 1420: **Do not bypass the secret-read boundary with a raw digest.** I verified that a .env password is replaced with the redaction marker while this field still returns the SHA-256 of the unredacted disk bytes and folds it into the ETag. That is an offline guess-verification oracle for low-entropy secrets. Require explicit allow_secrets intent for the physical digest, or omit/refuse raw-derived evidence whenever redaction fires.

internal/mcp/tools_fileops.go
```diff
@@ -1256,6 +1367,24 @@ func (s *Server) handleReadFile(ctx context.Context, req mcp.CallToolRequest) (*
...
+		result["hash_algorithm"] = "sha256"
+		result["hash_scope"] = "full_file"
+		result["hash_source"] = "disk"
+		result["content_source"] = contentSource
+		result["disk_verified"] = true
+		result["same_buffer_as_content"] = !contentAltered
```
zzet, line 1426: **same_buffer_as_content must reflect wire bytes.** For a file containing invalid UTF-8 bytes ff fe and no NUL, looksBinary is false and this reports true, but JSON serialization replaces those bytes. Base this claim on actual lossless representability/byte equality, or encode non-text content losslessly, and add an invalid-UTF-8-without-NUL regression.

zzet (changes requested):
Request changes. The revision resolves the earlier one-buffer, truncation, macOS path, and direct-symlink concerns; all GitHub checks are green, the full internal/mcp package and focused race tests pass, and the branch merges cleanly with current main. However, adversarial tests on this exact head reproduce two failures in the core physical-evidence guarantee: repository confinement can be bypassed by retargeting a symlink around the read, and same-size content drift is accepted when mtime is restored. A repository-local FIFO also blocks before rejection, redacted secret content still exposes its raw digest, and invalid UTF-8 is incorrectly labeled as the same returned buffer. These must fail closed before merge. Keep #545 open; mutation receipts, Windows junction coverage, and contract documentation remain follow-up scope.

tiendungdev:
Implemented the requested hardening in cf4429e2.

Changes mapped to the review blockers:
- Confinement now validates the already-observed resolved target, so an out-and-back symlink retarget cannot make the post-read check approve bytes read outside the repo.
- Physical evidence performs a second streaming SHA-256 pass on the same open handle and rejects digest drift, including same-size in-place changes with the original mtime restored.
- POSIX reads use a nonblocking open and fstat regular-file check; the Linux FIFO regression proves the call returns instead of blocking.
- Redacted content now refuses physical evidence unless allow_secrets=true, preventing raw digest/ETag disclosure as a secret oracle.
- Invalid UTF-8 is classified as altered wire content, so same_buffer_as_content cannot be true after JSON replacement.

Verification:
- go test ./internal/mcp -run TestReadFilePhysicalEvidence -count=1 -timeout=180s
- go test -race ./internal/mcp -run TestReadFilePhysicalEvidence -count=1 -timeout=300s
- go vet ./internal/mcp
- git diff --cached --check
- Linux container: the focused FIFO, out-and-back symlink, restored-mtime drift, secret-intent, and invalid-UTF-8 tests all passed.

I also ran the full internal/mcp package on Windows. It reached unrelated existing Windows/path/index fixture failures and a package timeout; none of the physical-evidence tests failed. Mutation receipts, Windows junction coverage, and contract documentation remain follow-up scope as requested.

PR #558, opened 2026-08-12
fix(mcp): publish directly callable request shapes

### LINKED ISSUE #542: bug: capabilities `request_shape` advertises a top-level `arguments` envelope that the facade then hard-rejects — self-refuting error traps agents in a retry loop
## Summary

`capabilities(detail:"schema")` advertises a `request_shape` whose top-level key is
`arguments`. An agent that reads that literally builds a call with a top-level
`arguments` key — and the facade then rejects it with

```json
{"error_code":"invalid_argument","message":"arguments must be an object","data":{"field":"arguments"}}
```

even when the caller did supply an object. Two things turn a small documentation
ambiguity into an unrecoverable loop:

1. `arguments` is the **only** unknown top-level key that is fatal. Every other
   unknown key is silently ignored.
2. The error message asserts the exact thing the agent believes it already did
   ("must be an object"), so the natural repair — "pass an object" — is already
   satisfied and the agent retries the same shape forever.

In the session that prompted this report an agent burned ~8 retries across 4
different gortex tools, restarted the daemon, and reconnected the MCP server,
concluding the whole gortex integration was broken. It was not: gortex was
healthy the entire time and correctly-shaped calls succeeded immediately.

## Environment

- gortex `v0.61.4+447e66c` (built 2026-07-25, go1.26.5, darwin/arm64), installed via the Homebrew cask
- surface: `facade-v1`
- MCP client: Claude Code 2.1.227
- daemon healthy: `state ready`, index warm

## Reproduction

Any indexed repo. Three calls to the same tool:

```jsonc
// 1. PASS — params at top level
search(operation: "symbols", query: "MyType")

// 2. PASS — free-form object params serialize fine; limit and format both take effect
search(operation: "symbols", query: "MyType",
       options: {"limit": 2}, output: {"format": "gcx"})

// 3. FAIL — the shape capabilities advertises
search(operation: "symbols", arguments: {"query": "MyType"})
// => {"error_code":"invalid_argument","message":"arguments must be an object",
//     "data":{"field":"arguments"}}
```

Where call 3's shape comes from — `capabilities(domain:"search", operation:"symbols", detail:"schema")`:

```json
"request_shape": {
  "arguments": {"operation": "symbols", "options": {}, "query": "<query>"},
  "tool": "search"
}
```

This is not specific to `search`. `capabilities(domain:"read", operation:"file")`
returns the same envelope:

```json
"request_shape": {"arguments": {"operation":"file","options":{},"target":{"file":"<file>"}}, "tool":"read"}
```

## Evidence matrix

| call | top-level key | value sent | result |
| --- | --- | --- | --- |
| `search(operation, query)` | — | — | **PASS** |
| `search(..., options:{"limit":2})` | `options` (declared, typed object) | object | **PASS**, limit applied |
| `search(..., output:{"format":"gcx"})` | `output` (declared, typed object) | object | **PASS**, format applied |
| `search(..., zzz_undeclared_param:{"a":1})` | undeclared, arbitrary name | string | **PASS** — silently ignored |
| `search(..., arguments:{"query":"MyType"})` | undeclared, name `arguments` | string | **FAIL** |
| `search(..., arguments:"hello")` | undeclared, name `arguments` | string | **FAIL**, identical error |

Two things fall out of the last three rows:

- The failure is **not** "unknown key rejected" — `zzz_undeclared_param` is
  ignored and the call succeeds. Only the name `arguments` is fatal.
- The failure is **not** "free-form object params are broken" — `options` and
  `output` are free-form in the public tool schema
  (`{"additionalProperties":true,"properties":{}}`) and both round-trip
  correctly as objects.

The failing request does not appear in `daemon.log`, so the rejection looks like
it happens in the facade's argument validation ahead of the daemon.

## Why the value arrives as a string

Because `arguments` is absent from the tool's `inputSchema`, the client has no
type for it and the value reaches the server as a JSON **string** rather than an
object — which is why the server's "must be an object" check fires on input the
agent authored as an object.

That stringification of untyped/undeclared parameters is a known client-side
behaviour, tracked upstream at
[anthropics/claude-code#82652](https://github.com/anthropics/claude-code/issues/82652)
(with prior art in #63253 / #77529 / #72248). I am **not** asking gortex to fix
that — I am reporting that gortex's own advertised `request_shape` steers agents
directly into it, and that the facade's asymmetric handling plus the wording of
the error make the resulting failure unrecoverable from inside an agent loop.

## Expected vs actual

**Expected** — one of:
- the advertised `request_shape` is unambiguous about being the JSON-RPC wire
  frame rather than the caller's parameter set; and/or
- a top-level `arguments` key is tolerated (unwrapped, or ignored like every
  other unknown key) rather than fatal.

**Actual** — the advertised shape produces a hard error whose message contradicts
what the caller supplied, and which no amount of retrying can fix.

## Suggested fix direction

Any one of these breaks the loop; the first two are cheap:

1. **Make the error self-correcting.** State the received type and the accepted
   shape, e.g.
   `arguments: unexpected top-level key (received string). Pass operation/query/options at the top level; "arguments" is the JSON-RPC envelope, not a parameter.`
   Even alone, this ends the retry loop.
2. **Rename or qualify the field in `capabilities`.** `request_shape.arguments`
   is accurate as wire documentation but ambiguous to an MCP-client caller who
   only supplies the inner object. Emitting the inner object as e.g.
   `request_shape.params` — or adding a sibling `call_example` showing the
   caller-level shape — removes the trap at the source.
3. **Tolerate the envelope.** If a top-level `arguments` is present and parses as
   an object (or as a JSON string containing one), merge it into the params and
   continue. This makes the advertised shape actually work.
4. **Be consistent about unknown keys.** Whatever the policy, `arguments` should
   not be the sole name that is fatal while all others are ignored.

Happy to test a patch against a real workload.

tiendungdev:
## Summary

- return `request_shape` as the direct caller-level facade input instead of wrapping it in `tool` and `arguments`
- keep legitimate operation-level `arguments` containers, such as `session.cursor`, intact
- return self-correcting diagnostics when an undeclared top-level `arguments` value is sent
- align public capability-schema validation, documentation, and existing request-shape consumers with the direct shape

## Verification

- `go test ./internal/mcp -run 'Facade|Capability' -count=1`
- `go test -race ./internal/mcp -run '^(TestFacadeCapabilityRequestShapeIsDirectCallerInput|TestFacadeInvalidArgumentsEnvelopeErrorIsSelfCorrecting|TestFacadeCapabilitiesPublicSchemasMatchEveryAvailableOperation)$' -count=1`
- `go vet ./internal/mcp`
- `go build ./cmd/gortex`
- `git diff --check`

A full `go test ./internal/mcp -count=1` was also run on Windows. The capability-related failure it exposed was fixed; the remaining failures are unrelated platform/index-fixture failures already outside this diff (Windows file URI/path normalization, symlink privilege, file-mode preservation, and several index fixtures).

Fixes #542

### DIFF
--- a/docs/mcp-facade-v1.md
+++ b/docs/mcp-facade-v1.md
@@ -207,7 +207,7 @@ Each facade advertises only its high-frequency stable fields. A read request ill
 
 The dispatcher flattens the advertised `arguments`, `options`, `source`, `context`, `guard`, and `output` objects into the selected legacy handler's arguments. Common top-level fields win where the adapter defines a friendly alias. Operation-specific fields that are not worth mounting in every model turn remain discoverable through `capabilities`.
 
-`operation` is a normalized string rather than an ever-growing schema enum. It MAY be omitted when a selector uniquely determines the common read/edit default. The public description lists common operations, while `capabilities(domain=<tool>, operation=<operation>, detail="schema")` returns both the exact operation-specific public input schema and a `request_shape`. The dispatcher MUST reject unknown values and return the valid operation names. Capability lists use the public keys `domains` and `domain`; compatibility terminology is not exposed to callers.
+`operation` is a normalized string rather than an ever-growing schema enum. It MAY be omitted when a selector uniquely determines the common read/edit default. The public description lists common operations, while `capabilities(domain=<tool>, operation=<operation>, detail="schema")` returns both the exact operation-specific public input schema and a `request_shape`. The `request_shape` object is the caller-level argument object: pass its fields directly to the named facade tool, without nesting it under an `arguments` or `params` key. The dispatcher MUST reject unknown values and return the valid operation names. Capability lists use the public keys `domains` and `domain`; compatibility terminology is not exposed to callers.
 
 ### 8.2 Target references
 
--- a/internal/mcp/facade_change_targets_test.go
+++ b/internal/mcp/facade_change_targets_test.go
@@ -141,9 +141,8 @@ func TestFacadeChangeTargetShapesAreAdvertisedOnce(t *testing.T) {
 				require.NotContains(t, optionProperties, "ids", "compatibility aliases must not duplicate the canonical target schema")
 			}
 
-			shape := payload["request_shape"].(map[string]any)
-			arguments := shape["arguments"].(map[string]any)
-			require.Contains(t, arguments, "target")
+			requestShape := payload["request_shape"].(map[string]any)
+			require.Contains(t, requestShape, "target")
 		})
 	}
 }
--- a/internal/mcp/facade_legacy_target_test.go
+++ b/internal/mcp/facade_legacy_target_test.go
@@ -203,9 +203,8 @@ func TestAnalyzeAdvertisedTargetMatchesAcceptedTarget(t *testing.T) {
 			aliased++
 		}
 		capability := srv.facadeCapability(spec, true)
-		shape, _ := capability["request_shape"].(map[string]any)
-		arguments, _ := shape["arguments"].(map[string]any)
-		_, publishesTarget := arguments["target"]
+		requestShape, _ := capability["request_shape"].(map[string]any)
+		_, publishesTarget := requestShape["target"]
 
 		id++
 		result := call(id, "analyze", map[string]any{
--- a/internal/mcp/facade_schema.go
+++ b/internal/mcp/facade_schema.go
@@ -28,7 +28,7 @@ func facadePublicCapabilitySchema(
 ) map[string]any {
 	definition := facadeToolDefinition(spec.Facade)
 	staticProperties := definition.InputSchema.Properties
-	requestArguments, _ := requestShape["arguments"].(map[string]any)
+	requestArguments := requestShape
 
 	properties := make(map[string]any)
 	requiredTop := make(map[string]struct{})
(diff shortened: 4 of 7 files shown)


zzet (approved):
@tiendungdev overall look good, a few points worth addressing:
 
1. The error message hardcodes "Pass operation/query/options at the top level" — search-specific wording. For read.file or change.impact the real fields are operation/target/options. It should derive the field list from the operation's schema. The trailing clause and the accepted_shape data still self-correct, so it's not blocking.

2. %T in received %s leaks Go type names for non-string values ("received []interface {}", "received float64").

3. The facadeRequestShape docstring still says the shape "shows where those fields belong in the stable public envelope" — the envelope is gone; and "…; Pass…" has a mid-sentence capital.

tiendungdev:
Addressed the three non-blocking review notes in 10dfb3d2: accepted top-level fields now come from the selected operation schema, invalid envelope values report JSON types instead of Go implementation types, and the request-shape comment no longer describes the removed envelope. Added coverage across search.symbols, read.file, and change.impact. Verified with focused internal/mcp tests, go vet ./internal/mcp, and git diff --check.

zzet (approved):

PR #573, opened 2026-08-13
fix(mcp): scope contract verification paths

### LINKED ISSUE #556: change.contract emits verification commands with graph repo prefixes as filesystem paths
## Summary

`change(operation:"contract")` can emit a Go verification command containing the graph repository prefix as if it were a directory below the repository root. The command is not executable from the repository root, even though it is presented as the stop-condition command.

## Reproduction

Observed while reviewing changed symbols in the tracked project/repository `gortex-pr544`:

```text
verification_command:
go build ./gortex-pr544/internal/mcp/... && go test -race ./gortex-pr544/internal/mcp/...
```

The actual checkout root already is the `gortex-pr544` repository. There is no nested `gortex-pr544/` directory, so the executable command is:

```text
go build ./internal/mcp/... && go test -race ./internal/mcp/...
```

## Code path

`internal/mcp/change_contract.go::buildVerificationCommand` derives package directories directly from `prediction.touchedFiles`:

```go
dirs["./"+filepath.ToSlash(filepath.Dir(f))+"/..."] = true
```

In a multi-repo/project graph, those file paths can be graph-qualified (`<repo-prefix>/internal/...`) rather than repository-root-relative filesystem paths.

## Impact

- the command advertised as the completion proof fails when copied verbatim;
- agents may report verification as blocked even though the correct package command is available;
- shell strings do not communicate which tracked repo should be used as the working directory when multiple repositories are affected.

## Suggested contract

At minimum, strip the selected graph repo prefix before synthesizing a repo-local package path.

For multi-repo changes, prefer a structured command plan:

```json
{
  "verification": [
    {
      "repo": "gortex-pr544",
      "workdir": "<resolved tracked repo root>",
      "argv": ["go", "test", "-race", "./internal/mcp/..."]
    }
  ]
}
```

The human-readable `verification_command` can remain as a projection of that plan for compatibility.

## Acceptance criteria

- generated Go package paths are valid from the declared/resolved working directory;
- graph repo prefixes are never mistaken for filesystem subdirectories;
- multi-repo changes produce one explicitly scoped command per repo, or a clear refusal if a single command cannot represent them;
- tests cover single-repo, repo-prefixed multi-repo, and Windows separator inputs.

## Environment

- Gortex daemon: `v0.63.2+6a2905e`
- Windows, project-scoped multi-repo query
- observed while implementing the read-side slice of #545

tiendungdev:
## Summary

- normalize graph-qualified verification paths to repository-local Go package paths
- preserve existing `verification_command` output for single-repository changes, including Windows path inputs
- refuse ambiguous multi-repository command synthesis with an explicit warning reason instead of emitting an invalid shell command
- expand unit coverage for repo-relative, prefixed, Windows, covering-test, multi-repo, and docs-only inputs

## Why

`change.contract` currently treats graph repository prefixes as directories below the checkout root. A command such as `go build ./gortex-pr544/internal/mcp/...` therefore fails when run from the `gortex-pr544` repository root.

The fix carries the resolved repo scope into the prediction, also derives scopes from graph nodes and impact metadata, and strips only a known prefix. It does not guess from the first path segment.

## Verification

- `go test ./internal/mcp -run '^(TestBuildVerificationCommand|TestAssembleEnvelopeReportsMultiRepoVerificationRefusal|TestChangeContractSymbolSource|TestChangeContractDiffSourceCleanTree)$' -count=1 -timeout=5m`
- `go test -race ./internal/mcp -run '^(TestBuildVerificationCommand|TestAssembleEnvelopeReportsMultiRepoVerificationRefusal)$' -count=1 -timeout=5m`
- `go vet ./internal/mcp`
- `go build ./internal/mcp`
- `git diff --check`

Closes #556


zzet (changes requested):
Hey @tiendungdev.

I found 2 issues:

1. Blocking: cross-repo impact silently kills the verification command

verificationRepoPrefixes folds p.impact.ByRepo into the refusal set (internal/mcp/change_contract.go:524-530). But ByRepo is keyed by the repo prefix of dependents at depth 1–3, and AnalyzeImpactContext populates it only when CrossRepoImpact is true (internal/analysis/impact.go:215-227). So any single-repo change whose blast radius reaches a second tracked repo now trips the multi-repo refusal.

Probe on the PR branch — change set entirely in gortex, dependents in gortex + gortex-cloud:

cmd=""  err=cannot synthesize one verification command for multiple repositories: gortex, gortex-cloud verdict="warn"  verification_command="" stop_condition="Done when no new tree-sitter parse errors are introduced."

Before this PR that same input produced go build ./internal/mcp/... && go test -race ./internal/mcp/... — correct, since source:diff already carries repo-relative paths (joinHunksToSymbols: "ChangedFiles keeps the diff-relative paths"). So the diff source had no bug to fix here, and it's the one that acquires the regression.

Three things go wrong at once: the command is dropped, the verdict escalates allow → warn, and stop_condition collapses to a trivially-satisfiable sentence. The riskiest changes get the weakest stop condition — the inverse of the intent. It also misses #556's own acceptance criterion: a single-repo change is representable by one command, so refusal isn't licensed there.

Fix is small: derive the prefix set from the change set only (p.repoPrefixes + seed p.nodes), drop the impact.ByRepo loop. If cross-repo TestFiles are the real worry, filter those to the command's repo rather than refusing the whole command.

2. Minor: prefix strip can eat a real path segment

repoLocalVerificationPath (:540-547) strips prefix+"/" unconditionally, but diff-source paths are already repo-relative. When a repo's prefix matches one of its own top-level directories, a real segment is consumed:

touchedFiles: ["web/handler/handler.go"], repoPrefixes: ["web"] → "go build ./handler/... && go test -race ./handler/..."   (want ./web/handler/...)

Narrow, and web is in your tracked list. It produces a wrong-but-plausible command — the same failure class the PR exists to fix. Guarding the strip (only for graph-sourced paths, or only when the stripped form resolves) closes it.

tiendungdev:
Addressed both review findings in 429cc9ef.

- Repository scope now comes only from the change set (explicit prefixes + seed nodes), not cross-repo impact dependents.
- Graph-qualified impact/simulation test paths are filtered back to the changed repository.
- Diff paths remain repo-relative, so a repository prefix such as web no longer strips a legitimate top-level web/ directory.
- Added regression coverage at both command synthesis and envelope/stop-condition levels.

Verification:
- go test -race ./internal/mcp with the focused BuildVerificationCommand and AssembleEnvelope regression set
- go test ./internal/mcp with the broader ChangeContract/VerificationCommand set
- go vet ./internal/mcp
- git diff --check

zzet (changes requested):
The latest head (`429cc9ef`) fixes the earlier cross-repository-impact refusal and top-level prefix collision, but two blocking `source=edit` paths remain.

1. **Blocking — absolute/file-URI WorkspaceEdit paths still produce invalid verification commands.**

`groupEditByFile` preserves the caller path as `overlayPath` (`tools_simulate.go:772-834`), including absolute paths decoded from `file://`. `buildSimulation` copies that spelling into `step.touchedFiles` (`tools_simulate.go:341-383`), while `lowerEditSource` marks the entire slice as graph-qualified without retaining the owning repository (`change_contract.go:236-245`). `repoLocalVerificationPath` only strips a known graph prefix; it never relativizes an absolute path against the repository root (`change_contract.go:536-563`). An edit of `/workspace/repo/internal/foo.go` can therefore emit `./workspace/repo/internal/...` instead of `./internal/...`; Windows/file-URI inputs fail similarly and can expose host-path components in the response.

2. **Blocking — newly added files bypass repository-scope and multi-repository detection.**

Repository scope is inferred from base-graph nodes (`change_contract.go:139-149,516-533`). A newly added file has no base node, so a graph-qualified path such as `gortex/newpkg/foo.go` retains the repository prefix in the generated command. With new files in `api/...` and `worker/...`, no prefixes are recovered, the `len(prefixes) > 1` guard is skipped (`change_contract.go:577-584`), and one invalid command is emitted for both apparent directories. This violates #556: multi-repository changes must produce explicitly scoped commands or a clear refusal.

Please canonicalize every touched edit file to per-file repository ownership plus a repository-local path before command synthesis; a slice-wide boolean cannot represent the accepted path forms. Add end-to-end `lowerEditSource` coverage for POSIX absolute/file URI, Windows absolute/file URI, a brand-new graph-qualified file, and one WorkspaceEdit adding files to two repositories. When synthesis is refused, the stop condition should also remain actionable instead of falling back to parse-errors-only completion.

All current checks are green, and `internal/mcp` passes both at the PR head and after a clean merge onto current `main`; these are uncovered behavioral failures rather than build failures.

tiendungdev:
Addressed the latest `source=edit` review blockers in `205d7125`.

- Replaced the slice-wide `touchedFilesRepoQualified` flag with per-file verification ownership: repository prefix plus repository-local path.
- `lowerEditSource` now canonicalizes every WorkspaceEdit file through the existing confined path resolver before command synthesis, so absolute paths and `file://` URIs cannot leak host path components into verification commands.
- New graph-qualified files are attributed without requiring a base graph node.
- WorkspaceEdits that add files to multiple repositories now retain both owners and produce the explicit multi-repository refusal.
- Verification refusal now emits an actionable stop condition requiring separate repository-scoped commands to be selected and run successfully.
- Symbol/range sources were migrated to the same per-file representation, preserving their existing behavior.

Added end-to-end `lowerEditSource` coverage for:

- native absolute path and native `file://` URI;
- a brand-new graph-qualified file with no base node;
- one WorkspaceEdit adding new files to two repositories;
- the refusal envelope and actionable stop condition.

The native absolute/URI test name contains `NativeSeparator`, so it runs both in the regular POSIX test matrix and in the focused native Windows MCP CI step.

Verification:

```
go test ./internal/mcp -run '^Test(ChangeContract|BuildVerificationCommand|AssembleEnvelope|LowerEditSource)' -count=1 -v -timeout=10m
PASS
ok github.com/zzet/gortex/internal/mcp 1.768s

go vet ./internal/mcp
PASS

git diff --check
PASS
```

The Go checks were executed in `golang:1.26` on Linux; native Windows path execution is covered by the PR's focused `NativeSeparator` CI step.

tiendungdev:
CI follow-up pushed in ee779c30. The failed govulncheck job never reached the scanner: outer setup-go restored the module cache, then the nested setup-go in govulncheck-action restored into the same read-only cache and tar failed with Cannot open: File exists. The action now reads the Go version only from go.mod and disables only its inner cache; outer setup-go remains the sole cache owner. actionlint v1.7.7 and git diff --check pass.

zzet:
p.s. No reason to skip the cache (it may impact the CI timing)

tiendungdev:
Applied the maintainer direction in b3ea2f25. Merged upstream/main including PR #579 without force-push, then restored the upstream govulncheck cache configuration. The net workflow and go.mod/go.sum now match upstream/main. Focused change-contract tests, go vet ./internal/mcp, and git diff --check pass on Go 1.26.6 Linux.

zzet (approved):

Issue #560, opened 2026-08-12
csharp: a params parameter is invisible to the grammar, erasing arity evidence and param nodes


zzet:
Two related gaps in how C# parameter lists reach the graph, both traced to the vendored grammar and both reducing what overload resolution can do.

## 1. A `params` entry is not a `parameter` node, so the method gets no arity at all

In tree-sitter-c-sharp v0.23.5 `_parameter_array` is a hidden rule and `params` is an anonymous token, so a `params` entry never materialises as a `parameter` node — the list flattens into loose `array_type` / `generic_name` + `identifier` siblings. Verified for `object[]`, `int[]`, `object?[]`, `ReadOnlySpan<int>` and `List<int>`.

`csharpParamArity` (`internal/parser/languages/csharp_arity.go:111`) therefore bails with `ok=false` — deliberately, since a partial count would be worse than none — and the method carries no `param_count` / `param_required` / `param_variadic`. It is then universally applicable, so it never narrows out and keeps its siblings ambiguous.

Measured: `l.Log("hello")` (`arg_count=1`) against `Log(this ILogger, string)` and `Log(this ILogger, string, params object[])` stays `unresolved::*.Log`. Dropping only the `params` keyword makes it bind. This is exactly the `Log(msg)` / `Log(fmt, params object[])` shape .NET logging helpers use.

Corollary: `param_variadic` can never be stamped, so the variadic branch in `csharp_applicability.go:137-139` is unreachable today. It is correct forward-compatible logic, not dead weight — but it is currently untested by anything real.

## 2. `params` and discard parameters emit no `KindParam` node

`internal/parser/languages/csharp_function_shape.go:38-40,55` skips any child that is not a `parameter` (which excludes `params` entries per the above) and any parameter named `_`. Measured: `Log(string fmt, params object[] rest)` → 1 param node; `ParamsOnly(params int[] xs)` → 0; `Disc(this ILogger l, int _, string name)` → 2.

Positions are correct since #536 (the discard occupies its slot), but the nodes are still absent, so any consumer counting `EdgeParamOf` under-counts the signature.

## Possible fix

Both would be solved by recognising the flattened `params` shape and reconstructing the entry. That is grammar-version-specific, so it needs a test that fails loudly if a grammar bump changes the shape — the current `ok=false` behaviour is the safe fallback and should stay as the default when the shape is unrecognised.

pbednarcik:
[...]

PR #575, opened 2026-08-13
fix(csharp): recover params parameter shape

tiendungdev:
## Problem

tree-sitter-c-sharp v0.23.5 flattens a params declaration into loose list children instead of a parameter node. The current parameter-only walk therefore drops the declaration from both arity evidence and the function-shape graph. A params overload remains universally applicable, while EdgeParamOf consumers see an incomplete signature.

## Fix

- normalize ordinary parameter nodes and the exact flattened params/type/identifier grammar shape into one csharpParamEntry representation
- share that representation between declaration arity stamps and KindParam/EdgeParamOf emission
- keep the flattened-shape recognizer strict and retain fail-open arity behavior for unknown grammar shapes
- retain an underscore method parameter as a real parameter node with Name "_" and its positional slot
- prove the extractor stamp reaches resolver applicability with a nullary extension-overload fixture

## Coverage

The grammar regression tests cover params declarations across arrays, nullable, spans, and generic types. They pin underscore parameter identity and positions. The end-to-end resolver fixture extracts a nullary overload beside an overload that still requires fmt before its params tail; the nullary call binds only when extracted applicability metadata reaches resolution.

The remaining normal-form-vs-expanded-form tie-break for calls where both fixed and params overloads are applicable is intentionally not addressed here.

## Verification

- go test -race ./internal/parser/languages -run CSharp -count=1 -timeout=10m
- go test -race ./internal/resolver -run CSharp -count=1 -timeout=10m
- focused extractor and extension-overload tests
- git diff --check

Note: go vet on Windows retains the pre-existing internal/parser/languages/grammar_load_windows.go:33 unsafe.Pointer diagnostic outside this diff.

Refs #560


zzet (changes requested):
The extraction fix itself checks out. I ran the branch beside its base commit and diffed 15 parameter shapes through the extractor: a `params` entry now stamps arity and emits `KindParam` / `EdgeParamOf` / `EdgeTypedAs` across `object[]`, `int[]`, `object?[]`, jagged `object[][]`, qualified `System.Object[]`, `ReadOnlySpan<int>`, and generics carrying an internal comma (`Dictionary<string,int>[]` — nested, so the new list-level `,` split does not mis-split it). `ref` / `out` / `in`, defaults containing a comma inside a string literal, lambdas, local functions, delegates and record primary constructors come out byte-identical to base. And there is a real resolution gain end-to-end: `l.Log()` against `Log(this ILogger)` + `Log(this ILogger, string, params object[])` moves from `unresolved::*.Log` to binding the nullary overload.

Three things I would like changed before this lands.

### 1. The `<discard>` rename slips past a test that asserts the opposite contract

`internal/parser/languages/csharp_overload_evidence_test.go:170` still asserts:

```go
assert.NotContains(t, pos, "_", "a discard names no value, so it emits no node")
```

`pos` is keyed by node **Name**, so that assertion passes only because the name became `<discard>`. This PR reverses the contract the assertion and its comment describe, and leaves both in place — the test now passes for a reason opposite to what it claims. Meanwhile `TestCSharpExtractor_DiscardEmitsParamNode` asserts the node by ID, type and position but never asserts Name, so `<discard>` is pinned by nothing at all.

On the merits I also think `_` should stay. In a C# *method declaration* `_` is an ordinary parameter identifier — usable in the body, and usable as a named argument (`Handle(_: 1, name: "x")`). Parameter-position discards exist only for lambdas. Every other extractor skips `_` outright (`go_function_shape.go:87`, `rust_function_shape.go:169`, `py_function_shape.go:146`, `ts_function_shape.go:435`, `java_function_shape.go:80`), so C# would now diverge twice: it emits the node *and* renames it, leaving the ID (`#param:_@0`) disagreeing with the Name.

Please keep `Name: "_"` and rewrite `csharp_overload_evidence_test.go:170` to assert the new behavior directly (the discard emits a node at position 0 under its own name).

### 2. Stale comment at `internal/parser/languages/csharp.go:1288`

> …because those carry no default-value marker (and a discard parameter emits none at all)…

No longer true after this change.

### 3. Nothing in the PR fails before and passes after

`internal/resolver/csharp_params_applicability_test.go` passes unchanged on the base commit — I ran it there. It hand-builds the meta map, so it never exercises the link this PR creates: that the extractor's stamp actually reaches the applicability window. `TestResolveCSharpExtension_VariadicWidensArity` passed on base too, for the opposite reason (no stamp at all made the overload universally applicable).

The fixture that does discriminate is a nullary sibling beside a `params` overload whose required count the call misses:

```csharp
// Ext.cs
public static void Log(this ILogger l) { }
public static void Log(this ILogger l, string fmt, params object[] args) { }

// Caller.cs
l.Log();   // base: unresolved::*.Log — both applicable
           // this PR: binds Ext.cs::LogExt.Log
```

That belongs in `csharp_extension_overload_test.go` alongside the existing arity cases.

### On `Fixes #560`

Both gaps the issue documents are genuinely closed. But the issue's own *measured* repro is not: `l.Log("hello")` against `Log(this ILogger, string)` and `Log(this ILogger, string, params object[])` still resolves to `unresolved::*.Log` after this change — verified. Both overloads are applicable at one argument, and there is no "normal form beats expanded form" tie-break (ECMA-334 §12.6.4.2) to split them. This PR is the prerequisite for that fix, not the fix. Either drop the auto-close and re-scope #560 to the remaining half, or file a follow-up so the tie-break does not get lost.

### Residuals — fine to defer, worth recording

- `[NotNull] params object[] xs` still yields no arity and no param node: the `attribute_list` hoists to list level, so the segment is 4 nodes and fails the strict `len(segment) == 3` check. Ordinary attributed parameters (`[My] int a`) are unaffected. Fail-open is the right behavior meanwhile, and dropping a leading `attribute_list` from the segment would close it.
- `params scoped ReadOnlySpan<int> xs` folds the modifier into the type slot: `type="scoped ReadOnlySpan<int>"` and an `EdgeTypedAs` to `unresolved::scoped ReadOnlySpan`, a target with a space in it that can never resolve.

Everything else looks good — the strict recognizer with fail-open fallback is the right shape for a grammar-version-specific fix, and the shared `csharpParamEntry` removes the duplicated walk cleanly. CI is green and the branch merges onto main without conflict.

tiendungdev:
Addressed the review in 6f09faa4.

- Kept the declared method parameter name as underscore, matching its symbol ID and valid C# named-argument semantics.
- Rewrote the stale discard assertion/comment and pinned Name directly.
- Removed the stale arity comment.
- Added an extractor-to-resolver regression: a nullary extension sibling now binds only because the extracted params overload still requires fmt.
- Changed the PR body to Refs #560 and documented the remaining normal-form vs expanded-form tie-break.

Verification: both parser/languages and resolver CSharp suites pass under -race, focused regressions pass, and git diff --check is clean. go vet retains the pre-existing Windows unsafe.Pointer diagnostic at grammar_load_windows.go:33.

zzet (changes requested):
Requesting changes for one blocking upgrade-path issue.

This PR changes persisted C# extraction output (`param_count`, `param_required`, `param_variadic`, `KindParam`, `EdgeParamOf`, and `EdgeTypedAs`) but leaves `extractorVersions["csharp"]` at `10` in `internal/indexer/extractor_version.go`. Both Merkle and non-Merkle restaging use that version. As a result, existing unchanged `.cs`, `.razor`, and `.cshtml` files keep their v10 graph after upgrade, so the resolver continues to miss the recovered `params` evidence until those files are touched or manually reindexed.

Please bump the C# extractor version to `11`, update its reason, and add a regression assertion for the C# salt/staleness behavior.

Everything else at `6f09faa42533d003aeef8cd3f9bad0c174550b94` checks out: fresh-index behavior is correct, the prior review items are addressed, all checks pass, focused race tests pass, and I found no security or API-compatibility blocker.

tiendungdev:
Addressed the upgrade-path blocker in a58d6334.\n\n- bumped the C# extractor version from 10 to 11 with a reason tied to the params shape/arity evidence change;\n- added staleness coverage proving a persisted v10 C# graph is flagged after upgrade;\n- added Merkle-salt assertions for .cs, .razor, and .cshtml, all expecting csharp@11.\n\nVerification:\n- go test ./internal/indexer -run '^TestStaleLangsDetection$' -count=1 -timeout=8m\n- git diff --check\n\nBoth passed.

tiendungdev:
CI follow-up pushed in 4b93f5d7. The failed govulncheck job never reached the scanner: outer setup-go restored the module cache, then the nested setup-go in govulncheck-action restored into the same read-only cache and tar failed with Cannot open: File exists. The action now reads the Go version only from go.mod and disables only its inner cache; outer setup-go remains the sole cache owner. actionlint v1.7.7 and git diff --check pass.

zzet:

tiendungdev:
Applied the maintainer direction in 2941c96f. Merged upstream/main including PR #579 without force-push, then restored the upstream govulncheck cache configuration. The net workflow and go.mod/go.sum now match upstream/main. Focused indexer staleness, C# parser, and C# resolver tests plus vet for all three packages pass on Go 1.26.6 Linux; git diff --check also passes.

zzet (approved):

PR #592, opened 2026-08-16
tstypes: true up spool capacity hints and page-cap/deadline comments

pbednarcik:
The small remainders from the #530 review (the `rows.Err` one you already fixed in
d3fc1a51, together with the linter that keeps the whole class out).

- `appendFiles` sized its per-chunk slices for 4 classes per file when there are 5.
  Capacity hint only, as you noted — now derived from the enum (`factClassCount`) so a
  sixth class can't repeat the drift.
- The page-cap comment now states that the byte cap charges only the paged class's
  payloads and that the imports side-fetch rides on top, so peak page bytes can modestly
  exceed it — the cap's actual contract rather than the implied one.
- The coverage-walk comment now records the deadline trade-off from your review: the walk
  is the first work under the pass deadline and flushes no edges, so an expiry during it
  yields a zero-edge partial. Accepted rather than special-cased — the walk is decode-free,
  so a budget too small for it could not have completed any apply phase either. If you'd
  rather exempt the walk from the deadline outright, happy to do that instead, but it
  changes budget semantics and didn't seem worth it for a window this small.

No behavior change; `./internal/semantic/tstypes` green, gofmt clean.

### DIFF
--- a/internal/semantic/tstypes/fact_spool.go
+++ b/internal/semantic/tstypes/fact_spool.go
@@ -177,7 +177,7 @@ const (
 	classAliases
 	classCalls
 
-	factClassCount = int(classCalls) + 1
+	factClassCount = iota // 5 — auto-tracks any class added above
 )
 
 // marshalClassPayloads encodes one file's facts as per-class JSON arrays,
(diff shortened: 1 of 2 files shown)


internal/semantic/tstypes/fact_spool.go
```diff
@@ -173,6 +176,8 @@ const (
 	classMetas
 	classAliases
 	classCalls
+
+	factClassCount = int(classCalls) + 1
```
zzet:
```suggestion
	factClassCount = iota // 5 — auto-tracks any class added above
```

zzet (approved):

PR #604, opened 2026-08-17
lsp: solution targeting for csharp-ls (pin env, auto-detect, targeted restore, idle-TTL knob)

pbednarcik:
## Problem

On my production C# monorepo (~117k nodes), the csharp-ls alt provider never produced a
usable workspace:

- The pre-spawn `dotnet restore` was a bare directory restore, which failed with
  **MSB1011** in my checkout — my daemon log had five historical pre-restore failures and
  zero successes, so the assets the workspace load needs were never written. Targeting the
  restore needs a concrete solution path, which gortex had no way to determine.
- Which solution the server loads was neither controllable nor observable. Current
  csharp-ls (I run 0.24.0) discovers solutions on its own — recursive `.sln`/`.slnx` glob
  with a most-projects heuristic since 0.19.0 — which is a reasonable default but gives
  the operator no override when it picks wrong, and gortex no known target for the
  restore. (An earlier revision of this PR claimed csharp-ls's discovery predates `.slnx`
  and can't handle multi-solution roots; that's stale for current versions — review
  caught it, comments and this body are corrected.)

## What this does

1. **Solution pin + auto-detect** — new `GORTEX_LSP_CSHARP_SOLUTION` env names the solution
   csharp-ls should load. Validated per root: an entry only applies when it resolves to a
   file *inside* that workspace root, so a daemon-global value is multi-repo safe (entries
   that don't resolve are ignored for that root). With no env, a root carrying exactly one
   `.sln`/`.slnx` auto-detects it; anything else keeps the server's own discovery. The
   resolved solution is appended as `--solution <file>` in `dialOrSpawn` (same pattern as
   the jdtls `-data` injection), unless the caller already pinned `--solution`/`-s` (bare
   or `=`-joined). An explicit falsy value (`0`/`off`/`false`/`no`/`none`) disables
   injection and auto-detect entirely, and the spawn logs the resolved pin with its
   provenance (`env` vs `auto-detect`) plus a warning for every env entry that did not
   resolve for the root.
2. **PATH-style list** — the env accepts an `os.PathListSeparator` list; each tracked root
   uses the first entry that resolves inside it, so one setting serves several repos with
   different solutions.
3. **Opt-in resolve-time C# helper** — the resolver LSP mux only wired TypeScript and
   Python; C# was unreachable regardless of provider health. `GORTEX_LSP_RESOLVER_CSHARP=1`
   (default **off**) adds csharp-ls for repos with C# intent (root-level
   `.sln`/`.slnx`/`.csproj`). Off by default because a Roslyn workspace load costs real
   memory and startup time — same opt-in posture as eager LSP and the jdtls trust-build env.
4. **`GORTEX_LSP_IDLE_TTL`** — Go-duration override for the router's idle reap timeout
   (`<= 0` disables reaping). Useful operator knob when a long enrichment window should
   keep a warm server around.

## Evidence (live, Windows, real monorepo)

- `lsp: csharp pre-restore complete (NuGetAudit suppressed)` — first successes ever after
  five MSB1011 failures; the restore now targets the resolved solution instead of the
  ambiguous directory.
- The csharp-ls startup banner confirms `solutionPathOverride = Some <the tracked root's
  own solution>` — picked up with **zero config** via auto-detect, since the tracked root
  carries exactly one root-level `.slnx`.
- The env's per-root validation was exercised live: a value pointing at an umbrella
  solution *outside* the tracked root was correctly ignored (fail-safe, fell back to
  auto-detect).
- Loaded this way, the server answers `workspace/symbol` ~30s after spawn and serves a
  full 84-minute enrichment pass with zero errors and zero reconnects.

## Measured impact

Together with the Windows answer-join fix (#603), the first complete enrichment
pass on my monorepo produced **33,841 LSP-origin edges at confidence 1.0** (26,178 calls,
3,426 implements, 2,944 overrides, 1,187 extends, …), confirming 5,474 heuristic edges and
adding 26,782 new ones. This PR is the prerequisite half: without a solution, csharp-ls
has nothing to answer from — with it, the workspace loads in ~30 seconds.

## Tests

- `csharp_solution_test.go`: env-inside-root wins (relative + absolute), env outside root
  ignored per root, PATH-list pins per root with auto-detect fallback, auto-detect table
  (single `.sln`, single `.slnx`, multi-solution ambiguity, nested-only, empty), argv
  injection + caller-override wins, targeted-vs-bare restore argv.
- `lsp_test.go`: C# intent detection (root-only glob), resolver opt-in gate parsing.
- `idle_ttl_env_test.go`: unset/valid/zero/negative/unparseable durations.
- Full `go test ./...` green vs my Windows baseline; acceptance battery on my counted C#
  fixture repo ×2 byte-identical.

### DIFF
--- a/docs/lsp.md
+++ b/docs/lsp.md
@@ -168,9 +168,12 @@ The router applies these defaults in `cmd/gortex/server.go` and
 | `WithMaxAlive`    | 6 servers    | LRU eviction kicks in when a seventh distinct server would spawn — the least-recently-used one closes. |
 
 These defaults suit a polyglot workspace where most languages are
-touched only intermittently. Override them by editing the
-`lsp.NewRouter(...).With...` chain in your build if you need a longer
-warm pool or a tighter memory bound.
+touched only intermittently. The idle timeout is overridable at runtime:
+`GORTEX_LSP_IDLE_TTL` accepts a Go duration (`45m`, `2h`); `0` or a
+negative value disables reaping entirely. Useful when a long enrichment
+window should keep a warm server around — a big Roslyn workspace is
+expensive to reload. The reaper interval and max-alive bound are still
+compile-time (`lsp.NewRouter(...).With...`).
 
 ### Workspace roots
 
@@ -400,8 +403,10 @@ Gortex applies two complementary, C#-scoped fixes, **both on by default**:
 
 | Env var | Default | Effect |
 | --- | --- | --- |
-| `GORTEX_LSP_CSHARP_RESTORE` | on | Before spawning the C# server, run `dotnet restore -p:NuGetAudit=false` in the workspace so the MSBuild workspace loads every project (root-cause fix). Best-effort: a failure logs and falls through to a normal spawn; skipped on passive IDE attach and when `dotnet` is not on `PATH`. |
+| `GORTEX_LSP_CSHARP_RESTORE` | on | Before spawning the C# server, run `dotnet restore <resolved solution> -p:NuGetAudit=false` in the workspace (falling back to a bare directory restore when no solution resolves) so the MSBuild workspace loads every project (root-cause fix). Best-effort: a failure logs and falls through to a normal spawn; skipped on passive IDE attach and when `dotnet` is not on `PATH`. |
 | `GORTEX_LSP_CSHARP_DIAG_FILTER` | on | Strip diagnostics whose code is the `NU####` NuGet family from `publishDiagnostics` before storing / fanning out (symptom fix). Deliberately narrow — real `CS####` compiler diagnostics always pass through. |
+| `GORTEX_LSP_CSHARP_SOLUTION` | unset | Solution csharp-ls loads and the pre-restore targets — a PATH-style list of `.sln`/`.slnx` paths (relative to a workspace root, or absolute inside one); each tracked root uses the first entry that resolves inside it, so one daemon-global value serves a multi-repo daemon. Unset auto-detects a lone root-level solution; a falsy value (`0` / `off` / `false` / `no` / `none`) disables injection and auto-detect, leaving csharp-ls's own recursive discovery untouched. The resolved pin and any ignored entries are logged at spawn. |
+| `GORTEX_LSP_RESOLVER_CSHARP` | off | Opt-in: add csharp-ls to the resolve-time LSP helper for repos with C# intent (root-level `.sln`/`.slnx`/`.csproj`). Any set, non-falsy value enables (`1` / `true` / `on` / `yes`); off by default because a Roslyn workspace load costs real memory and startup time. |
 
 Set either to a falsey value (`0` / `off` / `false` / `none`) to disable it —
 e.g. `GORTEX_LSP_CSHARP_RESTORE=0` for offline / air-gapped indexing or to
@@ -440,8 +445,8 @@ for repositories you trust.
   the one workspace — the spec stays available for every other tracked
   repo.
 - **Server keeps restarting:** the idle reaper closed it, then the
-  next request re-spawned. Increase `WithIdleTimeout` if this hurts
-  warm-cache benchmarks.
+  next request re-spawned. Raise `GORTEX_LSP_IDLE_TTL` (Go duration;
+  `0` disables reaping) if this hurts warm-cache benchmarks.
 - **High memory under polyglot load:** lower `WithMaxAlive` from 6 to
   3-4. The LRU evicts the least-recent server transparently.
 
--- a/internal/semantic/lsp/csharp_solution.go
+++ b/internal/semantic/lsp/csharp_solution.go
@@ -12,10 +12,18 @@ import (
 // entry that resolves inside it. The variable is daemon-global while
 // solutions are per-workspace, so per-root resolution is what makes one
 // setting serve a multi-repo daemon: entries that do not resolve inside a
-// given root are ignored for that root, and several repos with different
-// solutions each pick their own entry. Empty (the default), or no entry
-// resolving, auto-detects: a workspace root carrying exactly one .sln/.slnx
-// uses it; anything else keeps the server's own discovery.
+// given root are ignored for that root (and logged), and several repos with
+// different solutions each pick their own entry.
+//
+// Empty (the default) auto-detects: a workspace root carrying exactly one
+// root-level .sln/.slnx uses it; anything else keeps the server's own
+// discovery. Current csharp-ls discovers on its own — recursive
+// .sln/.slnx glob with a most-projects heuristic since 0.19.0, .slnx since
+// 0.18.0, per-.csproj fallback — so the pin's value is determinism, a
+// concrete target for the pre-spawn restore, and an operator override, not
+// a capability the server lacks. An explicit falsy value
+// (0/false/no/off/n/none) disables injection and auto-detect entirely,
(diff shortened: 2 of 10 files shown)


zzet (changes requested):
Thanks for this — the live evidence on a real 117k-node monorepo is exactly the kind of grounding these LSP knobs need, and the core mechanism is sound. I verified the wiring end to end and it holds up:

- `NewClient` has exactly one production call site (`provider.go:1923`, inside `dialOrSpawn`), so the `--solution` injection reaches every spawn path including the pooled `SpawnProviderForResolver` — no bypass.
- `lsp.NewRouter` likewise has exactly one production construction site, and it's the one you patched, so `GORTEX_LSP_IDLE_TTL` has no unwired path.
- `Router.Reap()` early-returns on `idleTimeout <= 0`, so the negative→0 clamp genuinely means "never reap".
- `NewProviderFromSpec` sets `args = alt.Args` on fallback, so csharp-ls gets a clean argv rather than the omnisharp spec's `-lsp`.
- The tests are non-vacuous — both conjuncts of `pathInsideRoot && solutionFileExists` are independently covered.

Merging current `main` into this head is clean: `golangci-lint` 0 issues, `go build ./...` OK, `internal/semantic/lsp` and `internal/serverstack` pass under `-race`.

Two things I'd like addressed before merge.

## 1. The stated premise is stale for current csharp-ls, and auto-detect regresses because of it

The PR body and the comments in `csharp_solution.go` say csharp-ls's discovery "predates `.slnx`" and "has no answer for a root that carries more than one solution file — the workspace silently loads nothing". That was true for older csharp-ls, but not for current versions. From upstream `src/CSharpLanguageServer/Roslyn/Solution.fs` on `main`:

- it globs `["*.sln"; "*.slnx"]` with `SearchOption.AllDirectories` — **recursive**, not root-level;
- on multiple hits it picks the max `SolutionFile.Parse(p).ProjectsInOrder.Count`;
- with no solution it falls back to loading every `.csproj`/`.fsproj` recursively.

And from the changelog: `.slnx` support landed in 0.18.0 (2025-06-23), and 0.19.0 (2025-08-20) added "Apply simple heuristics to select a .sln/.slnx file when multiple are found". `-s/--solution` has existed since 0.1.3.

The problem this creates is in the **zero-config** path. `csharpSolutionFor`'s auto-detect is root-level only (`os.ReadDir`, skips directories), so for a repo with a small root-level `Build.sln` alongside a larger nested `src/Everything.sln`, csharp-ls previously picked the nested one on project count and now gets hard-pinned to the root one. The loaded workspace silently shrinks for a user who set no env var, with no log line and no way to opt out — auto-detect always fires when exactly one root solution exists.

The explicit pin is clearly worth having; it's the implicit one I'd like narrowed. Any of these works for me:

- gate auto-detect behind the env var (pin only when asked), or
- keep auto-detect but skip it when a deeper solution exists, or
- keep it as-is and add an off switch,

plus a log line naming the resolved solution and how it was chosen. That last part matters independently: right now a non-resolving env entry is dropped silently, so a mis-spelled value presents to the operator as the exact pre-PR symptom.

Please also correct the comments in `csharp_solution.go` and `registry.go` — as written they'll mislead the next reader about what csharp-ls does today. Worth noting which csharp-ls version you're running, since that likely explains the difference between your box and upstream `main`.

## 2. The three new env vars are undocumented, and one doc statement is now wrong

`docs/lsp.md` is the only place in the repo that enumerates `GORTEX_LSP_*`, and none of `GORTEX_LSP_CSHARP_SOLUTION`, `GORTEX_LSP_RESOLVER_CSHARP`, `GORTEX_LSP_IDLE_TTL` appear in any `.md` file. Two specifics:

- The C# table at ~line 403 still describes the pre-restore as bare `dotnet restore -p:NuGetAudit=false`. That's now inaccurate on the **default** path, not just the opt-in one, since `maybeCSharpPreRestore` is gated on `servesCSharp()`.
- Lines ~171 and ~442 tell operators to edit the `lsp.NewRouter(...)` chain and rebuild in order to change the idle timeout — exactly what `GORTEX_LSP_IDLE_TTL` now makes unnecessary.

`GORTEX_LSP_RESOLVER_CSHARP` also accepts only `1`/`true`, while the sibling `GORTEX_LSP_RESOLVER` uses `IsFalsyEnv` (which understands `off`/`no`/`n`). Someone will reasonably write `=on` and get silence. Either widen the accepted spellings or document the exact ones.

## Minor

`csharpSolutionArgs`'s "caller already pinned" guard matches only the bare tokens `--solution` / `-s`, so the `=`-joined spelling slips through. It is reachable — `semantic.providers: [{name: omnisharp, command: csharp-ls, args: ["--solution=X"]}]` flows to `p.args` via `SpecWithOverridesConnect` — and would append a second `--solution`. Narrow and self-inflicted, so fine to leave, but cheap to cover with a `strings.HasPrefix` check.

Happy to re-review as soon as the auto-detect scope and the docs are sorted.

pbednarcik:
All addressed in f931fe25; the PR body's Problem section is also rewritten, since it
carried the same stale premise as the comments.

**1. Premise + auto-detect scope.** Verified against my install: I'm running csharp-ls
0.24.0, so your correction applies to my box too — the "discovery predates .slnx / loads
nothing" story in the comments was stale. What I had actually observed pre-PR was the
failed bare restore plus a separate answer-mapping bug (#603) that zeroed every result
regardless; I conflated that with discovery. Comments in `csharp_solution.go` and
`registry.go` now describe current csharp-ls (recursive glob, most-projects heuristic
since 0.19.0) and state the pin's real value: determinism, a concrete target for the
pre-spawn restore, and an operator override.

For the scope I went with your third option — keep auto-detect, add an off switch, log the
choice — rather than skipping when a deeper solution exists. Reason: my own repo is the
inverse of your scenario (the root solution is the big one; the nested one is a small
client tool), so the deeper-solution guard would disable zero-config exactly where it
works, while the project-count comparison csharp-ls does is more than I want to
reimplement Go-side. The off switch rides the same env var: a falsy value
(`0`/`off`/`false`/`no`/`none`) for `GORTEX_LSP_CSHARP_SOLUTION` disables injection and
auto-detect entirely, leaving the server's own discovery untouched. The spawn now logs
`lsp: csharp solution pinned {solution, source: env|auto-detect}`, and a non-resolving env
entry logs a warning with the reason (`outside the workspace root` / `no such file`)
instead of dropping silently — that one was a real trap, agreed.

**2. Docs.** `docs/lsp.md` now documents `GORTEX_LSP_CSHARP_SOLUTION` (with the exact
accepted spellings), `GORTEX_LSP_RESOLVER_CSHARP`, and `GORTEX_LSP_IDLE_TTL`; the restore
row describes the targeted form with the bare-restore fallback; and both
edit-the-`NewRouter`-chain passages now point at `GORTEX_LSP_IDLE_TTL` (reaper interval
and max-alive stay compile-time).

**3. Env spellings.** `GORTEX_LSP_RESOLVER_CSHARP` now accepts any set, non-falsy value —
same value vocabulary as `IsFalsyEnv`, so `on`/`yes` work.

**4. Minor.** The caller-pinned guard also recognizes `--solution=X` / `-s=X`, with tests.

Tests cover the off sentinel, the resolution provenance, the =-joined guard, and the
widened opt-in spellings; full suite at my Windows baseline.

zzet (approved):

PR #608, opened 2026-08-18
feat(contracts): htmx extractor — hx-* attributes as HTTP consumer contracts

### LINKED ISSUE #606: htmx extractor: wire hx-get/hx-post/hx-delete in HTML templates to registered routes as consumer edges
**Is your feature request related to a problem?**

HTTP routes whose only consumers are htmx attributes in HTML templates are reported as **orphan providers**.

Real case: [go-parts](https://github.com/madeinoz67/go-parts) (Go; `html/template` UI with htmx, templates embedded via `go:embed`). The post-task diagnostics (daemon v0.63.4) report 48 orphan providers — about 30 of them are actively consumed by `hx-get` / `hx-post` in `internal/ui/templates/*.html`, e.g.:

```html
<!-- layout.html -->
<input ... hx-get="/ui/parts/search" hx-trigger="keyup changed delay:0ms" ...>
<!-- loc-row.html -->
<tr hx-get="/ui/locations/{{.L.ID}}/exp" hx-swap="afterend">
<!-- loc-bulk-bar.html -->
<form hx-post="/ui/locations/bulk-archive" hx-target="#loc-tbody" ...>
```

Text search **does** index the `.html` files (a text search for `hx-get` returns template hits), so ingestion already works. What's missing is semantic extraction: nothing parses the htmx attributes into graph edges, so consumer/orphan analysis can't see template-side consumption and keeps flagging live routes.

**Describe the solution you'd like**

An htmx / HTML-template extractor that:

1. Parses HTML (including Go `html/template`) for the htmx attributes: `hx-get`, `hx-post`, `hx-put`, `hx-patch`, `hx-delete` (+ `hx-swap-oob` forms if useful).
2. Creates **consumer edges** from the template file node to the matching route-registration node — the registrations `analyze routes` / route-framework detection already produces.
3. Treats template path segments as wildcards when matching, so `/ui/parts/{{.P.ID}}/exp` matches a registered pattern like `/ui/parts/{p1}/exp` (same shape as the `{p1}` placeholder route nodes already emit).
4. Optional bonus: URL string literals in inline `<script>` (`fetch('/ui/...')`) and `<form action=...>` could ride the same template→route matching path.

**Use case**

Primarily MCP tools + CLI: `contracts` / orphan-provider diagnostics and `analyze routes` would gain front-end consumers and stop flagging template-consumed routes as orphans. Benefits any Go + htmx project (a very common pairing), and by extension jinja/htmx or plain static-HTML front-ends against any backend.

**For new language support:**

- Language: HTML (with Go `html/template` as a dialect)
- Tree-sitter grammar available in `go-tree-sitter`: tree-sitter-html exists (`github.com/tree-sitter/tree-sitter-html`); the plain HTML grammar suffices if `{{...}}` inside attribute values is treated as opaque text (community `tree-sitter-go-template` grammars exist if template-native parsing is wanted)
- Key constructs to extract: `hx-*` request attributes (URL value) → template→route consumer edges; `<form action>` and inline-`fetch` URL literals as stretch

Observed on daemon v0.63.4. Happy to test a build against the go-parts graph (48 orphans is a decent regression corpus) if useful.

madeinoz67:
**Title:** `feat(contracts): htmx extractor — hx-* attributes as HTTP consumer contracts`

## Summary

Adds an htmx extractor so server-rendered templates that issue `hx-get` / `hx-post` /
`hx-put` / `hx-patch` / `hx-delete` **attributes** are indexed as HTTP consumer
contracts and paired with the route providers that serve them. Before this, every
route consumed only from htmx template attributes showed up as an orphan provider.

Scope note: this covers the attribute lane. URLs assembled in JavaScript
(`htmx.ajax()`), whole-URL template variables (`hx-get="{{.Action}}"`), `hx-boost`
inheritance, and native `form action` / `a href` consumers are out of scope for
this PR — those routes remain reported as orphans.

Fixes #606

## Changes

- **HtmxExtractor** (`internal/contracts/htmx.go`): attribute-level scan (templates
  are routinely not well-formed HTML until rendered) for the five request verbs,
  emitting `http::<VERB>::<normalized path>` consumer contracts with
  `meta: {framework: htmx, method, raw_path}`. Scanning is comment-aware: HTML
  comments and Go template comments are blanked (offset-preserving) before
  matching, so commented-out markup does not mint consumers. Values that cannot
  produce a trustworthy route ID are skipped: empty/query-only/anchor/
  `javascript:` URIs (case-insensitive), literal-scheme and protocol-relative
  URLs (knowably external — checked after query/fragment stripping, so a local
  `/login?next=https://app/` still pairs), whole-URL template expressions, and
  values still containing template syntax after normalization (control-flow and
  declaration actions). `data-hx-*` attributes are matched (official htmx
  form); nothing else hyphen-prefixed.
- **Template-aware path normalization, local to the extractor**
  (`normalizeHtmxPath`): whole-segment Go template **value** expressions
  (`/ui/parts/{{.P.ID}}/exp`) collapse to positional params so consumer IDs
  collide with provider route IDs through the existing shared normalizer —
  which is itself unchanged (byte-identical to before this PR modulo gofmt).
  Control actions (`{{if}}`/`{{range}}`/`{{end}}`…) and declaration actions
  (`{{$x := …}}`) render nothing and are rejected rather than mis-collapsed. A
  leading expression is treated as a path param, not a base-URL slot
  (conservative: pairs only with a provider declaring a first param).
- **Indexer wiring** (`internal/indexer/indexer.go`): registers HtmxExtractor for
  `html`, `gotmpl`, and `templ` in both full and incremental paths (single
  `byLang` registration site). Graph edges flow through the existing
  `commitContracts` — zero new edge code; `KindContract` nodes + `consumes`
  edges are automatic. (`htmldjango` deliberately omitted: plain-Django
  providers mint method-less `http::ANY::` IDs that cannot collide with
  verb-specific consumers, and the idiomatic `{% url %}` form is unresolvable
  statically — worth revisiting if method-less providers gain verb bridging.)
- **Tests**: extractor unit tests (verbs, quoting, comments, skip shapes, dedup,
  `data-` form); a collision-gate test pinning that htmx consumer IDs are
  byte-identical to provider route IDs built through the same normalizer, plus
  a matcher-level test asserting a provider leaves the orphan set when its
  htmx consumer exists; indexer wiring test.

## Known limitations

- `.gohtml` templates are not a registered language anywhere in gortex yet, and
  `.tpl` is claimed by the Helm extractor ahead of gotmpl — templates using
  those extensions are not scanned.
- Routes consumed only via browser navigation (`<a href>`), external REST
  clients, or JS-built URLs remain in the orphan list — static analysis cannot
  see those consumers.
- **Existing indexes require a full re-index** to pick up htmx contracts:
  contract extraction is change-driven and there is no extractor-set
  versioning, so upgrading alone changes nothing until templates are re-indexed.

## Testing

- [x] Focused suites green (`internal/contracts`, `internal/indexer` wiring);
      `go build ./...`, `go vet`, gofmt clean on touched files. Full
      `go test -race ./...`: all packages pass on this machine except five
      failures verified identical at the merge base (pre-existing /
      environment-dependent: pricing-table + githooks fixtures fail the same on
      `main`; `store_sqlite` race needs CI's 30m budget; one opencode test is
      load-flaky and passes isolated).
- [x] New tests added for new functionality (incl. RED-verified regression pins:
      disabling the query-only guard or the comment-strip fails the pinned
      fixture counts).
- [ ] Benchmarks — not perf-relevant (regex over template text; ~1s for a
      176-file repo inside a full index).

End-to-end dogfood on the `go-parts` repo (XDG-isolated scratch daemon, A/B
against the merge-base binary on the same checkout, re-run on the final HEAD of
this branch):

- orphan providers **48 → 26**; 27 htmx consumer contracts across 14 template
  files; 22 distinct providers matched exclusively by htmx consumers
- the issue's live example `hx-get="/ui/parts/{{.P.ID}}/exp"` now pairs:
  `find_usages` returns `consumes` edges from `row.html` / `row-created.html`
- exhaustive check on the final HEAD: every hx-referenced target (37 raw
  attribute values, normalized, method-aware) is absent from the orphan list;
  the 26 remaining orphans are the external REST surface plus static/health/
  catch-all/browser-navigation routes htmx attributes never touch

### DIFF
--- a/internal/contracts/contract.go
+++ b/internal/contracts/contract.go
@@ -107,6 +107,18 @@ var tplBasePrefix = regexp.MustCompile(`^/?\$\{[^}]+\}`)
 // so consumer paths align with provider route declarations.
 var tplInlineParam = regexp.MustCompile(`\$\{([^}]+)\}|\$([a-zA-Z_][a-zA-Z0-9_]*)`)
 
+// tplWholeSegment matches a path segment that is ENTIRELY a template
+// expression — Go {{...}}, Jinja {%...%}, or ERB <%...%>. Interpolated
+// request-path slots (e.g. /ui/parts/{{.P.ID}}/exp) normalise to path
+// parameters exactly like a provider's :id / {id} declaration. Segments
+// that only CONTAIN an expression (/items-{{.ID}}) are left literal.
+var tplWholeSegment = regexp.MustCompile(`^\{\{.*\}\}$|^\{%.*%\}$|^<%.*%>$`)
+
+// tplParamName is the synthetic name handed to the positional normaliser
+// for whole-segment template expressions. paramPatterns renumbers it to
+// {p1}, {p2}, ... alongside declared params, in path order.
+const tplParamName = "{tplparam}"
+
 // NormalizeHTTPPath converts path parameters from various frameworks into the
 // canonical {param} form.  Examples:
 //
@@ -148,6 +160,32 @@ func NormalizeHTTPPathWithParams(path string) (string, []string) {
 			path = "/"
 		}
 	}
+
+	// Whole-segment template expressions are interpolated path slots. A
+	// LEADING one followed by more segments is the base-URL slot
+	// ({{.Base}}/v1/users) — strip it like tplBasePrefix below strips
+	// ${BASE}. Any other whole-segment expression becomes a path
+	// parameter so it lands in the same {p1}, {p2}, ... space as the
+	// provider's declared params.
+	segs := strings.Split(path, "/")
+	first := 0
+	for first < len(segs) && segs[first] == "" {
+		first++
+	}
+	strippedBase := false
+	for i := first; i < len(segs); i++ {
+		if segs[i] == "" || !tplWholeSegment.MatchString(segs[i]) {
+			continue
+		}
+		if i == first && len(segs)-first > 1 && !strippedBase {
+			segs = append(segs[:i], segs[i+1:]...)
+			strippedBase = true
+			i--
+			continue
+		}
+		segs[i] = tplParamName
+	}
+	path = strings.Join(segs, "/")
 
 	// Strip a leading template-literal placeholder (with optional leading
 	// slash) — the base-URL slot that a consumer interpolates. After this
--- a/internal/contracts/contract.go
+++ b/internal/contracts/contract.go
@@ -18,8 +18,8 @@ const (
 	// patterns the gRPC extractor recognises (New<Service>Client), so
 	// the matcher's canonical-name join treats grpc and thrift as one
 	// RPC family when pairing.
-	ContractThrift  ContractType = "thrift"
-	ContractGraphQL ContractType = "graphql"
+	ContractThrift     ContractType = "thrift"
+	ContractGraphQL    ContractType = "graphql"
 	ContractTopic      ContractType = "topic"
 	ContractWS         ContractType = "ws"
 	ContractEnv        ContractType = "env"
--- a/internal/contracts/contract.go
+++ b/internal/contracts/contract.go
@@ -107,18 +107,6 @@ var tplBasePrefix = regexp.MustCompile(`^/?\$\{[^}]+\}`)
 // so consumer paths align with provider route declarations.
 var tplInlineParam = regexp.MustCompile(`\$\{([^}]+)\}|\$([a-zA-Z_][a-zA-Z0-9_]*)`)
 
-// tplWholeSegment matches a path segment that is ENTIRELY a template
-// expression — Go {{...}}, Jinja {%...%}, or ERB <%...%>. Interpolated
-// request-path slots (e.g. /ui/parts/{{.P.ID}}/exp) normalise to path
-// parameters exactly like a provider's :id / {id} declaration. Segments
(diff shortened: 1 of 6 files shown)


madeinoz67:
Looked into the failing `govulncheck` check — it does not appear to be caused by this PR:

- The single finding is **GO-2026-6115** in `github.com/ledongthuc/pdf` (the PDF text-extraction dep), reported as **Fixed in: N/A** — no fixed version exists upstream yet.
- `ledongthuc/pdf` is in `go.mod` at this PR's merge base, so the dependency predates the change; the advisory is just newer than main's last `security` workflow run (which passed at PR #604). Main's next scheduled security run should hit the same finding.
- The `tar: Cannot open: File exists` lines in the same job are module-cache restore noise, separate from the vuln verdict.

Options as I see them: add GO-2026-6115 to the security workflow's ignore list until upstream ships a fix, or swap the `pdf` dep when one lands. Happy to do either as a separate PR if useful.

zzet (approved):

zzet:
@madeinoz67 thank you for your contribution.

P.s. The advisory was withdrawn, and the code has protection against that issue.

PR #611, opened 2026-08-19
feat(daemon): surface daemon upgrade skew and staleness to CLI clients

### LINKED ISSUE #609: brew upgrade leaves the old daemon running: no restart, no version-skew detection (macOS)
**Context:** macOS, gortex installed via Homebrew. The formula replaces `/opt/homebrew/bin/gortex`, but a daemon that is already running keeps serving from the old binary indefinitely after `brew upgrade`. Nothing restarts it, and nothing surfaces the skew: the CLI and every MCP client keep connecting to the daemon on the old version.

**Observed on my machine (v0.63.x line):**
- a daemon started days before an upgrade was still serving afterwards; `gortex daemon status` showed the old build while the on-disk binary was newer
- the launchd unit exists (`~/Library/LaunchAgents/com.zzet.gortex.plist`) but reports `not loaded`, while a manually-started daemon holds the socket — so even a user who ran `daemon install-service` once can end up with a unit that owns nothing

**Why I think this happens:**
1. A hand-started daemon (`gortex daemon start`) has no supervisor and no upgrade hook; swapping the binary does not touch the running process. macOS keeps executing the old inode.
2. `brew upgrade` does not reload or restart launchd units. If the unit was never loaded (or the daemon was started manually anyway), there is nothing to bounce.
3. The docs say `gortex mcp` exits "if no compatible daemon can be reached" — but compatibility evidently is not binary-version-level, because a new CLI drives an old daemon silently. I could not find any build-version comparison on the connect path (the protocolVersion strings I found are MCP-protocol constants).

**Impact:** stale graph behavior and bugfixes; mixed versions between CLI and daemon (`daemon status` shows one build, the binary is another); store migrations and behavior changes land late or never until someone notices and restarts by hand.

**Possible directions (not prescriptions):**
1. Version-skew detection on connect: CLI/`mcp` proxy compares its build version against the daemon's (status already carries a version) and either refuses or prints a loud one-liner: daemon is vX, binary is vY, run `gortex daemon restart`.
2. Make the restart story explicit after package-manager upgrades: a hint in `brew` output is formula-side, but gortex could detect its own staleness (e.g. running image vs on-disk binary) and report it in `daemon status`/`mcp` logs.
3. Consider making the supervised service the blessed path (KeepAlive restarts after upgrade), and have `daemon start` warn when a launchd unit is installed but not loaded, or when a hand-started daemon is shadowing a unit.

Happy to work on a fix if any of the directions sound right.

madeinoz67:
## Summary

Closes #609. When the homebrew cask upgrades `gortex`, a long-running daemon keeps serving from the old binary and every CLI invocation silently talks to the version it was just upgraded past. `docs/versioning.md` already promises that &#34;the daemon exposes its running version in the handshake ACK as `DaemonVersion` and on the control surface&#39;s `status` response — clients can feature-gate or warn on mismatch&#34; — this PR implements the client half of that documented intent: the MCP proxy now warns on mismatch, and `daemon status` shows both the CLI-vs-daemon version skew and the daemon&#39;s own on-disk-binary drift.

Three surfaces, one compare each:

- **MCP proxy connect-time warning** (`cmd/gortex/proxy.go`): `daemonSkewWarning(daemonVer, localVer)` compares the handshake-reported daemon version against `canonicalVersion()` and prints one stderr line through the existing `[gortex mcp]` prefix — warn-and-continue, never gate. The remedy is direction-aware: daemon older ⇒ `run &#39;gortex daemon restart&#39; to upgrade the daemon` (a restart respawns from the newer on-disk binary); binary older ⇒ `run &#39;gortex upgrade&#39;` — it upgrades the way gortex was installed (brew, scoop, go install, or the install script) and restarts the daemon around the swap itself; unparseable/same-precedence-different-build ⇒ the generic restart-or-upgrade remedy. Dev builds stay silent on either side (`v0.0.0-dev` as the local OR the daemon identity) so plain `go build` runs — CLI or daemon — never nag or advise &#34;upgrading&#34; a dev process.
- **Daemon staleness self-report** (`internal/daemon/server.go`): the daemon captures its own executable&#39;s path/size/mtime at startup and stats it again on every `ControlStatus`, populating additive `StatusResponse` fields (`BinaryChecked`/`BinaryStale`/`BinaryReplacedAtUnix`). A stat failure or an uncaptured identity reports *unknown* (both flags false) — unknown is never rendered as fresh. The `on-disk binary changed` warning is logged exactly once per detection (once-flag), so polling `status --watch` cannot spam the daemon log. Detection covers both hand-started and launchd daemons: it derives from the kernel&#39;s view of the running image (`os.Executable`), not from any launchd integration.
- **Status rendering** (`cmd/gortex/daemon.go`): two conditional rows in the header table — `cli   (differs from daemon)` when the same skew compare fires, and `binary  stale — on-disk image newer than running image; run &#39;gortex daemon restart&#39;` when the daemon reports drift. Matching versions and fresh binaries keep the table exactly as terse as before.
- **Doctor** (`cmd/gortex/doctor.go`, review follow-up): the handshake row downgrades from ✓ to `!` and prints the same remedy line — computed by the same `daemonSkewWarning`, so doctor, `daemon status`, and the MCP proxy give one verdict. `DoctorEnvironment` carries `cli_version`/`version_skew_warning` for `--json` consumers.

## Changes

- `cmd/gortex/proxy.go` — `daemonSkewWarning` + connect-time stderr warning in the proxy path; direction-aware remedies (`gortex upgrade`, never package-manager-specific); dev-build silence.
- `internal/version/compare.go` (review follow-up) — `Compare` orders two Versions by SemVer 2.0.0 precedence (build metadata ignored); fills a package gap — internal/version parsed semver but could not order it. The §11 precedence chain is pinned in `compare_test.go`.
- `internal/daemon/server.go`, `internal/daemon/proto.go` — startup binary-identity capture, per-status drift probe, log-once warning, additive `StatusResponse` fields (no protocol-version bump; older clients ignore the new fields).
- `cmd/gortex/daemon.go` — conditional `cli` and `binary` rows in `renderDaemonHeader`; `cmd/gortex/daemon_status_render_test.go` pins both rows and their negative cases.
- `cmd/gortex/doctor.go` — skew-aware handshake row + `DoctorEnvironment` fields; `cmd/gortex/doctor_env_test.go` pins the ✓/`!` rows.

## Testing

- [x] All tests pass (`go test -race ./...`) — **140 ok / 0 data races**; six non-green packages, every one dispositioned against clean `main` at `b55b9a0d` (throwaway-worktree reruns) or by isolation re-run, none a regression from this PR: `TestFormatModelCostSeparatesUnpricedFromZero` (cmd/gortex) and `TestCostAvoidedAll_IncludesDefaults` (internal/savings) both fail identically on main (stale pricing-table rows); `TestUninstallHookPostCommit_RemovesFileWhenStubOnly` (internal/githooks) fails identically on main; `internal/agents/opencode` + `internal/gitcmd` are load flakes that pass on this branch in isolation and on main; `internal/graph/store_sqlite` timed out at 606s only under the pegged full-suite run; `TestRouter_ReloadConfig_ConcurrentNoRace` (internal/daemon) hangs deterministically on this host, identically on main (workers stuck in `ServerClient.ProxyToolCtx` HTTP `getConn`; skipped via `-skip`; needs its own issue).
- [x] New tests added for new functionality:
  - `TestDaemonSkewWarning` — 7 subtests: equal, daemon-older, daemon-newer, same-version-different-build, unparseable daemon version, empty daemon version, dev build.
  - `TestComparePrecedence` (internal/version) — the semver.org §11 chain (alpha &lt; alpha.1 &lt; alpha.beta &lt; beta &lt; beta.2 &lt; beta.11 &lt; rc.1 &lt; release) pinned pairwise against `Compare`, plus build-metadata-ignored and numeric-vs-alphanumeric rows.
  - `TestRenderDaemonHeader_SkewRow` — 3 subtests: skewed appends the `cli` row, matching omits it, dev build omits it.
  - `TestRenderDaemonHeader_BinaryRow` — 3 subtests: stale appends the `binary` row, fresh omits it, unchecked omits it even if `BinaryStale` is set.
  - `TestPrintDoctorEnvironment_VersionSkew` — matching versions keep the ✓ row; skewed versions render `!` plus the remedy line.
  - `TestStatusBinaryDriftFreshThenStale` / `_SizeChange` / `TestStatusBinaryStatErrorReportsUnknown` / `TestStatusBinaryUncapturedIdentityReportsUnknown` — full lifecycle through the real `ControlStatus` handler: fresh → mtime-drift stale → log-once pinned; size-only drift; stat-error ⇒ unknown; uncaptured identity ⇒ unknown.
  - Gates: `go build ./...` + `go vet ./...` clean; `gofmt` clean on all branch-touched files; `go test -race ./cmd/gortex/ ./internal/version/ ./internal/daemon/` green except the two pre-existing dispositioned failures above.
- [ ] Benchmarks — not performance-relevant (one string compare per proxy connect, one `stat` per status poll).

### E2E dogfood (XDG-isolated scratch daemon; binary swapped beside it, simulating a cask upgrade)

```console
$ go build -o $E2E/bin/gortexA ./cmd/gortex &amp;&amp; cp gortexA gortex &amp;&amp; gortex daemon start
$ gortex daemon status                     # baseline — versions equal, binary fresh
 daemon    v0.63.4
 pid       80988
 socket    /tmp/gortex-skew-e2e/xdg/cache/gortex/daemon.sock
 ...                                       # no cli row, no binary row  → PASS

$ go build -ldflags &#34;-X main.version=9.9.9&#34; -o gortexB ./cmd/gortex &amp;&amp; mv gortexB gortex
$ gortex daemon status                     # after the simulated cask upgrade
 daemon    v0.63.4
 pid       80988
 cli       v9.9.9 (differs from daemon)
 binary    stale — on-disk image newer than running image; run &#39;gortex daemon restart&#39;
 ...                                       # both rows present  → PASS

$ echo | gortex mcp 2&gt;&amp;1 &gt;/dev/null | head -5
[gortex mcp] proxying to daemon (session 97d174cdd5feb62e2ac45de54f274e40, default_repo=&#34;&#34;)
[gortex mcp] warning: daemon v0.63.4 != binary v9.9.9 — run &#39;gortex daemon restart&#39; to upgrade the daemon
                                            # skew warning on stderr, direction-aware remedy  → PASS
```

Teardown by PID kill (never `daemon stop`); the live homebrew daemon was healthy and untouched afterwards (`v0.63.4+5f5fce2`, 5 sessions, `ready`).

## Review follow-ups (#4969353348)

All three proposals addressed (352131a2, 205b2e2e, 668dfa47):

- remedy text names `gortex upgrade` — install-method-agnostic, and it bounces the daemon around the swap itself
- semver precedence compare moved to `internal/version.Compare`
- doctor surfaces the same skew verdict on the handshake row

## Checklist

- [ ] Language extractor includes `Meta[&#34;methods&#34;]` for interfaces (if applicable) — N/A
- [ ] Methods have `EdgeMemberOf` edges to their containing type (if applicable) — N/A

## Known limitations

- **Restart remains manual.** The remedy text tells the operator what to run; nothing auto-restarts the daemon. Automating it (restart-on-upgrade) needs the cask-side stop/restart hook, which is deliberately deferred to its own change per the #609 plan.
- **Dev builds are silent by design.** A local `v0.0.0-dev` CLI identity never triggers the warning or the `cli` row — the spec&#39;s &#34;degrade silently&#34; posture, extended to the dev-sentinel case so contributor builds don&#39;t nag.
- **Detection is per-status-poll, not push.** The staleness probe runs when `status` is requested; a daemon nobody polls still drifts silently. The proxy warning independently covers every MCP client connect.


cmd/gortex/proxy.go
```diff
@@ -140,6 +146,110 @@ func runProxy(ctx context.Context, surface *gortexmcp.ToolSurface) (ran bool, er
...
+		return base + " — run 'gortex daemon restart' or upgrade this binary"
+	}
+}
+
+// compareSemver orders two Versions by SemVer 2.0.0 precedence:
+// major.minor.patch numerically, then pre-release (absent outranks
+// present; identifiers compared per comparePrereleaseIdent). Build
+// metadata is ignored — per the spec it never affects precedence.
+// Returns -1, 0, or +1.
+func compareSemver(a, b semver.Version) int {
```
zzet: @madeinoz67 IMO the internal/version is a better place for this code.

cmd/gortex/proxy.go
```diff
@@ -140,6 +146,110 @@ func runProxy(ctx context.Context, surface *gortexmcp.ToolSurface) (ran bool, er
...
+	d, dErr := semver.Parse(daemonVer)
+	l, lErr := semver.Parse(localVer)
+	if dErr != nil || lErr != nil {
+		return base + " — run 'gortex daemon restart' or upgrade this binary"
+	}
+	switch compareSemver(d, l) {
+	case -1: // daemon older — restarting respawns it from this newer binary
+		return base + " — run 'gortex daemon restart' to upgrade the daemon"
+	case 1: // daemon newer — this binary is the stale side
+		return base + " — this binary is older than the running daemon — upgrade it (e.g. brew upgrade gortex)"
```
zzet: `upgrade it (e.g., brew upgrade gortex)` is only for macOS users. At the same time, Gortex ships its own `gortex upgrade` command and an install-script path (docs/installation.md), so a curl-installed user gets advice that won't work. `gortex upgrade` covers everyone.

cmd/gortex/proxy.go
```diff
@@ -140,6 +146,110 @@ func runProxy(ctx context.Context, surface *gortexmcp.ToolSurface) (ran bool, er
...
+// (a restart respawns it from this newer binary), while a newer daemon
+// means this binary is the stale side and upgrading it is the fix —
+// restarting would downgrade the daemon to this older build. When
+// either side does not parse as semver, or only the build metadata
+// differs (which SemVer precedence ignores), a generic remedy that
+// covers both directions is emitted. Implements the documented intent
+// in docs/versioning.md: the daemon exposes DaemonVersion so "clients
+// can feature-gate or warn on mismatch"; this warns and continues —
+// never gates.
+func daemonSkewWarning(daemonVer, localVer string) string {
```
zzet, line 167: Worth wiring this logic into the `gortex doctor` https://github.com/zzet/gortex/blob/main/cmd/gortex/doctor.go#L330-L335 

That's the command a confused user actually runs when versions feel wrong.

zzet (approved):
@madeinoz67 good work, thank you!

I found a non-blocking proposals, please let me know if you'd like to address them.

cmd/gortex/proxy.go
madeinoz67: Moved in 205b2e2e — `Compare` now lives in internal/version (which parsed semver but had no ordering, so it fills a real package gap), with the §11 precedence chain pinned alongside it as `TestComparePrecedence`. The remedy-text half (`daemonSkewWarning`) stays in cmd/gortex — it's CLI messaging with gortex-specific remedies rather than version logic.

cmd/gortex/proxy.go
madeinoz67: Fixed in 352131a2 — the daemon-newer remedy now says `run 'gortex upgrade'`, and the generic branches say `or 'gortex upgrade'`. It's even better than portable: `gortex upgrade` detects the install method and stops/starts the daemon around the swap itself. Kept `gortex daemon restart` for the daemon-older case, where the CLI is already the newer binary and `gortex upgrade` would report "already the latest".

cmd/gortex/proxy.go
madeinoz67: Wired in 668dfa47 — the handshake row downgrades from ✓ to `!` and prints the same remedy line, computed by the same `daemonSkewWarning` the proxy uses, so doctor, `daemon status`, and the MCP proxy can't disagree. `DoctorEnvironment` also carries `cli_version` and `version_skew_warning` for `--json` consumers; dev builds stay silent via the existing `v0.0.0-dev` sentinel.

PR #646, opened 2026-08-21
fix(review): join the changeset to native-separator graph paths

tiendungdev:
## Problem

`reviewRulepackMatches` builds the changed-file key correctly, then looks it up with `filepath.Clean` — which rewrites the one separator a graph path deliberately keeps as `/`: the repo prefix.

Probed on windows/amd64 with the existing fixture:

```
changed map key             "repo-a/pkg\widget.go"
target.GraphPath            "repo-a/pkg\widget.go"    <- identical
filepath.Clean(GraphPath)   "repo-a\pkg\widget.go"    <- prefix slash lost
hit                         false
```

The narrowing therefore produced an empty target set and `review` reported zero findings on code its own detector bundle flags under `analyze --kind review`. That is the same false clean `reviewChangedGraphPaths` was written to fix — its doc-comment says so — reintroduced one line later by the `Clean`.

## Change

Two halves of one contract, both in `tools_review.go`:

1. **The join.** Compare in `graphpath.Norm` form on both sides. `Norm` is the canonical comparison spelling for a path shaped `"<prefix>/"` + native remainder, and it is `filepath.ToSlash`, so this is the identity on POSIX.

2. **The output.** `reviewRepoRelPath` feeds the `.gortex.yaml` rule globs, `rankFileRisk`'s row keys, and the forge comment API — the function's own doc-comment lists them, and all three speak `/`. On Windows it returned `pkg\widget.go`, which matches no glob and anchors no comment. Normalize there too.

## Effect

Whole `internal/mcp` package on windows/amd64, go1.26.6, `-count=1`:

| | failures |
|---|---|
| `main` (`c982cf52`) | 27 |
| with this change | **24** |

Diffing failing test *names*: exactly three flip, newly-broken set **empty**.

```
FIXED:  TestReviewRulepackMatches_JoinsRepoRelativeChangedFiles
        TestReviewRulepackMatches_AcceptsAlreadyPrefixedChangedFiles
        TestReview_PrefixedGraphReportsRulepackFinding
NEWLY BROKEN: (none)
```

## Verification

**Each half sabotage-verified separately**, because a fix that only looks necessary is not:

- revert the join → all three fail;
- revert the output normalization → the two that assert `m.File` fail on `Not equal`.

`golangci-lint run ./internal/mcp/...` reports 6 `staticcheck` SA5011 findings; **pre-existing**, identical with this change stashed. `git diff --check` clean.

## Why there is no cross-platform test

`graphpath.Norm` is `filepath.ToSlash`, deliberately a **no-op on POSIX** — the package doc is explicit that a backslash is an ordinary filename byte there and must survive. So a test feeding a backslash path and asserting a slash result would fail on linux/macos, and one built with `filepath.FromSlash` asserts nothing on either. The three names therefore join the existing windows native-separator step, whose package list already carries `internal/mcp` — no new step, no new compile.

### Declared, not swept under

`TestReviewPackNeverClaimsNoTestSymbolsWithTestTargets` fails on windows too and I deliberately left it out: it fails on `test_targets` being empty, not on the path join, and `testpath.IsTestFile` already normalizes separators itself — so it is a different cause and belongs in its own change rather than riding along here.

Branched from `main` at `c982cf52`. Windows 11, go1.26.6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,7 @@ jobs:
       - name: Test native-separator store path comparisons
         run: >
           go test -timeout=10m -count=1
-          -run 'NativeSeparator|MixedSeparator|ImportAdjacency|ParseDiffGitPaths|ParseDiffLinesNewSide|FeedbackDir|ReviewRulepack|PrefixedGraphReportsRulepack|GraphKey|JoinFileNodes|ChangedSymbolsForFiles|PrefixShadow|IdsOnPrefixedGraph|RankFileRisk'
+          -run 'NativeSeparator|MixedSeparator|ImportAdjacency|ParseDiffGitPaths|ParseDiffLinesNewSide|FeedbackDir|ReviewRulepack|PrefixedGraphReportsRulepack|GraphKey|JoinFileNodes|ChangedSymbolsForFiles|PrefixShadow|IdsOnPrefixedGraph|RankFileRisk|PrefixedDeleteAndRename'
           ./internal/analysis ./internal/graph/store_sqlite
           ./internal/mcp ./internal/resolver ./internal/persistence
           ./internal/review
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,12 +108,20 @@ jobs:
       # filesystem path (filepath.Join / filepath.Clean) or a '/'-keyed store
       # path. ParseDiffGitPaths, ParseDiffLinesNewSide and FeedbackDir all
       # assert on native paths, so they belong here rather than nowhere.
+      # The path-domain contracts belong here for the same reason. A graph
+      # key is "<prefix>/" plus a native-separator remainder, so GraphKey,
+      # JoinFileNodes, the prefix-shadow end-to-end pair and RankFileRisk's
+      # per-domain normalization only diverge once '/' and the native
+      # separator differ — every one of them passes on the linux/macos
+      # matrix whatever the code does. internal/review joins the list
+      # because the file-risk normalizer lives there.
       - name: Test native-separator store path comparisons
         run: >
           go test -timeout=10m -count=1
-          -run 'NativeSeparator|MixedSeparator|ImportAdjacency|ParseDiffGitPaths|ParseDiffLinesNewSide|FeedbackDir|ReviewRulepack|PrefixedGraphReportsRulepack'
+          -run 'NativeSeparator|MixedSeparator|ImportAdjacency|ParseDiffGitPaths|ParseDiffLinesNewSide|FeedbackDir|ReviewRulepack|PrefixedGraphReportsRulepack|GraphKey|JoinFileNodes|ChangedSymbolsForFiles|PrefixShadow|IdsOnPrefixedGraph|RankFileRisk'
           ./internal/analysis ./internal/graph/store_sqlite
           ./internal/mcp ./internal/resolver ./internal/persistence
+          ./internal/review
 
   build-linux-static:
     # The linux release ships a statically linked binary so it runs on any
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
       - name: Test native-separator store path comparisons
         run: >
           go test -timeout=10m -count=1
-          -run 'NativeSeparator|MixedSeparator|ImportAdjacency|ParseDiffGitPaths|ParseDiffLinesNewSide|FeedbackDir'
+          -run 'NativeSeparator|MixedSeparator|ImportAdjacency|ParseDiffGitPaths|ParseDiffLinesNewSide|FeedbackDir|ReviewRulepack|PrefixedGraphReportsRulepack'
           ./internal/analysis ./internal/graph/store_sqlite
           ./internal/mcp ./internal/resolver ./internal/persistence
 
--- a/internal/analysis/diffmap.go
+++ b/internal/analysis/diffmap.go
@@ -122,15 +122,32 @@ const (
 // unconditionally — never conditionally on how it happens to be spelled — so a
 // path that merely looks prefixed cannot be resolved against the wrong file.
 func GraphKey(repoPrefix, path string, domain PathDomain) string {
-	if domain == GraphKeyedPath || repoPrefix == "" {
+	if domain == GraphKeyedPath {
 		return path
 	}
-	// The key is "<prefix>/" + the remainder in the indexing machine's native
-	// separators (see internal/graphpath), so a git path — always '/'-spelled —
-	// is converted on the way in. FromSlash is the identity on POSIX; on
-	// Windows the stored key reads `repo/dirile.go`, and a '/'-joined key
-	// misses every file below the repo root.
-	return repoPrefix + "/" + filepath.FromSlash(path)
+	// A repo-relative path is always converted, prefix or not. The key's shape
+	// is "<prefix>/" + the remainder in the indexing machine's native
+	// separators (see internal/graphpath); with no prefix the remainder IS the
+	// key, and a '/'-spelled git or forge path still misses a key stored with
+	// native separators on Windows. FromSlash is the identity on POSIX.
+	key := filepath.FromSlash(path)
+	if repoPrefix == "" {
+		return key
+	}
+	return repoPrefix + "/" + key
+}
+
+// RepoRelPath is GraphKey's inverse: the repo-relative '/' spelling that
+// CODEOWNERS rules, forge comment APIs and report rows speak. Stripping the
+// prefix is safe here and only here, because a GraphKeyedPath carries it by
+// construction — the same strip applied to a repo-relative path is the
+// prefix-shadow bug.
+func RepoRelPath(repoPrefix, path string, domain PathDomain) string {
+	rel := filepath.ToSlash(path)
+	if domain == GraphKeyedPath && repoPrefix != "" {
+		rel = strings.TrimPrefix(rel, repoPrefix+"/")
+	}
+	return rel
 }
 
 // JoinFileNodes maps one changed-file path to the graph nodes its file
--- a/internal/analysis/diffmap.go
+++ b/internal/analysis/diffmap.go
@@ -101,36 +101,48 @@ func MapGitDiff(g graph.Store, repoRoot, repoPrefix, scope, baseRef string) (*Di
 	return joinHunksToSymbols(g, repoPrefix, hunks, files), nil
 }
 
-// JoinFileNodes maps one repo-relative changed-file path to the graph nodes
-// its file defines. Indexed file paths carry the repo prefix in multi-repo
-// mode ("<prefix>/<rel>") while git and forge APIs emit repo-relative paths,
-// so the raw lookup is tried first (single-repo / already-prefixed input)
-// and the prefixed form second.
-func JoinFileNodes(g graph.Store, repoPrefix, path string) []*graph.Node {
-	if nodes := g.GetFileNodes(path); len(nodes) > 0 {
-		return nodes
+// PathDomain names the vocabulary a file path is spelled in. The two overlap,
+// so the domain has to travel with the path rather than be recovered from it:
+// in a repo whose tree carries a top-level directory named like the repo
+// prefix, "repo-a/pkg/widget.go" is a well-formed path in BOTH domains — a
+// git-relative path naming the nested file, and the graph key of the entirely
+// different top-level "pkg/widget.go".
+type PathDomain int
+
+const (
+	// RepoRelativePath is git's and a forge's spelling, relative to the work
+	// tree. Every changed-file list in this package is this domain.
+	RepoRelativePath PathDomain = iota
+	// GraphKeyedPath is the graph's own node key: "<prefix>/<rel>" in
+	// multi-repo mode, the bare relative path when no prefix applies.
+	GraphKeyedPath
+)
+
+// GraphKey renders path as the graph keys it. A repo-relative path is prefixed
+// unconditionally — never conditionally on how it happens to be spelled — so a
+// path that merely looks prefixed cannot be resolved against the wrong file.
+func GraphKey(repoPrefix, path string, domain PathDomain) string {
+	if domain == GraphKeyedPath || repoPrefix == "" {
+		return path
 	}
-	if repoPrefix == "" || strings.HasPrefix(path, repoPrefix+"/") {
-		return nil
-	}
-	return g.GetFileNodes(repoPrefix + "/" + path)
+	// The key is "<prefix>/" + the remainder in the indexing machine's native
+	// separators (see internal/graphpath), so a git path — always '/'-spelled —
+	// is converted on the way in. FromSlash is the identity on POSIX; on
+	// Windows the stored key reads `repo/dirile.go`, and a '/'-joined key
+	// misses every file below the repo root.
+	return repoPrefix + "/" + filepath.FromSlash(path)
 }
 
-// JoinFilePath returns the graph-keyed variant of a repo-relative file path:
-// the raw path when it resolves (or no prefix applies), otherwise the prefixed
-// form when that resolves. Falls back to the raw path when neither does, so
-// the caller's downstream lookup misses exactly as it would have anyway.
-func JoinFilePath(g graph.Store, repoPrefix, path string) string {
-	if repoPrefix == "" || strings.HasPrefix(path, repoPrefix+"/") {
-		return path
-	}
-	if len(g.GetFileNodes(path)) > 0 {
-		return path
-	}
-	if prefixed := repoPrefix + "/" + path; len(g.GetFileNodes(prefixed)) > 0 {
-		return prefixed
-	}
-	return path
+// JoinFileNodes maps one changed-file path to the graph nodes its file
+// defines. domain states which vocabulary path is in; see PathDomain for why
+// it cannot be inferred.
+//
+// The previous contract tried the raw key first and the prefixed key second,
+// which resolved a legitimate git-relative "<prefix>/<rel>" against the
+// same-named top-level file instead — and returned nil when that shadow did
+// not exist, never trying the real key.
+func JoinFileNodes(g graph.Store, repoPrefix, path string, domain PathDomain) []*graph.Node {
+	return g.GetFileNodes(GraphKey(repoPrefix, path, domain))
 }
 
 // joinHunksToSymbols builds the hunk→symbol/file join shared by MapGitDiff
@@ -162,7 +174,7 @@ func joinHunksToSymbols(g graph.Store, repoPrefix string, hunks []DiffHunk, file
 		fileSet[hunk.FilePath] = true
 
 		// Find symbols whose line range overlaps the hunk
-		for _, n := range JoinFileNodes(g, repoPrefix, hunk.FilePath) {
+		for _, n := range JoinFileNodes(g, repoPrefix, hunk.FilePath, RepoRelativePath) {
 			// Check if symbol's line range overlaps with the hunk
 			if n.StartLine <= hunk.EndLine && n.EndLine >= hunk.StartLine {
 				addSymbol(n)
@@ -187,7 +199,7 @@ func joinHunksToSymbols(g graph.Store, repoPrefix string, hunks []DiffHunk, file
 			continue
 		}
 		fileSet[vanished] = true
-		for _, n := range JoinFileNodes(g, repoPrefix, vanished) {
+		for _, n := range JoinFileNodes(g, repoPrefix, vanished, RepoRelativePath) {
 			addSymbol(n)
 		}
 	}
--- a/internal/analysis/diffmap_test.go
+++ b/internal/analysis/diffmap_test.go
@@ -328,12 +328,21 @@ func TestJoinFileNodes(t *testing.T) {
 	if nodes := JoinFileNodes(g, "myrepo", key("a.go"), GraphKeyedPath); len(nodes) != 1 || nodes[0].ID != key("a.go")+"::A" {
 		t.Fatalf("graph-keyed path must be used verbatim: %#v", nodes)
 	}
-	// No prefix (standalone indexer): the path already is the key.
-	if nodes := JoinFileNodes(g, "", key("a.go"), RepoRelativePath); len(nodes) != 1 || nodes[0].ID != key("a.go")+"::A" {
-		t.Fatalf("empty prefix must not rewrite the key: %#v", nodes)
+	// A graph key is used verbatim whether or not a prefix is supplied.
+	if nodes := JoinFileNodes(g, "", key("a.go"), GraphKeyedPath); len(nodes) != 1 || nodes[0].ID != key("a.go")+"::A" {
+		t.Fatalf("a graph key must be used verbatim with no prefix: %#v", nodes)
 	}
 	if nodes := JoinFileNodes(g, "myrepo", "missing.go", RepoRelativePath); len(nodes) != 0 {
 		t.Fatalf("a miss must stay a miss: %#v", nodes)
+	}
+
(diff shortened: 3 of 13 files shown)


zzet (changes requested):
Requesting changes after an additional path-provenance validation.

Blocking correctness issue at internal/mcp/tools_review.go:700:

With repoPrefix repo-a and a legitimate Git-relative changed path repo-a/pkg/widget.go, normalization produces the same string. The HasPrefix check then treats it as already graph-prefixed, although the correct graph key is repo-a/repo-a/pkg/widget.go. This misses the changed target and, when pkg/widget.go also exists, can scan that unchanged shadow target instead. The ambiguity already exists on POSIX and this change newly exposes the wrong-target behavior on Windows.

Please address the following:

1. Carry explicit path provenance for Git-relative versus graph-keyed paths instead of inferring it with HasPrefix.
2. Prefix production MapGitDiff paths unconditionally, or otherwise make the input domain explicit.
3. Add a regression fixture containing both pkg/widget.go and repo-a/pkg/widget.go, mark only the nested path changed, and assert that only the changed target is scanned.
4. Rerun the focused Windows native-separator validation and the existing CI suite.

Adding both possible graph keys is not a safe workaround because it can still select unchanged code.

tiendungdev:
You were right, and the failure mode reproduces exactly as you described. Pushed `fdcf19d2`.

I verified the claim before changing anything: restoring the `HasPrefix` inference against the new fixture fails with

```
--- FAIL: TestReviewRulepackMatches_PrefixShadowedPathScansOnlyTheChangedTarget
    Messages: only the changed target may be scanned;
              "pkg/widget.go" is the unchanged shadow
```

— the changed `repo-a/pkg/widget.go` never scanned, the unchanged shadow scanned in its place.

### 1 + 2 — explicit provenance, unconditional prefixing

The domain now travels with the data instead of being recovered from it:

```go
changedPathsRepoRelative  // git's spelling, relative to the working tree
changedPathsGraphKeyed    // the graph's "<prefix>/<rel>" node key
```

`reviewRulepackMatches` takes it and passes it through; `reviewChangedGraphPaths` prefixes unconditionally when the domain is repo-relative and never inspects the string.

On "prefix production MapGitDiff paths unconditionally" — that is now what happens, and the contract already supported it. All four production callers (`tools_review.go` ×2, `tools_critique_review.go`, `tools_review_post.go`) pass `DiffResult.ChangedFiles`, and `joinHunksToSymbols` documents that field as keeping *"the diff-relative paths (callers re-join them with git pathspecs); only the node lookup is prefix-aware"*. So the repo-relative domain is the production domain by contract, not by assumption. `changedPathsGraphKeyed` is now reached only by the test that covers a caller holding node keys — I kept the capability rather than deleting it, but it is opt-in and named.

### 3 — regression fixture

`TestReviewRulepackMatches_PrefixShadowedPathScansOnlyTheChangedTarget`: the repo tree carries both `pkg/widget.go` and `repo-a/pkg/widget.go`, only the nested path is marked changed, and every match must report `repo-a/pkg/widget.go`.

Both files carry the detector fixture deliberately — otherwise a wrong-target scan would return empty and the test would pass for the wrong reason. With both seeded, the wrong target still produces matches and only the reported path separates the outcomes.

### 4 — validation

- Focused Windows native-separator selector: `ok` for `analysis`, `store_sqlite`, `mcp`, `resolver`, `persistence`.
- Whole `internal/mcp` on windows/amd64: 27 → 24 against `main`, failing test **names** diffed — same three fixed, newly-broken set empty, and the new test passes.
- `golangci-lint ./internal/mcp/...`: 6 `staticcheck` findings, all pre-existing (identical with the change stashed).

### One thing I removed

The `ci.yml` selector hunk is gone from this branch. #652 deletes that job outright, so leaving it would only hand you a conflict — and your point on #647 about selector lists applies here too.

Also worth flagging since it is upstream of this: the ambiguity is in the *data model*, not only in this function. `DiffResult.ChangedFiles` is documented as diff-relative but is plain `[]string`, so the next caller can make the same inference. If you want that closed properly the type belongs on the field rather than at this call site — happy to do it as its own change if you think it is worth the churn.

zzet (changes requested):
Second review round — requesting changes.

The rulepack narrowing fix itself is correct, but the prefix-shadow case is still incorrect end to end.

1. ChangedSymbols still bind to the unchanged shadow.

MapGitDiff passes Git-relative hunk paths to analysis.JoinFileNodes (internal/analysis/diffmap.go:109-117). JoinFileNodes tries the raw path first and returns it before trying the prefixed key. With repoPrefix repo-a and changed Git path repo-a/pkg/widget.go, the raw key repo-a/pkg/widget.go belongs to the unchanged pkg/widget.go shadow; the real changed graph key is repo-a/repo-a/pkg/widget.go. If the shadow exists it is returned, and if it does not exist the HasPrefix branch returns nil. The correct key is never tried.

That contaminates diff.ChangedSymbols. review and review_pack then derive impact, contracts, guards, test targets, classification, previews, receipts, and potentially the verdict from the unchanged file or from no symbols at all.

2. File-risk normalization repeats the domain collision.

internal/review/report.go:123-129 uses one normalizer that strips repoPrefix from graph-keyed ChangedSymbol.FilePath values and from already repo-relative findings and ChangedFiles. In the same case, the legitimate repo-relative repo-a/pkg/widget.go becomes pkg/widget.go. This attributes risk to the shadow path; after fixing JoinFileNodes it can also produce two inconsistent risk rows and incorrect finding counts/verdict input.

The new TestReviewRulepackMatches_PrefixShadowedPathScansOnlyTheChangedTarget directly calls reviewRulepackMatches, so it bypasses MapGitDiff, JoinFileNodes, rankFileRisk, and review_pack gates. The existing end-to-end test has no prefix-shadow fixture. Green CI therefore does not exercise either failure.

Please address before merge:

- Carry explicit repo-relative provenance through MapGitDiff/joinHunksToSymbols and prefix that domain unconditionally when resolving graph nodes. Audit JoinFilePath and the other JoinFileNodes callers rather than changing its mixed-domain contract implicitly.
- Normalize file-risk inputs by domain: strip the graph prefix only from graph-keyed symbol paths, not from repo-relative findings or ChangedFiles.
- Add end-to-end review and review_pack tests containing both pkg/widget.go and repo-a/pkg/widget.go, changing only the nested file. Assert the finding, ChangedSymbols, FileRisk, contracts/guards/test targets, previews, receipt, and verdict all refer only to repo-a/pkg/widget.go. Include deleted/renamed old-side paths if the join remains shared with those flows.

Non-blocking cleanup: changedPathsGraphKeyed currently has no production caller; only its test reaches that branch.

tiendungdev:
Both are real and both are fixed. Pushed `a7ca67b0`.

### 1 — `JoinFileNodes` resolved the wrong file

Confirmed exactly as you described, and the audit you asked for came back clean: **all five production call sites are repo-relative**, so nothing depended on the mixed contract.

| call site | source |
|---|---|
| `joinHunksToSymbols` — hunk paths | diff parsing |
| `joinHunksToSymbols` — `vanished` | `FileChange.Path` / `PreviousPath` |
| `changedSymbolsForFiles` (`tools_prs.go`) | forge file list |
| `suggest_reviewers` ×2 | its own comment already said *"the changed-file paths are repo-relative (git / forge)"* |

The domain now travels as `analysis.PathDomain` and a repo-relative path is prefixed unconditionally. `JoinFilePath` collapsed into `GraphKey` — with the domain explicit it was the same function, and probing the store to choose between two candidate keys is what made it ambiguous in the first place.

**One thing I found while building the fixture that is worth flagging:** `GraphKey` also has to convert separators. The key is `"<prefix>/"` + the remainder as the indexing machine spells it, so a `/`-spelled git path misses everything below the repo root on Windows. Probed on the real fixture:

```
GetFileNodes("repo-a/repo-a/pkg/widget.go")  -> 0 nodes
GetFileNodes("repo-a/repo-a\pkg\widget.go")  -> 4 nodes   <- what the indexer wrote
```

Without that, `ChangedSymbols` came back empty rather than wrong — a different failure, same root.

### 2 — file-risk normalization

Split by domain: `fromGraphKey` strips the prefix (ChangedSymbols), `fromRepoRel` does not (findings, ChangedFiles). `TestRankFileRiskNormalizesRepoPrefix` still merges the two spellings onto one row, which was the behaviour that normalizer existed for.

### 3 — end-to-end tests

`TestReview_PrefixShadowAttributesOnlyTheChangedFile` and `TestReviewPack_PrefixShadowAttributesOnlyTheChangedFile` drive the real pipeline — `MapGitDiff` → `JoinFileNodes` → `rankFileRisk` → the pack gates — over a repo whose tree carries a top-level directory named like the prefix. Both files carry the flagged source at the base commit and only the nested one is edited, **inside the function** so the hunk overlaps a symbol, so a wrong-target join still yields findings and only the attributed path separates the outcomes. They assert findings, `ChangedSymbols` (against the real graph key, not a `/`-joined guess), `FileRisk` and verdict.

Two fixture decisions worth stating, because both were mistakes I made first:

- **Flat paths, not nested.** With `pkg/widget.go` the changed git path and the shadow's graph key differ by separator on Windows, so the raw-lookup defect is masked here. One segment makes them coincide.
- **The edit is inside `Load`.** My first version touched line 1; the file changed, no symbol did, and `ChangedSymbols` was empty for a reason unrelated to the join.

**Limitation I would rather state than have you find:** on Windows I cannot sabotage-verify the e2e pair — `hunk.FilePath` is `filepath.Clean`ed to `repo-a\widget.go`, which never matches a `/`-joined key, so restoring raw-first is masked locally. The linux/macos shards are where they bind. What I *can* verify here is the unit level: restoring raw-first fails `TestJoinFileNodes` with *"prefix-shadowed repo-relative path must resolve to the nested file"*, and that fixture is platform-independent.

### Existing tests

Rewritten, not deleted. `TestJoinFileNodes` and `TestChangedSymbolsForFiles_RepoPrefixJoin` asserted the old "already-prefixed hits raw" tolerance — the defect itself — so they now assert the shadow resolves to the nested file. Their fixtures also build graph keys the way the indexer does instead of hard-coding the `/`-joined form, which is a POSIX-only spelling.

### Validation

- `internal/mcp` on windows: **27 → 24** against `main`, failing test *names* diffed, newly-broken set **empty**.
- `internal/review`: 7 failures, **unchanged and pre-existing** — byte-identical with this change stashed. They are `cleanPath`/`filepath.Clean` path-spelling assertions, not this change.
- Focused native-separator selector: `ok` across `analysis`, `store_sqlite`, `mcp`, `resolver`, `persistence`.
- `golangci-lint`: 6 staticcheck findings, also pre-existing.

### Non-blocking cleanup

`changedPathsGraphKeyed` is gone — folded into `analysis.PathDomain` so one vocabulary spans both packages, and `GraphKeyedPath` now has a real consumer in `JoinFileNodes`' contract.

tiendungdev:
CI is red but not on this change. It needs a re-run, which I cannot trigger from a fork.

**The failure is a wall-clock ratio assertion in a package this PR does not touch:**

```
--- FAIL: TestNamedChildrenLinearScaling (7.17s)
    iterator: 10.292818ms -> 246.449711ms  (23.9x growth — O(N) ~10x)
    naive   : 25.70379ms  -> 150.625358ms  (5.9x growth)
    iterator growth 23.9x over a 10x width increase — not linear (want < 20x)
FAIL github.com/zzet/gortex/internal/semantic/tstypes
```

`git diff --name-only origin/main...HEAD` touches **0 files** under `internal/semantic/`. The diff is `internal/analysis`, `internal/review` and `internal/mcp` only, and nothing in it reaches tree-sitter node iteration.

**The run's own numbers show the measurement was unstable.** `walkIndex` is the loop whose per-op cost is supposed to grow with N — the benchmark contrasts it with the iterator for exactly that reason — yet it grew only **5.9x** over a 10x width increase in that run, while the iterator grew 23.9x. Both cannot be true of the same machine behaving normally. The ratio also has a ~10 ms denominator measured over 50 runs against a ~250 ms numerator measured over 15, so scheduler noise on a shared runner lands almost entirely in the numerator.

Locally on windows/amd64, go1.26.6, `-count=3` on this branch:

| run | iterator | naive |
|---|---|---|
| 1 | **9.0x** | 11.1x |
| 2 | **9.8x** | 12.0x |
| 3 | **9.2x** | 12.1x |

Three passes, right at the expected ~10x, and the naive walk consistently grows *more* than the iterator — the opposite of what CI recorded.

**Two notes on reading that run**, since one of them cost me a detour:

- The ubuntu shard's log is not evidence of anything. It shows `##[error]The operation was canceled` with an orphaned `store_sqlite.test` and stops after 65 packages — that is `fail-fast` cancelling it once macOS failed, not a hang. Worth knowing while #652 is in flight: `fail-fast: false` there would have had both shards report, and the Windows shard's first enumeration will be much easier to read for the same reason.
- The previous commit on this branch (`fdcf19d2`) passed the same matrix, and the only code added since is the path-domain work plus tests.

If you would rather not re-run on faith, I am happy to send the scaling test's stabilisation as its own PR — the honest fix is to compare against a same-machine baseline rather than an absolute ratio, or to raise `iterLarge`'s run count so both sides average over comparable work. I have deliberately not touched it here.

zzet (changes requested):
Third review round — requesting changes.

The previous prefix-shadow fixes are correct in the main review paths, but two blocking correctness issues remain at head `a7ca67b`.

1. **High — `suggest_reviewers(ids=…)` double-prefixes graph-keyed paths.**

   `resolveReviewerChangeset`'s `ids` branch returns `n.FilePath` verbatim (`internal/mcp/tools_suggest_reviewers.go:142-153`), so those entries are already graph-keyed. The new ownership and co-change lookups at lines 100 and 111 nevertheless force `analysis.RepoRelativePath`.

   With prefix `repo-a`, a node path `repo-a/pkg/a.go` is looked up as `repo-a/repo-a/pkg/a.go`. Recent-author and co-change reviewer signals disappear. The existing IDs test uses an unprefixed graph, so it cannot expose this regression.

   Please either normalize the IDs branch to repo-relative paths or return/carry `PathDomain` with the files. Add a prefixed-graph IDs test that verifies CODEOWNERS, recent-author, and co-change signals.

2. **High — Windows file-risk normalization still produces duplicate/wrong rows.**

   In `internal/review/report.go:131-138`, `fromGraphKey` calls `cleanPath` before `strings.TrimPrefix(file, repoPrefix+"/")`. `cleanPath` uses `filepath.Clean`. On Windows:

   ```text
   repo-a/repo-a\widget.go
     -> repo-a\repo-a\widget.go
   ```

   The forward-slash prefix no longer matches. Changed-symbol impact/coverage stays on a graph-prefixed row while findings and `ChangedFiles` create a second repo-relative row.

   Normalize both domains to slash/`graphpath.Norm` form before removing the normalized prefix, and return the documented repo-relative `/` spelling. Strengthen both prefix-shadow end-to-end tests to require exactly one `FileRisk` row equal to the changed file; the current `NotEqual(shadow)` assertion permits both erroneous rows.

3. **Windows CI does not exercise these contracts.**

   The selector in `.github/workflows/ci.yml:101-106` does not run `TestGraphKey`, `TestJoinFileNodes`, `TestChangedSymbolsForFiles_RepoPrefixJoin`, or either new prefix-shadow end-to-end test. It also omits `./internal/review`, where the second blocker lives. Please extend the Windows coverage accordingly.

Non-blocking: `GraphKeyedPath` currently has only test callers; fixing the IDs path should give it a real production use. No new security/authentication/secret issue was found.

tiendungdev:
Both blockers confirmed and fixed, plus the coverage gap. Pushed `a1a5b87f`.

### 1 — `suggest_reviewers(ids=…)`

You were right, and it was my miss. `resolveReviewerChangeset`'s three sources do **not** share a vocabulary:

| source | domain |
|---|---|
| `ids` | graph node `FilePath` — **graph-keyed** |
| `base` | `diff.ChangedFiles` — repo-relative |
| `number` | forge file list — repo-relative |

My round-two audit read the call site's comment (*"the changed-file paths are repo-relative (git / forge)"*) and stopped there. That comment is true for two of the three producers, and I never walked the third.

The resolver now returns the domain with the files and both lookups use it. That also gives `GraphKeyedPath` its first production caller, which closes your non-blocking note.

New `TestSuggestReviewers_IdsOnPrefixedGraphKeepsAllSignals` builds a prefixed graph and asserts CODEOWNERS, recent-author and co-change all survive. **Sabotage-verified** — forcing the ids branch back to `RepoRelativePath`:

```
recent-author signal missing … [{Reviewer:org/secteam Kind:team Reasons:[CODEOWNERS for 1 file]}]
```

Only CODEOWNERS survives, exactly as you predicted — and exactly why the tool looked healthy.

### 2 — file-risk normalization

`fromGraphKey` called `cleanPath` (which ends in `filepath.Clean`) before the strip. Both domains now go through `graphpath.Norm` first and leave in the documented repo-relative `/` spelling.

**Sabotage-verified** with the strengthened assertion — restoring `cleanPath`-before-strip:

```
[{repo-a\repo-a\widget.go MEDIUM 0} {repo-a/widget.go LOW 1}]
  should have 1 item(s), but has 2
```

Two rows for one file: impact/coverage on the graph-prefixed row, the finding on the repo-relative one. Your point about `NotEqual(shadow)` was the important half — the old assertion passed on that output.

**Bonus:** this also fixes three pre-existing Windows failures in `internal/review` — `TestRankFileRiskUsesImpact`, `TestRankFileRiskNormalizesRepoPrefix`, `TestRankFileRiskExemptsTestFilesFromCoverageDebt`. The package goes **7 → 4**; the remaining four (`TestVerificationCommand`, `TestRunRelocatesAndBlocks`, both `TestRenderSummary_*Golden`) are untouched by this change and were already red.

### 3 — Windows coverage

Selector gains `GraphKey|JoinFileNodes|ChangedSymbolsForFiles|PrefixShadow|IdsOnPrefixedGraph|RankFileRisk`, and `./internal/review` joins the package list.

**A correction to what I told you last round:** I said the `ci.yml` selector hunk was dropped from this branch. It was not — it was still there from the first revision, which is what you were reading at `ci.yml:101-106`. I misreported that; sorry.

### The audit you asked for, done properly this time

Rather than trusting one comment again, I walked every producer for every consumer of the domain-carrying API:

- `changedSymbolsForFiles` is fed **only** by `resolvePRFiles` and the forge fetch in `prImpactForNumber` — four call sites, all forge PR file lists, all repo-relative. Its hardcoded `RepoRelativePath` is correct.
- The only two remaining `TrimPrefix(…, repoPrefix+"/")` sites are `report.go`'s `fromGraphKey` and `reviewRepoRelPath`; both normalize first and both have an established domain.

### Validation

- Exact Windows selector, all six packages: `ok`.
- `internal/mcp` whole package: **27 → 24** against `main`, failing test names diffed, newly-broken set **empty**.
- `internal/analysis`: `ok`. `internal/review`: 4, down from 7.
- `golangci-lint`: 6 staticcheck findings, all pre-existing.

One environment note on my numbers: Go 1.27 landed on this machine mid-session (for #654), so I pinned every figure above with `GOTOOLCHAIN=go1.26.6` to match this branch's `go.mod` and your CI, rather than reporting results from a toolchain CI does not run.

zzet (changes requested):
Fourth review round — requesting changes at head a1a5b87.

The previous IDs-mode double-prefixing and file-risk normalization blockers are fixed. Two path-domain correctness gaps remain:

1. High — internal/analysis/diffmap.go:113: GraphKey returns immediately when repoPrefix is empty, before filepath.FromSlash. On standalone Windows, forge paths such as pkg/auth/x.go remain slash-spelled and miss graph keys stored as pkg\auth\x.go. This silently drops changed symbols and reviewer signals in get_pr_impact and suggest_reviewers(number=...). Only GraphKeyedPath should return unchanged; RepoRelativePath should always pass through filepath.FromSlash and then conditionally receive the prefix. Add a multi-segment empty-prefix Windows regression test; the current a.go fixture cannot expose this.

2. Medium — internal/mcp/tools_suggest_reviewers.go:82: IDs mode returns graph-keyed paths, but CODEOWNERS still receives that spelling through relForRepo, which only strips an absolute repo root. A path such as repo-a/pkg/auth/login.go is passed instead of pkg/auth/login.go, so root-anchored rules such as /pkg/auth/ do not match. Convert the path to repo-relative form according to fileDomain before MatchFile and matched_files, and cover it with a root-anchored CODEOWNERS rule. The current unanchored pkg/auth/ fixture masks the mismatch.

Test gap: the deleted-file and renamed PreviousPath joins still have no prefixed/prefix-shadow regression cases. Their existing tests use an empty prefix, so they do not bind the new RepoRelativePath contract. Add prefixed delete and rename cases that resolve only the nested old-side graph key and include them in the Windows selector.

The first two behaviors also existed on the base branch, but they are directly inside the GraphKey/PathDomain contract introduced and exercised by this PR, whose purpose is cross-platform native-separator correctness. Current CI is green because the fixtures do not cover these cases. No new security or dead-code concern was found.

tiendungdev:
Both blockers and the test gap are addressed. Pushed `49b7e5ec`.

### 1 — `GraphKey` with an empty prefix

Confirmed. The guard conflated two conditions:

```go
if domain == GraphKeyedPath || repoPrefix == "" { return path }   // before
```

so a standalone index — no prefix, but still native keys — got the `/`-spelled path back unconverted. Now only `GraphKeyedPath` returns unchanged; `RepoRelativePath` always goes through `FromSlash` and then takes the prefix if there is one.

Regression: `TestGraphKey` and `TestJoinFileNodes` gain empty-prefix **multi-segment** cases. You were right that the `a.go` fixture cannot expose this — with no separator in the path, both spellings are identical and the assertion is vacuous.

### 2 — CODEOWNERS in ids mode

Confirmed, including `matched_files`. Added `analysis.RepoRelPath` as `GraphKey`'s inverse and routed both through it. It strips the prefix **only** for a `GraphKeyedPath` — the one domain that carries it by construction — so it cannot become the prefix-shadow bug in reverse.

The fixture is now root-anchored (`/pkg/auth/`) as you asked. **Sabotage-verified:** passing the graph key straight to `MatchFile` drops the codeowner signal and leaves only the two person reviewers.

### 3 — delete / rename old-side joins

`TestJoinHunksToSymbolsPrefixedDeleteAndRename` drives `joinHunksToSymbols` against a prefix-shadowed graph with both hunk-less change kinds — the delete's `Path` and the rename's `PreviousPath` — and asserts each resolves the nested old-side key and never the same-named shadow. Sabotage-verified: declaring the vanished path `GraphKeyedPath` makes both old-side symbols disappear.

`PrefixedDeleteAndRename` is in the Windows selector.

### Four more fixtures had the same defect as the code

Fixing (1) turned four tests red — `TestGetPRImpact_SuppliedFilesNoForge`, `TestPRTools_GCXTOONBudget`, and both `TestConflictsPRs_*`. I checked rather than reverted: `prToolsTestServer`, `conflictsTestServer` and `conflictsBudgetServer` all hard-code `/`-spelled graph keys **and** hand the same strings back as forge-supplied files. That describes an index a Windows daemon never writes, and it is why those paths looked correct before.

They now store native and supply the forge spelling — the shape actually under test. Same correction `TestChangedSymbolsForFiles_RepoPrefixJoin` needed in the previous round.

### Validation

- `internal/mcp` on windows/amd64: **27 → 24** against `main`, failing test *names* diffed, newly-broken set **empty**.
- `internal/analysis`: full package clean.
- `internal/review`: the same four pre-existing failures, untouched.
- Exact Windows selector across all six packages: `ok`.
- `golangci-lint`: 6 staticcheck findings, pre-existing.

Two things worth stating plainly, since both cost me a detour:

- A first combined run showed `TestMapGitDiffRepoPrefixJoin` and `TestMapGitDiffDeletedFile` red. They pass in isolation and the package passes clean on a rerun — load-dependent flakes on my machine, not this change. I would rather name them than have them show up later as a mystery.
- All figures above are pinned with `GOTOOLCHAIN=go1.26.6`, matching this branch's `go.mod`. Go 1.27 is installed here for #654 and I did not want to report numbers from a toolchain your CI does not run.

zzet (approved):
Well done @tiendungdev!

Issue #635, opened 2026-08-20
mcp: lint test for closed tools reading undeclared argument keys


pbednarcik:
Follow-up from #614, as discussed in review: a lint-style test that fails when a closed tool's handler reads an argument key its published schema does not declare. With the guard's reject mode, an undeclared-but-honored key is #597 in reverse — a working option the schema calls unknown, refused before the handler can honor it.

## Shape of the test

Sweep every registered tool with `additionalProperties: false`, map its `mcp.NewTool("x", …)` registration to its handler, and scan the handler body for `req.Get*` / `Require*` key reads; fail on any read of a key the schema does not declare (modulo the generic shaping keys the guard exempts). The reviewer ran this ad hoc against the #614 tip: **0 undeclared reads across 153 closed tools**, so the test starts green and pins the invariant from day one.

## Known caveat to encode: shared helpers are invisible to a body-only scan

`resolveScopeForRequest` reads `ref` on behalf of 8 tools that don't declare it (`audit_health`, `find_clones`, `find_files`, `get_communities`, `get_processes`, `get_recent_changes`, `run_inspections`, `search_text`). No rider fires today only because `reconcileToolParams` alias-rewrites `ref` → `repo` (levenshtein 2) before the guard sees it — a pre-existing rewrite that both masks the undeclared read and deserves its own look. The test should either follow key reads one call level into shared helpers or carry an explicit allowlist naming this case, so the mask can't silently grow.

## Related: the closed stamp publishes a slightly stricter contract than dispatch honors

The generic shaping keys are accepted on every tool but declared on few (measured on the #614 tip: `format` undeclared on 55 of 153 closed tools, `max_bytes` on 105, `max_tokens` on 141, `cursor` on 148, `fields` on 149). Sharpest instance: `read_file`'s description documents `format:"gcx"` while its schema now forbids `format`. Nothing breaks at runtime, but schema-aware clients are steered away from documented, working options. Options: declare the five keys where a tool honors them, or state the exemption once in the schema legend the compact-params gloss already points to.

PR #650, opened 2026-08-21
find_usages: honor declared limit/compact; close the exclude_tests node-kind leak

pbednarcik:
An agent-run evaluation battery against my production C# codebase caught two places where find_usages behaves differently from what its schema declares. Both are fixed here, test-first. The commits are independent if you want them separately.

## 1. `limit` was declared but never read, and `compact` was dead on gcx-default sessions

Observed: a call with `limit:12` returned 64 edge rows, and `compact:true` produced output byte-identical to the run without it. Two separate causes:

- `handleFindUsages` never reads `limit`. The schema promised "Max nodes (default: 50)"; the handler returned everything, always.
- The handler checks `isGCX` before `isCompact`, the opposite order of the shared `returnSubGraph` renderer. For clients whose session default is gcx (claude-code among them), `isGCX` is true with no `format` argument, so `compact:true` never reached the compact renderer.

This is the inverse of the class the #635 lint targets: that one catches keys a handler reads without declaring, this was a declared key the handler never read. A schema-trusting agent gets no signal that its `limit` did nothing; it just pays for the rows.

What the fix does:

- `limit` caps the usage rows (edges). The cap runs after every filter and after the `usage_summary` rollup, so the summary still describes the complete usage set while the page below it is bounded. `total_edges` / `total_nodes` keep the full post-filter counts, `truncated` marks the cut, and nodes no longer referenced by a surviving edge are pruned. `limit:0` opts out, mirroring the `max_bytes` convention.
- On the GCX wire, `truncated=true` plus `total_edges=N` ride the header meta only when the cap fired, so uncapped responses stay byte-identical.
- `compact:true` takes precedence over the gcx session default, the same order `returnSubGraph` already applies for every other traversal tool.
- `group_by:"file"` buckets the same capped page; a truncated grouped response carries the full `total_edges` next to its per-page counts.
- The `limit` schema text now states the real semantics (rows, default 50, 0 = no cap, truncation indicator).

One deliberate behavior change to flag: the default of 50 is now enforced. A symbol with more usages answers with 50 rows, a `truncated` marker, and the full total, instead of everything. That is what the schema has always promised, but callers relying on the undocumented uncapped behavior will see the difference, and `limit:0` restores it per call. If you would rather keep the default uncapped and make the cap opt-in, that is a small change to the same code path; I went with the schema's published promise.

Two scope notes: I deliberately left out a cursor for paging into the truncated tail (`limit:0` reaches it, and a cursor felt like its own discussion), and `max_bytes` / `max_tokens` turned out to be already honored through the shared budget layer, so they are untouched.

## 2. `exclude_tests` leaked node kinds the test-edge pass never stamps

Observed: `exclude_tests:true` on a hot symbol dropped 122 edges to 64, but the "production-only" result still contained rows from `Test\**`: the test file's own file-level reference rows and a `#param` node of a test method. Method and class usages filtered correctly.

Root cause: `markTestSymbolsAndEmitEdges` stamps `Meta["is_test"]` on function and method symbols only. A file node under a test directory gets a different key (`is_test_file`), and parameter nodes get no test metadata at all, while `isTestSource` trusted `is_test` alone. The filter was exactly as good as the stamp, and the stamp only covers two kinds.

I fixed it on the filter side: `isTestSource` keeps the stamp as the first authority and falls back to `testpath.IsTestFile` for unflagged nodes. `testpath` is the canonical predicate the stamp itself derives from, so the fallback cannot disagree with the stamp; it only extends it to kinds the pass never visits. It takes effect on existing stores immediately, with no reindex, and it covers every from-node kind without persisting metadata on high-cardinality param nodes. Query-time path classification is also the established pattern elsewhere in the tree: astquery's `ExcludeTests` and the resolver's test-path gate both do it. Because the fallback lives in `isTestSource`, the `exclude_tests` flag on get_callers and the name-only-candidates path inherit the fix as well.

## Tests

- `TestFindUsagesScoped_ExcludeTestsCoversUnstampedKinds` builds the graph exactly as the indexer leaves it (stamped test method, file node with `is_test_file` only, bare param node, all under `Test\`) and failed with 3 surviving edges where 1 belongs.
- The mcp suite (`find_usages_output_options_test.go`) pins `limit` capping, the default of 50, the `limit:0` opt-out, the capped-only GCX truncation meta, the grouped-page total, and compact precedence over gcx. The limit tests failed with the full row set returned; the compact test failed with GCX bytes.

### DIFF
--- a/internal/daemon/federation.go
+++ b/internal/daemon/federation.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/zzet/gortex/internal/graph"
 	"github.com/zzet/gortex/internal/query"
+	"github.com/zzet/gortex/internal/respbudget"
 )
 
 // federationReadTools is the allowlist of read traversal tools eligible
@@ -133,6 +134,16 @@ func (f *Federator) Augment(ctx context.Context, tool string, body, localResult 
 
 	localTool, wrapped := unwrapToolJSON(localResult)
 
+	// Renderings that cannot round-trip as JSON (compact, gcx, toon,
+	// mermaid, dot) have no merge adapter: fanning out would silently
+	// discard every remote row behind a local-only body. Skip the
+	// fan-out entirely and make the partiality explicit instead —
+	// merging the canonical graph before rendering is the deeper fix,
+	// but until then a labeled local answer beats an unlabeled one.
+	if !json.Valid(localTool) || len(localTool) == 0 || (localTool[0] != '{' && localTool[0] != '[') {
+		return annotateLocalOnlyFormat(localResult, wrapped, len(remotes))
+	}
+
 	results, meta := f.fanOut(ctx, tool, body, remotes)
 
 	merged, origins := f.merge(tool, body, localTool, results)
@@ -157,11 +168,123 @@ func (f *Federator) Augment(ctx context.Context, tool string, body, localResult 
 			len(meta.RemotesFailed), remoteOnlyOrPartial(meta))
 	}
 
+	// The caller's byte/token budget binds the FINAL representation:
+	// each daemon budgeted only its own page, so a multi-peer merge can
+	// overshoot every source's cap combined. Same arithmetic and
+	// truncation meta as the per-daemon budget layer (respbudget), so
+	// a trimmed merge is legible the same way a trimmed page is.
 	mergedWithMeta := attachFederation(merged, meta)
+	mergedWithMeta = budgetMergedJSON(mergedWithMeta, respbudget.EffectiveFromArgs(argsMapFromBody(body)))
 	if !wrapped {
 		return mergedWithMeta
 	}
 	return rewrapToolJSON(localResult, mergedWithMeta)
+}
+
+// budgetMergedJSON applies the caller's effective budget to the merged
+// representation with the same structural trim (and truncation meta)
+// the per-daemon budget layer uses. The generic trim bounds the
+// top-level lists; the federation origins map scales with the node set
+// — a coupling the shape-agnostic trim cannot know — so origins are
+// re-pruned to the surviving nodes and the trim re-run until the
+// result fits or nothing further can shrink.
+func budgetMergedJSON(raw []byte, budget int) []byte {
+	if budget <= 0 || len(raw) <= budget {
+		return raw
+	}
+	var generic map[string]any
+	if err := json.Unmarshal(raw, &generic); err != nil {
+		return raw // non-JSON body: some other layer's contract
+	}
+	for pass := 0; pass < 3; pass++ {
+		trimmed, _ := respbudget.Apply(generic, budget)
+		m, ok := trimmed.(map[string]any)
+		if !ok {
+			break
+		}
+		generic = m
+		pruneOriginsToNodes(generic)
+		out, err := json.Marshal(generic)
+		if err != nil {
+			return raw
+		}
+		if len(out) <= budget {
+			return out
+		}
+	}
+	out, err := json.Marshal(generic)
+	if err != nil {
+		return raw
+	}
+	return out
+}
+
+// pruneOriginsToNodes drops origins entries whose node the budget trim
+// removed — an origin annotates a returned row (attachFederation puts
+// the map at top level, beside the rows), so one keyed to a trimmed
+// node is dead weight against the budget.
+func pruneOriginsToNodes(m map[string]any) {
+	origins, ok := m["origins"].(map[string]any)
+	if !ok {
+		return
+	}
+	keep := make(map[string]bool)
+	if nodes, ok := m["nodes"].([]any); ok {
+		for _, n := range nodes {
+			if nm, ok := n.(map[string]any); ok {
+				if id, ok := nm["id"].(string); ok {
+					keep[id] = true
+				}
+			}
+		}
+	}
+	for id := range origins {
+		if !keep[id] {
+			delete(origins, id)
+		}
+	}
+}
+
+// argsMapFromBody extracts the tool arguments map from the routed
+// request body's production envelope ({"arguments":{...}}, the shape
+// subGraphArgsFromBody and bareNameFromBody parse). Nil when absent.
+func argsMapFromBody(body []byte) map[string]any {
+	var env struct {
+		Arguments map[string]any `json:"arguments"`
+	}
+	if err := json.Unmarshal(body, &env); err != nil {
+		return nil
+	}
+	return env.Arguments
+}
+
+// annotateLocalOnlyFormat appends an explicit local-only note to a
+// non-mergeable-format response as a second content item, leaving the
+// primary rendering byte-for-byte intact. Error results and unwrapped
+// bodies pass through unchanged — an error is already explicit, and a
+// bare body has no envelope to carry the note.
+func annotateLocalOnlyFormat(localResult []byte, wrapped bool, peerCount int) []byte {
+	if !wrapped {
+		return localResult
+	}
+	var m map[string]any
+	if err := json.Unmarshal(localResult, &m); err != nil {
+		return localResult
+	}
+	if m["isError"] == true || m["is_error"] == true {
+		return localResult
+	}
+	content, ok := m["content"].([]any)
+	if !ok {
+		return localResult
+	}
+	note := fmt.Sprintf("note: local results only — %d enabled federation peer(s) were not queried because this response format cannot be merged; use the JSON format for federated results.", peerCount)
+	m["content"] = append(content, map[string]any{"type": "text", "text": note})
+	out, err := json.Marshal(m)
+	if err != nil {
+		return localResult
+	}
+	return out
 }
 
 func remoteOnlyOrPartial(meta FederationMeta) string {
@@ -278,6 +401,12 @@ func (f *Federator) merge(tool string, body, local []byte, remotes []remoteResul
 	sort.Slice(remotes, func(i, j int) bool { return remotes[i].slug < remotes[j].slug })
 	switch tool {
 	case "find_usages", "get_callers", "get_call_chain", "get_dependents":
+		// group_by:"file" renders a grouped shape, not a SubGraph —
+		// round-tripping it through the flat merge would discard every
+		// grouped field and answer with an empty subgraph.
+		if tool == "find_usages" && groupByFileRequested(body) {
+			return mergeGroupedUsages(body, local, remotes)
+		}
 		return mergeSubGraph(tool, body, local, remotes)
 	case "search_symbols":
 		return mergeKeyedList(local, remotes, "results")
@@ -311,6 +440,165 @@ func subGraphArgsFromBody(body []byte) subGraphMergeArgs {
 	return subGraphMergeArgs{}
 }
 
+// groupByFileRequested mirrors the handler's group_by:"file" gate on
+// the routed body, so the merge dispatch and the render agree on when
+// the grouped shape is in play.
+func groupByFileRequested(body []byte) bool {
+	gb, _ := argsMapFromBody(body)["group_by"].(string)
+	return strings.EqualFold(strings.TrimSpace(gb), "file")
+}
+
+// The wire shape of a group_by:"file" find_usages response, as
+// groupUsagesByFile serializes it on every daemon.
+type groupedUsagesWire struct {
+	GroupedBy  string            `json:"grouped_by"`
+	FileCount  int               `json:"file_count"`
+	TotalUses  int               `json:"total_uses"`
+	Groups     []*usageGroupWire `json:"groups"`
+	Truncated  bool              `json:"truncated"`
+	TotalEdges int               `json:"total_edges,omitempty"`
+}
+
+type usageGroupWire struct {
+	File  string         `json:"file"`
+	Count int            `json:"count"`
+	Uses  []usageRowWire `json:"uses"`
+}
+
+type usageRowWire struct {
+	Line        int    `json:"line"`
+	EdgeKind    string `json:"edge_kind"`
+	Context     string `json:"context,omitempty"`
+	ReturnUsage string `json:"return_usage,omitempty"`
+	SymbolID    string `json:"symbol_id,omitempty"`
+	SymbolName  string `json:"symbol_name,omitempty"`
+}
+
+// mergeGroupedUsages merges group_by:"file" find_usages responses
+// group-wise: rows dedupe on (file, line, kind, symbol), counts and
+// totals are recomputed over the union, and the caller's limit is
+// reapplied once, globally, mirroring the flat merge's contract.
+func mergeGroupedUsages(body, local []byte, remotes []remoteResult) ([]byte, map[string]string) {
+	origins := map[string]string{}
+	var lg groupedUsagesWire
+	if err := json.Unmarshal(local, &lg); err != nil || lg.GroupedBy == "" {
+		return local, origins
+	}
+	byFile := make(map[string]*usageGroupWire, len(lg.Groups))
+	order := make([]string, 0, len(lg.Groups))
+	rowSeen := map[string]bool{}
+	rowKey := func(file string, u usageRowWire) string {
+		return file + "\x00" + strconv.Itoa(u.Line) + "\x00" + u.EdgeKind + "\x00" + u.SymbolID
+	}
+	addRows := func(g *usageGroupWire) {
+		dst := byFile[g.File]
+		if dst == nil {
+			dst = &usageGroupWire{File: g.File}
+			byFile[g.File] = dst
+			order = append(order, g.File)
+		}
+		for _, u := range g.Uses {
+			k := rowKey(g.File, u)
+			if rowSeen[k] {
+				continue
+			}
+			rowSeen[k] = true
+			dst.Uses = append(dst.Uses, u)
+		}
+	}
+	for _, g := range lg.Groups {
+		if g != nil {
+			addRows(g)
+		}
+	}
+	// A source that already capped or trimmed its page makes the merged
+	// totals a floor; total_edges falls back to total_uses for sources
+	// that answered complete (they omit the explicit floor).
+	anyTruncated := lg.Truncated || budgetTrimmed(local)
+	totalEdgesFloor := max(lg.TotalEdges, lg.TotalUses)
+	for _, rr := range remotes {
+		var rg groupedUsagesWire
+		if err := json.Unmarshal(rr.toolJSON, &rg); err != nil || rg.GroupedBy == "" {
+			continue
+		}
+		anyTruncated = anyTruncated || rg.Truncated || budgetTrimmed(rr.toolJSON)
+		totalEdgesFloor = max(totalEdgesFloor, max(rg.TotalEdges, rg.TotalUses))
+		for _, g := range rg.Groups {
+			if g != nil {
+				addRows(g)
+			}
+		}
+	}
+	// One deterministic global order before any cut: rows within a
+	// group by (line, kind, symbol), groups by count desc then path —
+	// the same order the local renderer emits.
+	groups := make([]*usageGroupWire, 0, len(order))
+	mergedRows := 0
+	for _, file := range order {
+		g := byFile[file]
+		sort.Slice(g.Uses, func(i, j int) bool {
+			a, b := g.Uses[i], g.Uses[j]
+			if a.Line != b.Line {
+				return a.Line < b.Line
+			}
+			if a.EdgeKind != b.EdgeKind {
+				return a.EdgeKind < b.EdgeKind
+			}
+			return a.SymbolID < b.SymbolID
+		})
+		g.Count = len(g.Uses)
+		mergedRows += g.Count
+		groups = append(groups, g)
+	}
+	sort.Slice(groups, func(i, j int) bool {
+		if groups[i].Count != groups[j].Count {
+			return groups[i].Count > groups[j].Count
+		}
+		return groups[i].File < groups[j].File
+	})
+	totalEdgesFloor = max(totalEdgesFloor, mergedRows)
+
+	args := subGraphArgsFromBody(body)
+	limit := 50
+	if args.Limit != nil {
+		limit = *args.Limit
+	}
+	if limit > 0 && mergedRows > limit {
+		remaining := limit
+		kept := groups[:0]
+		for _, g := range groups {
+			if remaining == 0 {
+				break
+			}
+			if len(g.Uses) > remaining {
+				g.Uses = g.Uses[:remaining]
+				g.Count = remaining
+			}
+			remaining -= g.Count
+			kept = append(kept, g)
+		}
+		groups = kept
+		anyTruncated = true
+		mergedRows = limit
+	}
+
+	out := groupedUsagesWire{
+		GroupedBy: "file",
+		FileCount: len(groups),
+		TotalUses: mergedRows,
+		Groups:    groups,
+		Truncated: anyTruncated,
+	}
+	if anyTruncated {
+		out.TotalEdges = totalEdgesFloor
+	}
+	b, err := json.Marshal(out)
+	if err != nil {
+		return local, origins
+	}
+	return b, origins
+}
+
 // mergeSubGraph merges query.SubGraph responses: nodes deduped by string
 // ID (local wins), edges by (From,To,Kind,FilePath,Line) so distinct
 // call sites of the same pair stay distinct rows. Origins keys each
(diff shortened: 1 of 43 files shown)


zzet (changes requested):
Requesting changes after a full diff and downstream call-path review at fb45a552.

The two local fixes work in their basic in-memory cases, but the new contracts do not hold across the production paths.

1. The usage limit is per daemon, not global.

Federator.Augment forwards the same request to every peer, so each daemon independently applies limit:N. mergeSubGraph then appends every materialized peer page without reapplying the requested limit, overwrites TotalEdges/TotalNodes with the merged page lengths, and ignores remote Truncated/full-total metadata. With limit:2, a local result plus one peer can return four rows; a complete local row plus a remote 10-row result capped to two can return three rows with truncated:false and total_edges:3. With the default, the response can grow to 50 times the daemon count.

Please apply the limit once after a deterministic global merge and preserve remote truncation/full-count semantics. A post-merge slice alone cannot provide an exact deduplicated full total after peers have discarded their tails, so this needs either uncapped/paginated peer retrieval or explicitly inexact lower-bound metadata. Add federation tests covering global row count, remote-only truncation, and distinct call sites at different file/line locations.

2. Compact and TOON violate the advertised capped-response contract.

The schema says a capped response is marked truncated and total_edges retains the full count. Compact instead reports TotalNodes in its truncation line and labels len(page) as “edges: N total.” Flat TOON exposes a node total plus truncated=true but no full edge total. The current compact test has only four rows, so it never activates the cap.

Also, making compact override a GCX-default session now moves that request from the budgeted GCX path to returnSubGraph’s direct compact renderer, where max_bytes/max_tokens are not enforced. limit:0 makes this unbounded. Please add a capped format matrix for JSON, GCX, compact, and TOON, including compact+GCX-default with byte/token budgets.

3. The capped subset is arbitrary and backend-dependent.

truncateUsageRows slices sg.Edges[:limit] without a semantic/stable sort. FindUsagesScoped preserves incoming-edge order; memory returns insertion/mutation order while SQLite orders by kind/id. The default page can therefore differ by backend and graph history, and earlier weak ast_inferred rows can evict later lsp_resolved usages before the relevance-aware JSON degradation layer sees them.

Define one stable global order before truncation—evidence tier/confidence first, then file/line/kind/from/to tie-breakers—and add memory/SQLite parity tests with more than 50 mixed-tier rows plus a re-export union case.

4. exclude_tests is applied after the SQLite frontier cap in get_callers/BFS.

ExpandFrontier stops after opts.Limit raw hops, then admit applies isTestSource. If the first 50 rows are tests and a production caller sorts later, all fetched rows are discarded, the production caller is never examined, and the result can report Truncated:false. This ordering predates the PR, but the broader path fallback materially exacerbates it by filtering more rows after the cap.

Filter before consuming the eligible-result limit, or continue fetching until the accepted limit or true exhaustion. Add a SQLite get_callers regression where filtered tests exhaust the first raw page.

5. The node-kind leak is only partially closed.

Annotation-discovered tests may live in production-looking paths such as src/lib.rs or src/Service.java. Only function/method nodes receive is_test, so their params/locals/closures still have neither their owner’s stamp nor a test-looking path and remain in exclude_tests results. In addition, usage_summary.n_test_refs and GCX from_is_test still use the metadata-only classifier, so path-classified rows are filtered as tests but labeled/count as production when exclusion is disabled.

Please propagate or resolve test ownership for child nodes and centralize the classifier used by filtering and output metadata. Add annotation-owner child tests and MCP summary/GCX assertions for unstamped path-classified kinds.

All current checks are green, but the added tests cover only an in-memory graph, same-tier edges, JSON/basic GCX/grouped JSON, uncapped compact output, and one conventional C# test directory. They do not exercise the failures above.

pbednarcik:
All five points were real; thanks for the depth of the pass. Five commits address them, one per finding, each test-first with the failure watched before the fix.

1. Federation (431f0c5c). The merge now recaps once, globally: remotes merge in slug order (completion order was nondeterministic), find_usages rows re-sort into the same stable order the local cap consumes and cut to the caller's limit once, BFS-shaped tools cap nodes on the deterministic merge order. Truncation from any source propagates. On the exact-total question I took your second option: once any source discarded its tail the exact deduplicated total is unknowable, so the merged totals become an explicit floor and lower_bound is set; uncapped peer retrieval would reintroduce the unbounded payloads this PR exists to close. The edge dedup key gains file/line, so distinct call sites of the same (from, to, kind) survive the merge. Tests: global row count under limit, remote-only truncation propagation, distinct call sites.

2. Compact and TOON (66de7589). The compact footer now reads "edges: N of M total" when a row cap fired, and TOON emits total_edges alongside truncated in the same case; uncapped responses keep their wire shape byte-for-byte. Compact also joins the byte/token budget: the renderer trims to effectiveBudget at a line boundary with a marker line, so compact over a gcx session default cannot shed max_bytes/max_tokens, and limit:0 is no longer unbounded there. The format matrix now covers JSON, GCX, compact, and TOON with the cap actually firing, plus compact with a byte budget. One find along the way: toon-go rejects the named ZeroEdgeClass string type, so every caveat-bearing subgraph was silently falling back to JSON; the caveat now lowers into a plain-string projection and TOON renders again.

3. Stable order (8f33b354). SortEdgesForPage defines the order any row cap consumes: origin tier rank, then confidence, then file/line/kind/from/to. Applied only when a cap fires, so uncapped responses are unchanged. Tests: an early ast_inferred row can no longer evict an lsp_resolved usage, memory and SQLite return the identical page on mixed edge kinds (your kind-grouped iteration order reproduced exactly), and the barrel re-export union pages the same way.

4. BFS starvation (a2da8162). With a post-fetch filter active (test exclusion or workspace scope) the walk now takes the per-node path, which filters before any cap and fetches to true exhaustion; it is the same gate the BFSCapable fast path already applied, so the expander keeps serving the unfiltered fast case. This trades round-trips for correctness on filtered disk-backend walks; if that shows up in your numbers, a paginated expander is the follow-up I would try. Regression test on the SQLite backend: a raw page of stamped test callers ahead of one production caller.

5. One classifier (80e6bdc3). graph.NodeIsTest is now the single authority for both filtering and labels: the node's own stamp, then the owner's stamp for child nodes through the `<owner-id>#...` convention (covers annotation-discovered tests in production-looking paths, whose params carry neither stamp nor test path), then the path predicate. Wired into the query-side filter and every output surface that rides next to it: find_usages GCX rows, the usage summary, the shared traversal encoder, and subgraph TOON rows. Pure listing tools keep the raw stamp. Tests: annotation-owner param exclusion, and summary/GCX labels matching the filter on unstamped kinds.

Full suite compared by FAIL name-set against a clean checkout of the base: no new failures from the branch. go test -race over query, mcp, daemon, and graph is clean.

zzet (changes requested):
Requesting changes on head `431f0c5` after a second full pass and exact-head reproductions.

1. **Federation parses the wrong request shape** (`internal/daemon/federation.go:274-277,348-353`). Production stdio and streamable MCP send `{"arguments": {...}}`, but `subGraphMergeArgs` expects flat `{id,limit}`. The merge therefore ignores explicit limits and defaults to 50. A production-shaped `limit:2` reproduction returned four edges. The new federation tests use the incorrect flat body, so they remain green. Please parse the real envelope and test omitted, 0, positive, and greater-than-50 limits through the production routing shape.

2. **Remote completeness metadata is discarded** (`internal/daemon/federation.go:311-347`). The merge reads remote truncation/totals/nodes/edges but never propagates `rsg.LowerBound`, `Boundaries`, or `DynamicBoundaries`. A remote `get_call_chain` or `get_callers` result with `lower_bound:true,truncated:false` becomes falsely exhaustive. Remote usage summaries and caller notes also remain local-only. Please define and test aggregation for these fields.

3. **TOON assigns `is_test` to the wrong retained node** (`internal/mcp/tools_core.go:156-184,391-399`). `nodesToTOONRows` skips File/Import nodes, then the new loop indexes the filtered rows using indexes from the original `sg.Nodes`. Every skipped node shifts later classifications. I reproduced a production function inheriting a skipped test-file flag. Classify each node while its row is built, or join by node ID, and add a mixed File/Import plus production/test TOON test.

4. **The federated confidence ordering cannot survive JSON serialization** (`internal/daemon/federation.go:354-361`, `internal/graph/edge.go:563-575`). Global recap sorts equal-origin rows by numeric confidence, but `Edge.Confidence` is `json:"-"`; after unmarshalling, every peer confidence is zero and file path can make a weaker row displace a stronger one. A `0.1` versus `0.9` same-origin reproduction kept `0.1`. Preserve a sortable ranking value across the federation boundary and add a mixed-confidence peer test.

5. **Overlay filtering and output classification use different graph views** (`internal/mcp/tools_core.go:2805-2840`). The query uses `engineFor(ctx)`, while the new summary/GCX/TOON classifiers receive base `s.graph`. An overlay-added annotation test can be excluded through the overlay reader but have child rows counted or labeled as production. Pass the request-scoped reader/NodeGetter and add overlay owner-child assertions for summary, GCX, and TOON.

6. **Compact output can exceed its hard budget** (`internal/mcp/tools_core.go:257-272`). When `0 < budget < len(marker)`, `trimTextToBudget` returns the complete marker. `max_bytes:1` returned 65 bytes, and `max_tokens:1` has the same issue. Make the marker fit the budget and add boundary tests for 1, marker length minus 1, marker length, and a tiny token budget.

All 11 existing checks pass, but the added tests do not exercise these production paths. No new auth bypass, secret exposure, or introduced dead code was found. The stable local page ordering, filtered SQLite BFS correctness, and central test classifier logic are otherwise improved.

pbednarcik:
All six were real. Four of them come down to the same mistake on my side: the round-one tests exercised request shapes I made up instead of what production routes. The new tests go through the real envelope, the serialized wire, the overlay view, and the boundary values. Fixes in 7a944659 and 8f717552, plus a self-review commit 504d9c0c on top.

1. Request shape: subGraphArgsFromBody reads the "arguments" envelope now, the same shape bareNameFromBody parses. I dropped the flat-body fallback entirely in 504d9c0c, the envelope is the only shape anything sends. Tests route the production shape with omitted, 0, small, and above-default limits.

2. Remote completeness: lower_bound ORs in, boundaries append deduplicated under the walk's own 50 cap, caller notes union with local precedence, and the usage summary is recomputed over the merged rows. Exact when every source was complete, a floor with lower_bound set otherwise.

3. TOON: rows join to nodes by ID now. Your skipped-file repro is the regression test.

4. Confidence: SortEdgesForPage falls back to the confidence label after the numeric compare. In process the numeric decides; in the merge every row crossed the wire (the local body included), so the label always decides there. Pinned as EXTRACTED displacing AMBIGUOUS under limit:1 with the tie-breakers pointing the wrong way. If you'd rather serialize the numeric across federation instead, that's a small follow-up, I just didn't want to widen the edge wire unasked.

5. Overlay views: nodeGetterFor(ctx) hands every output surface the same request-scoped view the query ran on. The test adds an owner and its param child that exist only in the overlay and asserts the child counts as a test ref.

6. Budget floor: the budget is a hard ceiling at every size, a budget at or below the marker length gets a truncated marker. Boundary tests at 0, 1, marker-1, marker, and a one-token budget through the handler.

The self-review pass before pushing (504d9c0c) caught more of the same class: the merged summary could shrink below a source's whole-set counts (now the element-wise floor of every source's rollup, remote ones included), a stale "appears unused" caveat surviving a merge that disproves it, dynamic boundaries merging uncapped, caller notes dangling after the post-merge prune, and the zero-edge caveat calling overlay-resolved symbols mistyped. Also folded the rollup into one query.UsageSummaryOf and moved the label rank next to its producer in graph, so the copies can't drift.

Verification as before: FAIL name-sets diffed against clean main with no new names, race detector clean over daemon, mcp, query, and graph.

zzet (changes requested):
Requesting changes on head `504d9c0` after a second-pass exact-head review and focused reproductions.

The earlier fixes are largely correct, and the normal/race suites plus all 11 hosted checks pass, but these production paths still violate the declared contracts:

1. **Federation destroys `find_usages(group_by:"file")` output** (`internal/daemon/federation.go:302-307`, `internal/mcp/tools_core.go:2968-3023`).

   Federation dispatches by tool name and unmarshals the grouped JSON into `query.SubGraph`. The grouped fields are treated as unknown and discarded, so a valid response becomes an empty `nodes:null, edges:null` subgraph. This was reproduced with local and remote grouped results. Preserve/merge the grouped representation and add a production-envelope federation test covering groups, totals, and limits.

2. **The final federated response is not constrained by `max_bytes` or `max_tokens`** (`internal/daemon/federation.go:272-307`).

   Each daemon applies the budget independently, then `Augment` merges the pages without reapplying the caller's global budget. A focused reproduction with `max_bytes:400` returned 1,102 bytes. Budget truncation metadata is also discarded during the SubGraph round trip. Apply the budget to the final deterministic representation and test multiple peers for both byte and token limits.

3. **Explicit TOON bypasses response budgets entirely** (`internal/mcp/tools_core.go:198-236`).

   The TOON branch calls `subGraphToTOON` directly before the budget-aware response path. A reproduction with `format:"toon"`, `limit:0`, and `max_bytes:120` returned 8,651 bytes; `max_tokens` follows the same unchecked path. Route TOON through a budget-aware renderer and cover byte, token, and tiny-budget boundaries.

4. **Federated caveat/completeness merging is stale and unsafe** (`internal/daemon/federation.go:328-393`).

   Remote caveats and fields such as `tier_filtered`, `text_matched_suppressed`, and `name_only_candidates` are not aggregated. Caveat reconciliation only runs when merged edges are non-empty. A local `likely_unused` plus remote `coverage_incomplete` therefore remains `likely_unused`, which can make uncertain removal advice appear conclusive. Define conservative precedence and test zero-row conflicts, remote incomplete coverage, and remote-only suppression metadata.

5. **Non-JSON federation silently returns local-only data** (`internal/daemon/federation.go:302-307`, `internal/mcp/tools_core.go:198-236`).

   `compact`, `gcx`, `toon`, Mermaid, and DOT responses cannot be unmarshalled as SubGraph. The remote fan-out still occurs, but the merge returns the unchanged local text and omits all remote rows. Merge the canonical graph before rendering, or explicitly gate/reject non-federated formats rather than silently presenting partial results.

6. **Overlay zero-edge caveats are classified against the base graph** (`internal/mcp/tools_core.go:2397-2427,2686-2696`, `internal/graph/extraction_gap.go:156-166`).

   Both `get_callers` and `find_usages` resolve through the request-scoped overlay, then pass `s.graph` into zero-edge classification. Overlay structural evidence, tombstones, and boundaries can therefore produce incorrect or missing caveats. The new test asserts only that "mistyped" is absent, not the actual caveat class. Pass the request reader through classification and add exact-class tests for overlay-only structural/weak evidence and tombstoned base callers.

Also add regression coverage for capped filtered BFS membership: the fallback consumes backend/insertion order before the limit, so page membership is not stable across graph history/backends. Because scoped filtered traversals now bypass batched frontier expansion, a query-count test or benchmark should also guard against a per-node-query regression.

No new auth bypass, secret, injection path, or dead production helper was found. The budget bypasses do present a resource-amplification concern.

pbednarcik:
All six were real, and the pattern behind four of them was the one you named in round two from the other side: I swept the paths I was touching but never tested the declared contract from the outside — request in, bytes out, measured against what the schema promises. That battery now exists (a contract matrix over every find_usages rendering, both backends, limit and both budget axes), and the pre-push review over these fixes surfaced and closed nine further defects of the same families before you had to. Fixes in 4207e482 and a6b745ba, hardening in 07977585.

1. group_by through federation: grouped bodies dispatch to a grouped merge — rows dedupe on (file, line, kind, symbol, name, context), counts and totals recompute over the union, the caller's limit is reapplied once globally, and rows carry origins like the flat merge. The group and row shapes moved to internal/query so the renderer and the merge share one wire contract and a field added on one side can never be silently stripped by the other. Your empty-subgraph repro is the regression test, through the production envelope. The flat merge also refuses shapes outside its nodes/edges contract now, so a future grouped variant on another tool passes through local-only instead of zero-valuing into an empty answer.

2. Global budget: the budget arithmetic moved to one shared package (internal/respbudget) and the merge applies it to the final representation after the federation block attaches, with origins re-pruned to the surviving rows of whatever lists the shape carries — nodes, keyed-list results, or grouped rows. Multi-peer byte and token tests included. A source page the budget layer already trimmed propagates truncated and lower_bound: the _truncated_by_budget marker is a map key the typed round trip discards, so the merge reads it off the raw source bytes.

3. TOON budget: TOON — and mermaid and dot, which had the same bypass — render inside the same effectiveBudget cap as every other format, and the cut re-renders over the largest edge prefix that fits instead of slicing bytes, so the output stays syntactically valid (a sliced TOON table contradicts its own row-count headers; a sliced dot document loses its closing brace). Byte, token, and tiny-budget boundary tests included.

4. Caveat merging: with zero merged rows the most conservative source caveat survives (coverage_incomplete over possible_extraction_gap over likely_unused; the rank lives in graph beside the class producer). Two refinements from testing the topology honestly: a peer that resolved nothing answers an extraction-gap caveat about its own graph, not the union, so it cannot displace a classification from a daemon that resolved the node — otherwise likely_unused becomes unreachable in any multi-peer install; and a tier_filtered emptiness keeps carrying no zero-edge caveat, mirroring the handler contract. text_matched_suppressed, name_only_candidates, and tier_filtered floor across sources.

5. Non-JSON federation: gated explicitly. A rendering that cannot round-trip as JSON skips the fan-out and carries a visible local-only note naming the peer count, instead of silently presenting a partial result as federated. Merging the canonical graph before rendering is the right end state; I kept it out of this PR because it means either moving the renderers below the daemon or re-rendering after merge, and I did not want to widen the diff unasked. Happy to take it as a follow-up.

6. Overlay caveats: classification takes a minimal reader interface and both handlers pass the request-scoped view, so overlay structural evidence, tombstones, and overlay-only symbols classify against what the query actually saw. The tests assert exact classes — likely_unused for overlay structural evidence, coverage_incomplete for hidden weak evidence, likely_unused when a tombstone removed the only base caller. One known cost: OverlaidView does not implement the InDegreeForNodes fast path, so a zero-edge result for a common method name inside an overlay session takes the per-edge fallback; a view-aware counter is a small contained follow-up.

The BFS asks: the per-node fallback sorts each node's fetched edges into the same evidence order the usage page cap consumes before admitting — gated on the cap being reachable so uncapped expansions and hub walks pay nothing — and capped filtered page membership is pinned memory-vs-SQLite. A query-count guard pins the walk at one edge fetch per expanded frontier node. Scope note: the batched expander and BFSCapable fast paths still page in store order; aligning them is a behavior change to unfiltered paging I did not want to fold in here.

The matrix's own catch beyond your six: the token-budget decoration (_max_tokens / _truncated_by_tokens, and the GCX comment) landed after the fit, so a max_tokens response overshot its own cap by the decoration's size. The decoration's bytes are reserved up front now, and a budget too small to fit the decoration drops it instead of letting it break the contract it documents.

Verification as before: FAIL name-sets diffed against clean main with no new names, race detector clean over daemon, mcp, query, graph.

zzet (changes requested):
Requesting changes on head `079775859dba969998439924cfa9725469f46b5e` after the round-four exact-head review and focused reproductions.

The branch merges cleanly onto current `main`, all 11 hosted checks are green, and the normal/race suites pass. It is still not safe to merge because these production contracts remain broken:

1. **Typed slices bypass `max_bytes` / `max_tokens`** (`internal/respbudget/respbudget.go:184-188,253-275`; live path at `internal/mcp/tools_core.go:1498-1506`). `Apply` shallow-clones a `map[string]any`, but its trimmer recognizes only dynamic `[]any`. `get_symbol detail:"full"` passes typed `[]*graph.Edge` values and has no degradation shape that normalizes them first. With 40 edges and `max_bytes:600`, the exact handler returned 6,782-byte JSON and 4,694-byte TOON. This regresses the prior implementation, which JSON-normalized the map before trimming. Normalize typed slices (or always round-trip) and add direct `respbudget` plus full-symbol JSON/TOON tests.

2. **Budgeting can corrupt node-only structured output** (`internal/mcp/tools_core.go:292-358`). `renderSubGraphWithinBudget` searches only the edge prefix, while `nodesForEdgePage` retains every node with no incident edge. With 100 isolated nodes, zero edges, and `max_bytes:300`, DOT fell through to the generic text trimmer and lost its closing `}`. Fit/prune nodes too and always re-render the final structured subset. Add zero-edge/many-node DOT, Mermaid, and TOON tests.

3. **The federation local-only annotation is outside the caller budget** (`internal/daemon/federation.go:125-135,261-287`). Non-JSON formats return early after `annotateLocalOnlyFormat` appends a second content item, so the final budget pass never runs. A `max_bytes:100` reproduction returned 264 combined text bytes. Budget the complete response and cover both byte/token axes for compact, GCX, TOON, Mermaid, and DOT.

4. **`get_callers` boundary metadata still reads the base graph under overlays** (`internal/mcp/tools_core.go:2472-2537`). The query and zero-edge caveat use the request-scoped engine/reader, but `CallerBoundaries` and the dynamic-boundary lookup use `s.graph`. An overlay-only implementation was returned in the result while `boundaries` stayed empty and `lower_bound` stayed false. Use the request-scoped view throughout and test overlay-added plus tombstoned implements/overrides boundaries.

5. **Federated zero-edge caveat aggregation is still asymmetric** (`internal/daemon/federation.go:675-690,728-750`). A local `possible_extraction_gap` survives even when a remote resolves the target and reports `likely_unused`, because resolution filtering is applied only to remote caveats. Separately, any peer's `tier_filtered` marker clears another source's `coverage_incomplete` warning. Normalize caveats symmetrically and preserve independent cross-source uncertainty; add both source-orientation fixtures.

6. **Capped grouped federation output is source-order dependent** (`internal/daemon/federation.go:468-470,514-559`). The dedup identity includes `Context`, but the comparator used before `limit` does not. Reversing equivalent peer inputs changed which context survived `limit:1`. Make the comparator a total order over every semantic identity field and test reversed peer/backend order.

Also define or enforce the tiny/scalar-heavy JSON budget floor, and preserve remote `BudgetHit`, `StoppedAtDepth`, and `LastSynced` metadata during subgraph merging.

No new auth bypass, secret exposure, conventional injection issue, or material dead production code was found. The earlier fixes for grouped shape, production envelopes, ordinary final JSON budgeting, explicit TOON budgeting, and overlay zero-edge classification are otherwise confirmed.

pbednarcik:
All six confirmed and fixed test-first. Response side in 39f934e1, federation side in 8d311119. The pre-push review caught seven more of the same families, four of them in this round's own fixes, so those are folded in too.

1. Typed slices: my own regression. A fast path I added while hardening round three skipped the JSON round trip that normalizes typed edge slices into what the trimmer can cut. The round trip is unconditional again and the comment now says why it must stay. Direct respbudget test plus full-symbol JSON and TOON tests at your repro scale included. I decided against re-adding a guarded fast path: bounded cost on a cold path is not worth another shot at this bug class.

2. Node-only output: the fit search got a second stage. When no edge prefix fits it binary-searches a node prefix, and only an empty rendering falls through to the text trim, so structured grammars never get byte-sliced. The review caught two things in my first version, both fixed: the node cut now orders edge-touched nodes first so the queried symbol survives, and the diagram note measures the cut against the rows the handler returned, because a limit cap stamps the same shape and the old note would tell you max_bytes:0 restores rows it cannot.

3. Local-only annotation: budgeted by reservation, not by dropping. The router shrinks the budget forwarded to the local dispatch by the note's size when the args show a non-mergeable format, so the note fits by construction, same rule as the token decoration. The annotation still enforces the caller's original cap over the combined content and drops the note only when it truly cannot fit (session-default formats the args cannot reveal, caps below the note's own size). My first fix dropped unconditionally and the review rejected it: a saturating rendering would have erased the only "peers not queried" signal exactly for budget-conscious callers. Byte and token axes across compact, GCX, TOON, mermaid and DOT tested.

4. get_callers boundaries: CallerBoundaries takes a minimal reader interface now and both it and the dynamic-boundary lookup use the request-scoped view. Overlay-added and tombstoned implements boundaries both tested. The review found one more layer: the dispatch scan sliced the on-disk file by the overlay node's buffer coordinates, fabricating boundaries from a different function's body. It reads the session buffer now, same as the source-reading handlers.

5. Caveats: symmetric, with the class semantics moved into graph next to the class producer so the merge stops re-deriving them by enum comparison. An unresolved source's gap caveat yields only to an actual classification from a resolving source; a resolving peer with no caveat erases nothing, so a mixed-version fleet can never merge into a bare "0 usages". On tier_filtered I kept one deliberate asymmetry: the marker names edges that exist below the tier, which refutes a likely_unused conclusion from any source, but says nothing about another source's coverage uncertainty, so coverage_incomplete and gap caveats now ride alongside it. A federated JSON response can therefore carry both markers where a single daemon's cannot; per-source the handler contract stays exclusive. All orientations are fixtures.

6. Grouped cap: the comparator is a total order over the full dedup identity, Context included, and a reversed-peer fixture pins order independence.

The floor is defined, not enforced: the minimum for a structured payload is its scalar skeleton with every list emptied and the meta stamped, pinned by test; text renderers keep their hard ceiling. Walk metadata survives the merge: budget_hit ORs, stopped_at_depth min-merges and drops when a source stopped on budget without reporting depth, last_synced takes the stalest source.

Verification as before: FAIL name-sets diffed against clean main with no new names, race detector clean over daemon, mcp, graph, respbudget.

zzet (approved):

PR #652, opened 2026-08-21
ci: run full test suite on Windows

zzet:
## Summary

- add `windows-latest` to the full test matrix (`fail-fast: false`) and drop the selector-based `build-windows` job
- run the Windows shard without the race detector (Linux and macOS keep `-race`): with it the shard overran the 60-minute job cap before half the packages had reported; the platform contract is what Windows is here to check
- port the suite to Windows: the last complete Windows run had 184 failing tests in 19 packages; every one is fixed below

## What was wrong on Windows, and what changed

**Production fixes** (behaviour on Windows only, byte-identical on POSIX):
- `indexer` / `store_sqlite`: graph paths are keyed with forward slashes on every platform. The indexer kept OS-native separators in graph keys while symbol IDs, URIs and cross-repo lookups assumed `/`, so on Windows the same file existed under two spellings and dozens of lookups (find_import_path, barrel re-exports, localization anchors, contract edges, TS path aliases, GDScript resolution) missed.
- `resolver`: directory keys derived with `/`, so import resolution binds on Windows (bare specifiers, workspace membership, cross-package guard, extern edges).
- `review`, `analysis`, `forge`, `audit`: repo-relative paths in findings, diff headers, hunks, verification commands and discovered config files stay slash-separated (`path.Clean`, not `filepath.Clean`).
- `cmd/gortex`: `reapStaleEmbeddedStores` released the store lock before `RemoveAll` (Windows refuses to delete an open file, so the reap was a silent no-op); `loadHistory` / `runSavings` and `serverstack.SharedServer.Close` release their sidecar SQLite handles; `isAmbiguousLaunchCWD` recognises a drive root; the short-name repo match and the uninstall preview agree on separators.
- `daemon`: a dial to a missing socket is classified as "daemon unavailable" on Windows too — Winsock reports `WSAECONNREFUSED` (10061), which is not Go's `syscall.ECONNREFUSED` there — so hooks, the proxy's restart retry and embedded fallback behave as on POSIX.
- `platform`: one file-claim / file-replace policy (`ClaimFile`, `ReplaceFile`) used by the localization receipt, the hooks tool snapshot and the crash-pool quarantine: POSIX `rename(2)` hands a contended file to exactly one caller, Windows `MoveFileEx` can fail every caller, so exclusive claims go through an `O_EXCL` marker and replaces retry sharing violations.
- `mcp`: graph paths are looked up in the graph's forward-slash spelling everywhere (review pack, PR review context, PR risk, sibling diff, find_import_path, file summary, symbol-ID anchoring), an absolute citation is rejected in either spelling, note ordering is total, and every localization file tier is ordered by declaration site — the sort used to run only when the page had a member to demote, so a tight window kept whatever arrival order the ranked tail produced (which differs between POSIX and Windows).
- `serverstack`: the global memories sidecar opened by `InitMemories` is released on `Close`.

**Test fixes**:
- `testenv.ShortTempDir` replaces every `os.MkdirTemp("/tmp", …)` used for short AF_UNIX socket paths (33 daemon/cmd/hooks failures were `GetFileAttributesEx /tmp`).
- Shell-script stubs (`fake-claude.sh`, `/bin/sh`, `/bin/cat`) became helper-process re-execs of the test binary; PATH stubs carry a PATHEXT extension on Windows.
- POSIX absolute fixtures (`/work/alpha`, `/repo`, `/abs/config`) are built natively; URI expectations derive from native absolute paths; JSON and quoted-path expectations account for `\`.
- Assertions on POSIX-only semantics — mode bits, exec bits, chmod-denied access, self-signalling — are skipped on Windows with a stated reason; nothing is weakened on POSIX.
- The embedding provider test downloads its model from Hugging Face on every run; a stalled download used to hang the package until the 45-minute test timeout, so the download is bounded and a stall skips with a stated reason instead of killing the shard. The `platform.ClaimFile` winner no longer loses its claim when concurrent readers keep the payload open on Windows (Go opens files without `FILE_SHARE_DELETE`, so removal fails while any reader holds the file).
- Three timing assertions measured the runner instead of the behaviour: two health endpoints' `uptime > 0` raced Windows' ~15 ms clock tick (now backdated and asserted `>= 3 s`), and the PASSIVE WAL-checkpoint test timed the CI disk back-filling 408 frames rather than lock contention (PASSIVE never consults the busy handler; the test now times the pass after the back-fill).
- The hooks suite's stdout capture drains the pipe while the hook writes: Windows pipes buffer 4 KiB (POSIX 64 KiB) and the SessionStart briefing is larger, so the first test past that size deadlocked and the whole package hit the go-test timeout — the single reason the shard used to run out its 60-minute cap. `TestRouter_ReloadConfig_ConcurrentNoRace` no longer resolves a real hostname per iteration (183 s on Windows DNS).

## Validation

- macOS: full suite green (the `cmd/gortex` write-through guard trips only when a live daemon on the developer machine appends to the real query log; identical on unmodified `main`).
- Windows: every package and test file cross-compiles (`GOOS=windows CGO_ENABLED=1` with mingw); the CI shard is the behavioural oracle — 153/153 packages pass, in about half an hour. The shard writes no coverage profile: nothing uploads it there, and the runner's PowerShell tokenised `-coverprofile=coverage.out` into an extra `.out` package argument, which was the last thing failing the step.
- `go vet`, `golangci-lint`, `actionlint` clean.

## Follow-ups noted, not in this PR

- `internal/graphpath` still documents the native-separator premise its callers no longer need; `PrefixForms`' native arm only folds graphs indexed by older versions.
- `hostPhysicalMemory()` returns 0 on Windows, so cold-index GC tuning is inert there.
- `-v` on the Windows shard is temporary, to expose per-test timings until the shard is reliably green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,19 @@ jobs:
         run: go build -o gortex ./cmd/gortex/
 
       - name: Test
+        if: matrix.os != 'windows-latest'
         run: go test -race -timeout=45m -coverprofile=coverage.out ./...
+
+      # The Windows shard checks the platform contract, not the race
+      # contract: the race detector multiplies runtime several times over and
+      # the windows runners are already the slowest of the three, so the two
+      # together overran the job cap before the suite finished. Linux and
+      # macOS keep -race, which is where a data race would surface anyway.
+      # -v is temporary: per-test timings are how we find which packages hang
+      # here, and it comes out once the shard is reliably green.
+      - name: Test (windows, no race detector)
+        if: matrix.os == 'windows-latest'
+        run: go test -v -timeout=45m -coverprofile=coverage.out ./...
 
       - name: Upload coverage
         if: matrix.os == 'ubuntu-latest' && matrix.go-version == '1.27'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,9 +38,13 @@ jobs:
       # macOS keep -race, which is where a data race would surface anyway.
       # -v is temporary: per-test timings are how we find which packages hang
       # here, and it comes out once the shard is reliably green.
+      # It writes no coverage profile either: nothing uploads one from this
+      # shard, and the runner's default PowerShell tokenises
+      # -coverprofile=coverage.out into an extra ".out" package argument, so
+      # the job failed on an unresolvable package before any test ran.
       - name: Test (windows, no race detector)
         if: matrix.os == 'windows-latest'
-        run: go test -v -timeout=45m -coverprofile=coverage.out ./...
+        run: go test -v -timeout=45m ./...
 
       - name: Upload coverage
         if: matrix.os == 'ubuntu-latest' && matrix.go-version == '1.27'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,9 @@ jobs:
     # per-step budget needs headroom; this job-level cap still bounds a hang.
     timeout-minutes: 60
     strategy:
+      fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
         go-version: ['1.27']
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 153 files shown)


tiendungdev:
I have the enumeration this draft is waiting for — I ran the full suite on Windows locally while working on #641/#644/#645/#646/#647, so here it is up front rather than after a red shard.

windows/amd64, go1.26.6, `-count=1`, no `-race`.

| package | failures | measured at |
|---|---|---|
| `internal/mcp` | **27** | `c982cf52` (after #645) |
| `internal/indexer` | **32** | `b8b13ca7` |
| `internal/analysis` | 4 | `b8b13ca7` |
| `internal/persistence` | 1 | `b8b13ca7` |
| `internal/graph/store_sqlite` | 5 | pre-#641 |

`store_sqlite` and `persistence` are closed by #641/#644. Two of the four in `analysis` are closed by #644; the other two are **not yours** — see the environment section.

### `internal/indexer` — 32

| cluster | n |
|---|---|
| `TestLoadCompileCommands_*` (clangd compile DB) | 5 |
| `TestGDScript*` | 5 |
| `TestTSConfigPathAlias_*` | 3 |
| symlink confinement (`*SymlinkOutOfRepo*`) | 4 |
| startup-barrier / marker permission fail-closed | 2 |
| unclustered | 13 |

### `internal/mcp` — 27

| cluster | n | note |
|---|---|---|
| symlink / file-mode | 5 | `SeCreateSymbolicLinkPrivilege`, and Windows has no POSIX mode bits |
| review rulepack join | 4 | 3 addressed by #646, 1 is a different cause |
| localization ranking (Swift ×3, Rust, C#, PHP) | 6 | not diagnosed |
| fidelity glob | 1 | #647 |
| POSIX-spelled assertions | ~4 | e.g. `TestPathToFileURI_Absolute` wants `file:///work/main.go`, Windows resolves `file:///D:/work/main.go` |
| unclustered | ~7 | |

## Two things that will bite the shard, and one is the same class as the failure you already hit

**1. Some local Windows failures are the developer's environment, not the platform.** Your note about `.claude/worktrees/*` and `TestNewIsFencedToIndexerStaging` is exactly this class. I hit my own: `TestMapGitDiffRepoPrefixJoin` and `TestMapGitDiffFileChangeKinds` fail here on `TempDir RemoveAll cleanup: …\.git: The directory is not empty` — because the `git` on my PATH is a non-stock fork that writes `.git/ai/logs` into every `git init`. Reproduced with a bare `git init` in an empty directory outside any repo. A clean runner with stock git will **not** show those two, so my `analysis` count of 4 should read as 2 for CI purposes.

Worth stating in the workflow or the contributing notes that the Windows shard is authoritative and a local red is suspect until reproduced there — the inverse of the usual advice, and non-obvious.

**2. `fail-fast: false` plus a full suite will surface order-dependent flakes, and there is at least one.** `TestNotesManager_SaveQueryDelete` passed in one full-package run and failed **3 of 3** when run in isolation on the same commit. I nearly credited a fix for it in #647 on the strength of a before/after count. Anyone reading the first red shard should diff failing test *names* against a baseline rather than counts, or a flake will read as a regression — and vice versa.

## On removing the `build-windows` job

No objection from me, and the reasoning in #647 is fair — a selector list is a guess and mine did drift. One consequence worth deciding deliberately rather than inheriting: those steps also carried three non-test guarantees that the full-suite matrix does not reproduce — the CGO/tree-sitter **build** smoke test (`go build ./...` on native Windows), and the `gortex.exe` CLI build. If the full shard only runs `go test`, a Windows build break lands without a red until someone runs the suite. Cheap to keep as two steps in the new shard if you want that signal.

Happy to take any of the clusters above once the shard has enumerated them for real — the `.NET`/`GDScript`/`TSConfig` ones in particular look like one root cause each rather than per-test work.

tiendungdev:
Full classification, as promised. Measured on `origin/main` `699f351e`, windows/amd64, go1.26.6, `-count=1`, no `-race`, clean worktree.

**173 failing tests across 28 packages** — but **2 of those are my machine, not the repo**, so read the number as 171.

> `internal/analysis` (`TestMapGitDiff*`) and `internal/releases` fail on `TempDir RemoveAll: .git: The directory is not empty`. I inspected the residue: the only thing left in every leftover directory is `.git/ai/working_logs/initial`, created by a `git` fork on my PATH. Not yours. I mention it because the same shape appears in bucket 4, where it *is* real.

## Buckets

| # | Cluster | n | What it is |
|---|---|---:|---|
| 1 | Test harness assumes a POSIX shell | 24 | `fake-claude.sh` / `fake-codex.sh` → `%1 is not a valid Win32 application` (17); `/bin/sh` and `/bin/cat` spawns in `internal/semantic/lsp` (4); `cannot signal this process on this platform` (1); launchd-plist and XDG-env tests running on Windows (2) |
| 2 | `os.Symlink` → `A required privilege is not held` | 12 | every **symlink-confinement** test: `internal/indexer` (4), `internal/mcp` (4), `internal/artifacts` (2), `internal/search/trigram` (2) |
| 3 | POSIX file modes | 5 | `file perm = 666, want 0600`; `should be executable (mode=-rw-rw-rw-)` |
| 4 | Leaked handles → `TempDir` cleanup | 5 | `sidecar.sqlite: The process cannot access the file because it is being used by another process` — `cmd/gortex` (3), `internal/serverstack` (2). Same family as #641 |
| 5 | CRLF checkout | 2 | **fixed in #655** |
| 6 | Not yet attributable | 7 + 1 pkg | spinner glyphs, a GC-percent assertion, a nanosecond-equality assertion, `internal/parser/crashpool` rename race — plus `internal/embedding`, where `TestNewLocalProvider_ReturnsWorkingProvider` **hung for the full 40m timeout** inside `ensureHugotModel`. That is a model download; it may well be my network rather than Windows, and CI's `-timeout=30m` would be exceeded either way. Needs a clean-runner data point |
| 7 | Path domain — `/` vs `\` | ≈115 | the remainder, and the actual work |

Buckets 1–6 are individually enumerated (I have the test names). Bucket 7 is the residual, classified by first error line, so a handful may move as each is opened.

Bucket 7 is the same family as #646 and it reaches production code, not just fixtures. Confirmed samples:

- `internal/forge`: `hunk FilePath = "internal\forge\forge.go", want "internal/forge/forge.go"` — a forge API path, native-spelled
- `internal/savings`: `EventsPathFor("/tmp/savings.json") = "\tmp\savings.jsonl"` — `filepath` where `path` belongs
- `internal/pathkey`: `foldPath` returns `\Users\me\project`
- `internal/excludes`: a pattern containing `\` no longer matches

## Three things I need from you

**1. File modes (bucket 3) — a real decision, and one is security-shaped.**

`internal/daemon` asserts `servers.json` is written `0600`. On Windows that property genuinely does not exist: `os.Chmod` there only toggles the read-only bit, so the file lands `0666` no matter who runs it. This is not a test problem to skip past — it is a narrower confidentiality guarantee on Windows than on POSIX. Options: accept and document it, skip the assertion on Windows, or implement an ACL. Same question, lower stakes, for the emitted `*.sh` hook handlers' exec bit.

**2. The launchd / XDG tests.** `TestServiceStatusReportsAnUnreadablePlist` and `TestXDGServiceEnv_OnlyAbsoluteSet` exercise unix-only service integration and currently run on Windows. `//go:build !windows` rather than a skip, unless you want them asserting something Windows-specific?

**3. Sequencing.** The Windows leg cannot go green in one step — 171 failures across 26 packages is several PRs of work, and one PR carrying all of it would not be reviewable. Two ways to land it:

- **(a)** merge this PR with the Windows leg marked `continue-on-error: true`, land fixes in tranches, flip it to blocking when it is green;
- **(b)** hold this PR, land fixes guarded by the existing narrow selector step, merge this last and delete `build-windows`.

I would take **(a)**. It reports real numbers from day one, it does not ask anyone to maintain a selector list you have already said can miss the regression it is meant to protect, and the flip to blocking becomes a one-line PR at the end.

I have not measured bucket 2 on your runner. GitHub's windows image may well hold `SeCreateSymbolicLinkPrivilege`, in which case those 12 need nothing at all — the first red run under (a) settles it, which is another reason to prefer it.

#655 is the first tranche and is independent of whichever you pick.

tiendungdev:
Correction to my own numbers above, before they mislead anyone.

**The baseline I measured, `699f351e`, does not contain #646.** `7eae340e` (#646) landed immediately after it, and I fetched between the two without noticing. `git merge-base --is-ancestor 7eae340e 699f351e` says no.

Re-measured on current main `812eb2d8`:

| package | in my table above | actual, current main |
|---|---:|---:|
| `internal/mcp` | 27 | **24** |
| `internal/review` | 7 | **4** |

#646 closed 3 in each — the three `ReviewRulepack` / `PrefixedGraph` cases and the three `RankFileRisk` cases.

So the headline is **167 failing tests**, not 173, and **165** after subtracting the two that are my machine. The bucket shape is unchanged: the six I enumerated are unaffected, and the correction comes entirely out of bucket 7, which drops from ≈115 to **≈109**.

Everything else in that comment stands, including the three questions. #655 is unaffected — it was measured against `origin/main` at the time and its own before/after numbers were taken on the same tree.

pbednarcik:
@tiendungdev - if you need any help with Windows tests let me know.

PR #654, opened 2026-08-22
Add worktree and branch graph views

### LINKED ISSUE #663: Workspace id colliding in 2 different branches
**Is your feature request related to a problem?**
I have 2 different branches created and indexed same folder from both branches
It worked for first branch and failing for second branch

**Describe the solution you'd like**
Both index even if same folder name should work fine.
Worksapce id should be runtime generated and not the folder name

**Use case**
Create 2 branches
index same folder in both branches
RUn daemon on both
when query executed it gives error "Error: the gortex daemon does not track  <folder path> — add it with `gortex track <folder path>`"

Analysis:
Both workspace gets same id assigned as folder names are same. 
Either there should be way to provide workspace id as argument ot always generate new id

zzet:
## What this adds

Gortex previously indexed one corpus per tracked repository and mutated it in place on every branch switch, and a sibling git worktree either got a full duplicate index or nothing. This PR gives every checkout and ref its own coherent, queryable view:

- **Automatic worktree views.** A linked worktree of a tracked repository is discovered, gets a sparse commit layer plus a dirty layer over the family's primary corpus, and every MCP query from inside it sees exactly that worktree — committed state, uncommitted edits, and editor buffers, in that precedence. Switching a branch away and back reuses the cached layer with zero re-indexing.
- **Explicit ref and commit views.** A request can select a full ref name or exact commit for any ready dedicated graph and get a read-only committed-tree view, built locally with no network access, cached with LRU retention, and served with truthful resolved-identity riders. File reads route through the git object database under a view-scoped identity; a pruned object withdraws the file-read capability instead of misreporting.
- **Checkout lifecycle.** One component owns discovery, tracking intent, availability and removal clocks, and cleanup. Removal requires positive evidence — a validated inventory omission or a prunable record backed by same-volume path evidence; every ambiguous condition (permissions, IO, unmounted volumes, failed inventory) parks as inaccessible and never deletes. Cleanup runs as journaled, crash-resumable sagas guarded by incarnation and epoch tokens. Promotion, demotion, primary designation, and destructive untrack are previewed transactions with explicit confirmation. All checkout and worktree administration lives under `gortex repos` (`families`, `explain-view`, `forget`, `reconcile`, `set-primary`) — there is no `gortex checkouts` command; the reference is in `docs/cli.md`.

## How it works

**Storage.** The SQLite payload schema is re-keyed by a view generation (schema v14–v18): nodes and edges carry the generation as the trailing identity column so every existing index and query plan stays byte-identical, sidecars lead with it, and two partial indexes serve sparse enumeration. Every statement binds the handle's pinned generation or is explicitly repository-wide administration. A leak fence seeds divergent corpora at two generations and drives every store read method and capability interface, proving isolation; a disturb table proves writer families cannot touch a neighbor generation.

**Composition.** The in-memory overlay's verified precedence semantics were extracted into a layer contract; persisted generations implement it over ownership masks (file replace/delete, node tombstones, edge-source markers), and views nest bottom-to-top: base, commit layer, dirty layer, editor buffers. A materializer builds a checkout's stack from its route, holds leases so retirement cannot pull a generation out from under a reader, and unions producer completeness upward.

**Building.** Sparse generations are produced by the production parse/resolve pipeline running against a pinned handle over a content source (checkout filesystem behind an os.Root-confined reader, or a git tree behind a long-lived cat-file batch child that can never fetch). The build closure iterates to a fixed point: reverse dependents of the change, forward dependencies, references the change introduces, and manifests unconditionally; cap truncation surfaces as producer incompleteness rather than silent divergence. A property-based oracle generates mutation scripts and asserts the composed view equals a flat index of the same tree — nodes, both edge directions, files, counts, and resolver output — across randomized seeds plus soak volume.

**Serving.** A request resolves its view before handlers run: explicit selector over session binding over base. Sessions inside a worktree bind the family scope, search enumerates candidates from every generation in the stack with mask-then-dedup-then-interleave composition, text search answers from the checkout's own root, and per-checkout language servers run under a global cap. Operations evaluate their required capabilities against the served view's completeness and answer with typed errors or degradation annotations — base-scoped engine results under a non-base view are labeled, never passed off as view-scoped. Mutating tools refuse non-filesystem views. Hook probes and the control socket resolve views the same way, with fail-open preserved for undiscovered worktrees and daemon outages.

**Observability.** A bounded-cardinality metrics registry and log fields at every lifecycle seam make four questions answerable: which layers served a query, why a view is stale, what evidence classified a checkout, and why a generation still exists. The daemon status payload gains an additive views census.

## Compatibility

- All schema changes are additive in-place migrations (user_version 13→18); existing stores upgrade without re-indexing, and fresh versus migrated shapes are asserted equal.
- Wire changes are optional fields; old clients get byte-identical answers, pinned by goldens.
- Base-corpus request behavior is unchanged and pinned by goldens, including a zero-allocation assertion on the search fast path.
- Requires Go 1.27 (stdlib uuid, synctest-based virtual-clock tests); CI, release images, and golangci-lint 2.13.1 move together with the go.mod directive.

## Verification

Every round landed through an implement/adversarial-review/fix cycle. Gates on the final tree: full-module build, vet, golangci-lint clean; the complete test suite under `-race`; the equivalence oracle at default and soak volume; plan-fence tests proving no query regressed to a table scan; and the cross-generation leak fences. The 39 commits are ordered for review: storage first, then composition, builders, lifecycle, serving, and surfaces.

Closes #663


tiendungdev:
Not a review of the diff — 533 files is more than I can review usefully, and I would rather not spend your time on a low-quality pass. This is the one thing I can contribute that your CI structurally cannot tell you: what this branch does on Windows.

Measured at `2bd2582a`, windows/amd64, go1.27.0, `-count=1`.

## The Windows job does not cover any of the new code

`build-windows` is untouched by this PR, and its steps are `./internal/agents`, `./internal/pathguard`, `-run TestCloseSidecar ./internal/persistence`, the plan-lock selector, `./internal/testenv ./internal/platform`, `./internal/graphpath`, and the native-separator selector. **`internal/gitstate`, `internal/reconcile` and the new `cmd/gortex` checkout/view tests are in none of them.** The green check means `go build ./...` compiles on Windows — nothing more.

Running them gives **9 failures**:

```
internal/gitstate
  TestInventoryMainOnlyRepo
  TestInventoryLinkedWorktreesWithAwkwardPaths
  TestInventoryLockedWorktree
  TestInventoryMissingRootSurvivesThenPruneRemovesIt
  TestInventoryDetachedHEAD
  TestInventoryFromLinkedWorktreeSeesWholeFamily
  TestSamplePathEvidenceOnFilesystemRoot
  TestTwoPathsAndResolveAgainst
internal/reconcile
  TestReconcileRecordsHeadAndPathEvidence
```

## Those 9 are test-side, and I checked before saying so

They all look like the `/` vs `\` family I have been fixing, e.g.

```
inventory_test.go:134: no record for "D:\tmp\...\locked";
    inventory has "D:/tmp/.../repo", "D:/tmp/.../locked"
```

`git worktree list --porcelain` prints `/` on Windows and `WorktreeRecord.Path` keeps it verbatim, which is exactly what its doc-comment promises. So I went looking for a consumer that compares it against a native path, and did not find one:

| site | why it is safe |
|---|---|
| `checkout_lifecycle.go:2108` | `pathkey.EqualPaths` → `foldPath` → `filepath.Clean`, so both spellings fold equal |
| `graphview/binding.go:48,52` | cleans both sides before comparing |
| `checkout_lifecycle.go:966` `trackedByRoot` | a map keyed by path — the one shape that fails silently — but keys come from `r.absPath` (`multi.go:2071`) and lookups from `filepath.Abs`, both native |

So the contract holds and the failures are fixtures comparing git's spelling against `t.TempDir()`, plus a few POSIX absolute literals (`/repo` → `\repo`, losing the volume). Worth fixing so the packages *can* run on Windows, but not a defect in this branch.

## One real gap: path evidence is inert on Windows

`pathevidence_other.go:10` returns `VolumeKindUnsupported` with empty tokens, so `rootVolumeUsable()` is always false, and `reconcile/evidence.go:249` short-circuits:

```go
case !persisted.rootVolumeUsable():
    return "no usable persisted root volume token to compare a fresh sample against", false
```

`prunableConfirmed` therefore **can never return true on Windows**. The inventory-omission removal path still works, but the "git still lists it, marks it prunable, root is absent from a mounted volume" path does not — a worktree whose directory was deleted without a `git worktree prune` parks as `CheckoutInaccessible` forever and keeps its generation and catalog rows. Fail-closed, which is the right direction, but unbounded.

## And the fix is not a drop-in — the signature blocks it

Windows does have the equivalents, but not where `pathIdentity` looks. Measured:

```
Sys() concrete type: *syscall.Win32FileAttributeData
  volume serial from FileInfo? NO
  file index (inode analogue) from FileInfo? NO
from a HANDLE: volumeSerial=1214872080 fileIndex=6291456:164656
```

`fs.FileInfo.Sys()` on Windows carries attributes, times and size — no `dwVolumeSerialNumber`, no `nFileIndex`. Those come only from `GetFileInformationByHandle` on an open handle, which needs the **path**, not the `FileInfo`. So `pathIdentity(fs.FileInfo)` cannot be implemented on Windows as declared; it would need the path passed alongside (`FILE_FLAG_BACKUP_SEMANTICS` to open a directory).

That is one call site today, in a package that does not exist on main yet. It is much cheaper to widen now than after this lands.

## Offer

Happy to send `pathevidence_windows.go` plus that signature change as a PR against `feat/worktree-branch-views` rather than `main`, with the volume/identity semantics matching the unix pair. Say the word and I will; I have not started, because the branch is currently conflicting and I would rather build on whatever comes out of your rebase.

The 9 test fixtures I would rather do after you answer the sequencing question on #652 — whether they get a selector step or the full matrix changes what the fix should look like.

Which is also the wider point: this branch adds roughly five thousand lines of path- and git-shaped code, and the Windows job as it stands cannot execute one line of it.

Issue #450, opened 2026-08-04
bug: a bare npm specifier also binds to a same-repo directory of the same name, creating false import edges


chrisweirich:
**Version:** v0.62.0 (also reproduces on v0.61.2)

## Summary

When a repo contains a directory whose name equals a bare npm specifier, `import … from '<pkg>'` emits an `imports` edge to a file **inside that local directory**, *in addition to* the correct `external::<pkg>` edge. The importing file has no relationship to that local file.

This is worse than a missing edge: the graph asserts something false rather than admitting a gap, and it accumulates on the victim file. In a large private TypeScript repo, one test fixture that happens to live in a directory named `graphql/` collected ~200 phantom importers — every one of them merely `import … from 'graphql'` — so `dependents`/`usages` on that file is entirely wrong.

## Minimal reproduction

```
package.json                              deps: graphql, zod, lodash  (node_modules NOT installed)
src/consumer-graphql.ts                   import { parse } from 'graphql'
src/feature/graphql/create.mutation.gql   (any file)
src/consumer-zod.ts                       import { z } from 'zod'
src/other/zod/helper.ts                   (any file)
src/consumer-lodash.ts                    import { merge } from 'lodash'   # no matching directory
```

No `tsconfig.json`. Index the repo, then list the `imports` edges per consumer file:

| from | `imports` edges |
|---|---|
| `consumer-graphql.ts` | `external::graphql::parse` **and** `src/feature/graphql/create.mutation.gql` |
| `consumer-zod.ts` | `external::zod::z` **and** `src/other/zod/helper.ts` |
| `consumer-lodash.ts` | `external::lodash`, `external::lodash::merge` — nothing else |

**Expected:** only the `external::` edges, as in the `lodash` case.
**Actual:** the two colliding specifiers additionally bind to an unrelated local file.

The `lodash` control isolates the trigger to the directory-name collision. It is independent of the target's file type (`.gql` and `.ts` both reproduce) and of directory depth.

## Root cause

All in `internal/resolver/resolver.go`, `resolveImport`:

1. `buildDirIndexes` (:1756) keys `lastDirIndex` on the **last path component of each file's directory** — by design; the comment reads *"so an import of `logger` matches any file under .../logger/"*.
2. When `dirIndex[importPath]` misses, the cascade walks `lastDirIndex[lastPathComponent(importPath)]` (:3310). For a bare specifier `graphql` that is every indexed file under any `*/graphql/` directory.
[...]

PR #665, opened 2026-08-25
fix(resolver): resolve relative, alias and component imports from SFC <script> blocks

ishpagin:
## Summary

Verified on a real project: [mutoe/vue3-realworld-example-app](https://github.com/mutoe/vue3-realworld-example-app), indexed with the daemon (123 files, 1328 nodes, 4264 edges).

`src/components/ArticlesList.vue` imports one composable through a tsconfig alias and three sibling components:

import { useArticles } from 'src/composable/use-articles'
import AppPagination from './AppPagination.vue'
import ArticlesListArticlePreview from './ArticlesListArticlePreview.vue'
import ArticlesListNavigation from './ArticlesListNavigation.vue'

Before: all four were external stubs, `find_usages useArticles` returned nothing.

After: the same file's import edges in the graph:

src/components/ArticlesList.vue -> src/composable/use-articles.ts
src/components/ArticlesList.vue -> src/composable/use-articles.ts::useArticles
src/components/ArticlesList.vue -> src/components/AppPagination.vue
src/components/ArticlesList.vue -> src/components/ArticlesListArticlePreview.vue
src/components/ArticlesList.vue -> src/components/ArticlesListNavigation.vue

and `find_usages useArticles` lists `ArticlesList.vue:39`.

## Changes

- `isJSTSPath` accepts `.vue` / `.svelte` / `.astro`. The predicate gates six sites (`resolveJSTSImportTarget`, `isJSTSBareSpecifier`, `pickImportEvidenceCallee`, the top-level value-callee pick, `importedDirForSpec` and the cross-repo `resolveImport` mirror); all of them reason about "the importing file has JS/TS module semantics", which an SFC script does. Deliberate side effect: bare specifiers from an SFC (`'vue'`, `'pinia'`) now pass the entry-point gate of #450 instead of `bare=false` letting the directory cascade bind them to an arbitrary same-named local directory.
- `probeJSTSFile` accepts an explicit `.vue` / `.svelte` / `.astro` specifier verbatim, so `import Foo from './Foo.vue'` binds to the SFC file node. No extension probing on top of it: `./Foo` never picks up `Foo.vue`.
- Tests: `TestIsJSTSPathAcceptsSFC`, `TestResolveJSTSImportTargetFromVueScript`, `TestProbeJSTSFileAcceptsExplicitSFC` (unit), and `internal/mcp/find_usages_sfc_test.go` end to end through the real indexer and the `find_usages` handler (`.vue` calling a `.ts` composable, `.vue` importing `.vue`).

## Testing

- [ ] Benchmarks run if performance-relevant (not relevant: one extra `case` in two `switch` statements)

## Checklist

- [ ] Language extractor includes `Meta["methods"]` for interfaces (n/a, resolver-only change)
- [ ] Methods have `EdgeMemberOf` edges to their containing type (n/a)

Follow-up in a stacked PR: cross-directory single-candidate calls from JS/TS (including SFC scripts) still land at `text_matched` and are hidden by `find_usages` by default; that is a separate resolver bug, fixed on `fix/jsts-single-candidate-import-evidence`.

### DIFF
--- a/internal/mcp/find_usages_sfc_test.go
+++ b/internal/mcp/find_usages_sfc_test.go
@@ -1,0 +1,61 @@
+package mcp
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/zzet/gortex/internal/config"
+	"github.com/zzet/gortex/internal/graph"
+	"github.com/zzet/gortex/internal/indexer"
+	"github.com/zzet/gortex/internal/query"
+	"go.uber.org/zap"
+)
+
+// A .vue <script setup> that imports a .ts composable and calls it must show
+// up in find_usages on the .ts symbol, and a component-to-component
+// `import Foo from './Foo.vue'` must bind to the SFC file, not an external
+// stub. End-to-end through the real indexer + resolver + handler.
+func setupSFCServer(t *testing.T) *Server {
+	t.Helper()
+	dir := t.TempDir()
+	write := func(name, body string) {
+		require.NoError(t, os.WriteFile(filepath.Join(dir, name), []byte(body), 0o644))
+	}
+	write("useX.ts", "export function useX() { return 1 }\n")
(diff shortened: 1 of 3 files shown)


zzet:
Hey @ishpagin, thank you for the fix!

PR #675, opened 2026-08-26
Replace regex extractor with tree-sitter for Julia

NilsWildt:
## Summary

Replaces the regex-based Julia extractor with a tree-sitter extractor, upgrading .jl files from shallow line-scan extraction to full concrete-syntax-tree analysis. Bumps the extractor version to 2 so stale regex-era extractions are re-indexed.

## Changes

- internal/parser/languages/julia.go — rewritten around tree-sitter-julia (go-sitter-forest):
  - Modules (module/baremodule) index as KindType carrying Meta["exports"]; nested definitions get member_of edges
  - Types: abstract/primitive/mutable/parametric structs; fields as KindField + member_of; supertypes emit extends edges with the full parametric RHS preserved in Meta["base_path"] (python-extractor convention)
  - Functions/macros: long-form, short-form (f(x) = body, incl. where clauses), unicode/bang identifiers, operator methods (Base.:+) as KindMethod with Meta["receiver"] + EdgeMemberOf (luau convention), macros flagged via Meta["macro"]
  - Imports: all four forms (using M, using M: a, b → names meta, import M as Alias → alias meta, dotted paths); include("f.jl") still emits an imports edge, preserving the legacy extractor's contract
  - Calls: bare, qualified (MyMod.helper), broadcast (sin.(xs) → broadcast meta), and macro invocations (@time f(x) emits both macro and inner call edges); direct self-recursion suppressed
  - Preceding and body docstrings attach as Meta["doc"]
- internal/parser/languages/julia_test.go — 8 scenarios covering modules/exports, type & field extraction, definition forms, imports, calls (incl. closure attribution + recursion suppression), docstrings, and empty input
- internal/indexer/extractor_version.go — "julia": 2 version bump + .jl salt entry so existing .jl graph entries re-extract
- docs/languages.md — Julia promoted into the bespoke tree-sitter tier; new "Julia specifics" section documenting node kinds and metadata conventions

## Testing

- All tests pass (go test -race ./...) — branch-scoped packages green (internal/indexer, merkle, internal/parser/languages); two unrelated packages flake only under full-suite parallel load and pass on isolated rerun
- New tests added for new functionality
- Benchmarks run if performance-relevant (N/A — extraction-correctness change, no hot-path algorithm changes)

## Checklist

- Code follows existing patterns in the codebase (mirrors python base_path and luau receiver conventions already documented in docs/languages.md)
- No unnecessary abstractions added (only two small internal state structs: juliaScope, juliaWalkState)
- Language extractor includes Meta["methods"] for interfaces (N/A — Julia has no interface construct)
- Methods have EdgeMemberOf edges to their containing type (qualified methods → receiver type; struct fields and module members likewise)

### DIFF
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -11,7 +11,7 @@ Three engine tiers are used, in order of decreasing extraction depth:
   symbols, resolved call edges, ORM/contract/dataflow extraction, and accurate
   node ranges. Languages: Go, TypeScript, JavaScript, Python, Rust, Java, C#,
   Kotlin, Swift, Scala, PHP, Ruby, Elixir, C, C++, Dart, OCaml, Lua, Bash, SQL,
-  HTML, CSS, Markdown, OrgMode, Protobuf, YAML, TOML, HCL, Dockerfile.
+  HTML, CSS, Markdown, OrgMode, Protobuf, YAML, TOML, HCL, Dockerfile, Julia.
 - **regex** (~60 languages) — pattern-matched line scanning with indent / brace /
   keyword block heuristics. Captures top-level symbols and imports; call edges
   vary per language. Used where no upstream tree-sitter grammar is available
@@ -84,6 +84,7 @@ on interface nodes stores the expected method set for implementation matching.
 | Dart | Full | Full | Classes/Enums/Mixins/Extensions | Abstract interface | Full | Full | Full |
 | OCaml | Full | Full (class) | Types/Modules | Module types | open | Full | Full |
 | Lua | Full | Full (M.func/M:method) | - | - | require() | Full | Full |
+| Julia | Full (long + short form, `where` syntax) | Full (qualified `Base.show`, operators) | Structs/abstract/primitive + fields | - | Full (`using`/`import`/`include`, selective lists) | Full (incl. broadcast, macro calls) | `const` |
 
 ### Rust specifics
 
@@ -107,6 +108,27 @@ exports (`#[no_mangle]`, `#[wasm_bindgen]`, pyo3, napi). Declarations inside an
 language, so a missing definition is not a missing implementation.
 
 Recent extraction refinements (each covered by a per-feature CI golden): Java `@interface` annotation types index as interfaces and participate in implementation matching; Java `new T(){…}` and C# `new { … }` anonymous classes/types become synthetic type nodes with an `extends` edge (to the instantiated type, or to `object` for C#); JS/TS arrow-valued class fields (`x = () => {…}`) are emitted as callable methods; JS/TS named imports emit one `imports` edge per binding (alias-aware via `Edge.Alias`) and barrel re-exports emit `re_exports` edges; JS/TS cross-file imports resolve onto the target file/symbol for relative specifiers (`./x`, `../x`) and for `tsconfig.json` / `jsconfig.json` `compilerOptions.paths` / `baseUrl` path aliases (`@/lib/x`), so callers / usages / blast-radius span aliased imports the same as relative ones; chained / factory-call receivers (`New().Build()`) carry an inferred `receiver_type`. See [features.md](features.md) and [architecture.md](architecture.md) for the edge semantics.
+
+### Julia specifics
+
+`module` / `baremodule` index as `KindType` nodes (the graph's `KindModule`
+is reserved for ecosystem packages) and carry the module's `export` list in
+`Meta["exports"]`; definitions inside a module get `member_of` edges to it.
+Qualified method definitions (`function Base.show(...)`, `function Base.:+`)
+become `KindMethod` with `Meta["receiver"]`, mirroring the Lua `M.func`
+convention. Struct fields are `KindField` nodes with `member_of` into the
+struct; supertypes (`struct X <: Y`) emit `extends` edges with the full
+right-hand path in `Meta["base_path"]` (python extractor convention).
+
+`using M: a, b` / `import M as Alias` keep the module as the import target
+with the selected names / alias in the edge `Meta` for the resolver; all
+imports including `include("file.jl")` target `unresolved::import::<path>`.
+Calls attribute to the enclosing function-like definition (long form, short
+form, macro, or nested closure) and cover qualified (`Mod.f`) and broadcast
+(`f.(x)`, `Meta["broadcast"]`) callees plus macro invocations
+(`Meta["macro"]`). Operator calls (`a + b`) are not emitted — dispatch on
+operators is not statically attributable. Docstrings (preceding string
+literal or first body statement) attach as `Meta["doc"]`.
 
 ## Data, config, build
 
@@ -188,7 +210,6 @@ What is **not** covered:
 
 | Language | Extensions | What it extracts |
 |----------|------------|------------------|
-| Julia | `.jl` | `function`, `struct`, `module`, `using` / `import` |
 | R | `.r`, `.R` | Function defs; `library` / `require` / `source` |
 | MATLAB | `.mlx` | `function` (end-terminated), `classdef`, `import a.b.c` |
 | Mathematica | `.wl`, `.wls`, `.nb` | `name[args_] := body`, `SetDelayed`, `Get[…]` / `Needs[…]` |
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -120,9 +120,17 @@ convention. Struct fields are `KindField` nodes with `member_of` into the
 struct; supertypes (`struct X <: Y`) emit `extends` edges with the full
 right-hand path in `Meta["base_path"]` (python extractor convention).
 
-`using M: a, b` / `import M as Alias` keep the module as the import target
-with the selected names / alias in the edge `Meta` for the resolver; all
-imports including `include("file.jl")` target `unresolved::import::<path>`.
+`using M: a, b` / `import M as Alias` keep the module as the import target,
+including when the path is dotted or relative (`using A.B: x`, `import
+..Up: q`). A selective list additionally emits one edge per binding to
+`unresolved::import::<module>::<name>` — the per-binding shape JS/TS
+already uses — and a rename rides on `Edge.Alias`, with the module edge
+keeping it in `Meta` too because the SQLite edges table has no alias
+column. A module alias is applied at extraction time, so `import Foo as F`
+followed by `F.process(x)` records a call to `Foo.process`: the same target
+the unaliased spelling produces. Nothing in the resolver reads the import
+`Meta` — the edge targets are the consumable surface. All imports including
+`include("file.jl")` target `unresolved::import::<path>`.
 Calls attribute to the enclosing function-like definition (long form, short
 form, macro, or nested closure) and cover qualified (`Mod.f`) and broadcast
 (`f.(x)`, `Meta["broadcast"]`) callees plus macro invocations
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -114,10 +114,16 @@ Recent extraction refinements (each covered by a per-feature CI golden): Java `@
 `module` / `baremodule` index as `KindType` nodes (the graph's `KindModule`
 is reserved for ecosystem packages) and carry the module's `export` list in
 `Meta["exports"]`; definitions inside a module get `member_of` edges to it.
-Qualified method definitions (`function Base.show(...)`, `function Base.:+`)
-become `KindMethod` with `Meta["receiver"]`, mirroring the Lua `M.func`
-convention. Struct fields are `KindField` nodes with `member_of` into the
-struct; supertypes (`struct X <: Y`) emit `extends` edges with the full
+Node ids stay flat — the enclosing module rides on `Meta["scope_mod"]`, the
+Rust `mod` convention — and two definitions that would collide on one id
+(`f` in module `A` and `f` in module `B`) separate through the shared
+line-suffix helper. Qualified method definitions (`function Base.show(...)`,
+`function Base.:+`) become `KindMethod` with `Meta["receiver"]`, mirroring
+the Lua `M.func` convention. Julia has no constructor keyword, so a callable
+named after a type is that type's constructor and takes the cross-language
+`<Type>.<init>` spelling Java, C# and Swift already emit — outer, short and
+inner forms alike. Struct fields are `KindField` nodes with `member_of` into
+the struct; supertypes (`struct X <: Y`) emit `extends` edges with the full
 right-hand path in `Meta["base_path"]` (python extractor convention).
 
 `using M: a, b` / `import M as Alias` keep the module as the import target,
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -130,9 +130,10 @@ right-hand path in `Meta["base_path"]` (python extractor convention).
 including when the path is dotted or relative (`using A.B: x`, `import
 ..Up: q`). A selective list additionally emits one edge per binding to
 `unresolved::import::<module>::<name>` — the per-binding shape JS/TS
-already uses — and a rename rides on `Edge.Alias`, with the module edge
-keeping it in `Meta` too because the SQLite edges table has no alias
-column. A module alias is applied at extraction time, so `import Foo as F`
+already uses, bounded by the same cap so one statement cannot dwarf a
+file's graph — and a rename rides on `Edge.Alias` plus `Meta["alias"]`,
+because the SQLite edges table has no alias column and Meta is the half
+that survives the round-trip. A module alias is applied at extraction time, so `import Foo as F`
 followed by `F.process(x)` records a call to `Foo.process`: the same target
 the unaliased spelling produces. Nothing in the resolver reads the import
 `Meta` — the edge targets are the consumable surface. All imports including
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -141,8 +141,14 @@ Calls attribute to the enclosing function-like definition (long form, short
 form, macro, or nested closure) and cover qualified (`Mod.f`) and broadcast
 (`f.(x)`, `Meta["broadcast"]`) callees plus macro invocations
 (`Meta["macro"]`). Operator calls (`a + b`) are not emitted — dispatch on
-operators is not statically attributable. Docstrings (preceding string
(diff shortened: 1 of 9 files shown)


zzet (changes requested):
Nils, thank you for your first contribution to Gortex — and especially for taking on something this substantial. Replacing a language extractor is ambitious work, and the tree-sitter direction, the versioning intent, the documentation, and the breadth of the initial tests are all valuable.

I’m requesting changes because a second adversarial pass found several correctness gaps that the current green CI cases do not exercise. A few findings share the same underlying AST/scope assumptions, so this looks fixable as a focused follow-up rather than a rejection of the approach.

I’ve included a reproducer and a proposed correction for every finding below.

### 1. P1 — Make the Julia version bump work for existing non-Merkle indexes

`internal/indexer/extractor_version.go:91`

A previous release snapshot cannot contain `julia`, because `.jl` was not in `extractorSaltExtLang`. `staleLangsBetween` iterates only keys in the stored snapshot, so the new `julia@2` key is never considered stale in the default non-Merkle/mtime path. Unchanged repositories therefore keep their regex-era Julia graph.

A minimal regression test is:

~~~go
func TestNewVersionedLanguageIsStaleFromOlderSnapshot(t *testing.T) {
    previous := extractorVersionsSnapshot()
    delete(previous, "julia") // shape persisted by the previous release

    stale := staleLangsBetween(previous, extractorVersionsSnapshot())
    require.Contains(t, stale, "julia")
}
~~~

Run with:

~~~text
go test ./internal/indexer -run TestNewVersionedLanguageIsStaleFromOlderSnapshot -count=1
~~~

It fails at this head. The end-to-end incremental path likewise returns `StaleFileCount == 0`.

Proposed solution: for a non-empty stored snapshot, iterate current versioned languages and treat a missing stored key as baseline version `1`; flag it only when the current version is greater than `1`. Preserve the existing no-baseline behavior for an empty/absent snapshot. Please also add a non-Merkle incremental reindex test, since Merkle mode already catches the changed salt but is not the default path.

### 2. P1 — Unwrap return-type expressions before requiring a call signature

`internal/parser/languages/julia.go:393`

Reproducer:

~~~julia
function long_form(x)::Int
    helper(x)
end

short_form(x)::T where T = helper(x)
~~~

At this head extraction returns only the file node: both functions and the `helper` call disappear. The base regex extractor still emitted `long_form`, so the long form is a direct regression. The grammar nests the call as `signature -> where_expression/typed_expression -> call_expression`, but `juliaUnwrapSignature` and `handleAssignment` unwrap only `where_expression` and then require a direct call.

Proposed solution: recursively peel `signature`, `where_expression`, and `typed_expression` until reaching `call_expression` (or a non-wrapper), and use the same helper for long and short definitions. Add tests for `function f(x)::Int`, `f(x)::Int = ...`, and `f(x)::T where T = ...`, asserting both the definition and body call edges.

### 3. P1 — Preserve `import_path` in combined selective/aliased imports

`internal/parser/languages/julia.go:589`

Reproducer:

~~~julia
using A.B: x, y
import C.D as CD
using .Local: z
import .Local as L
~~~

Current output targets `x`, `CD`, `z`, and `L`. The expected import targets are `A.B`, `C.D`, `.Local`, and `.Local`; the selected names/aliases should be metadata on those imports. The base extractor at least retained `A.B`/`C.D`. In these combined forms the module is an `import_path` child, which the identifier-only loops ignore.

Proposed solution: decode `import_path` as the target inside `selected_import` and `import_alias`, then collect the remaining identifier/operator children as selected bindings or alias. Add cross-product tests for dotted + selective, dotted + alias, and relative + both forms; the current tests cover each feature only in isolation.

### 4. P2 — Store import bindings in a contract the resolver consumes

`internal/parser/languages/julia.go:580`

Reproducer:

~~~julia
import Foo as F
run(x) = F.process(x)
~~~

The extracted import has `Meta["alias"] == "F"`, but `Edge.Alias` is empty. Repository-wide consumer tracing found no resolver reader for `Meta["alias"]` or `Meta["names"]`; `Resolver.resolveImport` consumes the import target, while `graph.Edge.Alias` is the canonical persisted alias field. Consequently an extractor-only test of the Meta map passes, but resolution cannot map `F.process` back to `Foo`. Selected names are similarly inert.

Proposed solution: populate the canonical `Edge.Alias` representation for aliases. For selective imports, either emit binding-aware import edges in the representation the resolver already understands or add explicit Julia selected-name handling. Add an integration test that runs extraction through resolution and asserts the aliased/selected call resolves, rather than only checking the raw Meta map. If resolution is intentionally out of scope, please narrow the resolver claim in the docs.

### 5. P1 — Include lexical module/type scope in IDs and model constructors separately

`internal/parser/languages/julia.go:417`

Two reproductions expose the same flat-ID problem:

~~~julia
module A
f() = from_a()
end

module B
f() = from_b()
end
~~~

This produces one `file::f` node, marks it `member_of A`, and attributes both calls to it; `B.f` is absent.

~~~julia
struct Box
    x::Int
end

Box(x) = make_box(x)
~~~

This produces only the `KindType` node `file::Box`. The constructor is suppressed by `seen`, and the `make_box` call is emitted from the type node. Long and inner constructors fail the same way.

Proposed solution: carry a qualified lexical scope, including nested modules and the current type, in `juliaScope`; use it when constructing module/type/function IDs. Give constructors a distinct callable/method identity and an `EdgeMemberOf` to their type rather than reusing the type ID. Add assertions for same-named functions in separate modules plus outer, short, and inner constructors.

### 6. P2 — Extract assignment-shaped fields used by `Base.@kwdef`

`internal/parser/languages/julia.go:320`

Reproducer:

~~~julia
Base.@kwdef struct Config
    host::String = "localhost"
    port::Int = 8080
end
~~~

Extraction emits `Config` but neither field. These direct struct children are `assignment` nodes whose LHS is a `typed_expression`; the field scanner currently handles only direct `typed_expression` and `identifier` children.

Proposed solution: when a direct struct child is an assignment, unwrap its LHS and feed that through the existing typed/bare field-name logic. Keep the check limited to direct struct members so ordinary assignments inside nested methods are not fields. Add a `Base.@kwdef` test asserting both `KindField` nodes and their `member_of` edges.

### 7. P2 — Follow Julia’s preceding-and-adjacent docstring semantics

`internal/parser/languages/julia.go:467`

These two cases currently go in opposite wrong directions:

~~~julia
"""short-form docs"""
short(x) = x       # doc is lost

function value()
    "runtime string value"  # incorrectly stored as Meta["doc"]
end
~~~

A string separated from a long definition by a blank line is also attached even though it is not adjacent. Julia’s documentation syntax attaches a string immediately before the documented object; a first body expression is ordinary executable code. `pendingDoc` is not passed into `handleAssignment`, while the body-specific block creates the false positive.

Proposed solution: remove first-body-string inference, pass the pending doc into short-form definition emission, and check source-line adjacency before attaching it. Add tests for a valid long doc, valid short doc, blank-line separation, an intervening comment, and a function whose first/return expression is a string.

### 8. P2 — Cover the advertised “Full” forms or narrow the documentation

`docs/languages.md:85`

Additional valid inputs currently produce incomplete graphs:

~~~julia
function calls(xs)
    Vector{Int}(xs)       # no call edge
    Base.@time helper(xs) # helper is seen, qualified macro call is not
end

module M
export @m, ⊗             # neither export is recorded
const X = 1              # no member_of M
end
~~~

Parametric callees are `parametrized_type_expression`, qualified macro identifiers are nested under a field expression, and export handling accepts only identifiers. Nested modules/constants also do not receive the documented containment edge.

Proposed solution: extend callee decoding for parametric expressions; decode qualified macro identifiers; accept macro/operator export nodes; and attach `member_of` for constants/nested modules. Add one focused test for each AST shape. If these are intentionally deferred, please change the “Full” cells and Julia-specific prose so users are not promised coverage the extractor does not provide yet.

### 9. P3 — Remove or use the newly introduced dead data flow

`internal/parser/languages/julia.go:224`

`juliaNameAndParams` collects and returns type parameters, but its only consumer discards them with `name, super, _ := juliaTypeHeadInfo(...)`. `emitShortFunction` also receives `sig` and `src` without reading either parameter. The compiler permits unused parameters, so CI does not flag them.

Proposed solution: if type parameters are part of the intended graph contract, persist them in node metadata and assert them in the parametric-type tests. Otherwise simplify the helpers to return only the used name/supertype data. Remove the unused `sig`/`src` parameters from `emitShortFunction` unless signature metadata is implemented in this PR.

Again, thank you for putting this together. The core direction is good, and I don’t want the size of this review to be discouraging — this is a large extractor surface, and these issues came from deliberately adversarial combinations beyond the current suite.

If you’d like, I can prepare the follow-up patch and push it directly to this PR branch once you confirm that’s okay and maintainer edits are enabled. Alternatively, I’m happy to handle only the index migration and scope/constructor pieces while you continue with the Julia AST cases.

NilsWildt:
Hey Andrew, thank you for this careful review! I'm very glad you caught the possible regressions and especially the non-covered migration path. I've enabled maintainer edits and would be delighted to get a follow-up patch from you. I'm actually not sure how best to handle the AST part, so I'll wait for your patch and then try to fix '8. P2 — Cover the advertised "Full" forms or narrow the documentation afterwards. I might maybe ask in the julia slack for some brainstorm help what would be best in users opinion. Cheers!

zzet:
Nils — the follow-up patch is on the branch, 21 commits. **Finding 8 is untouched and yours**, as you asked; there are notes at the bottom on what I ran into next door so it costs you less.

Two things first, because they matter more than the patch: **two of my own findings were wrong**, and I corrected them rather than implement them as written. I also broke three things while fixing the rest and had to go back for them — those are called out too, since they're the parts most worth your eye.

## What landed

| # | Commits | |
|---|---|---|
| 1 | `indexer: flag a newly tracked extractor language as stale`, `keep the staleness comparison inert without a baseline` | as proposed, plus a consequence I had to fix — see below |
| 2 | `parser(julia): recover definitions whose head is not a bare call` | + `function f end` |
| 3 | `parser(julia): keep the module as the target of a combined import` | + operator selections, `import Foo: bar as baz` |
| 4 | `parser(julia): make an import alias reach the call it renames` | **premise corrected**; later fixed again for module scope, cap, alias persistence |
| 5 | `parser(julia): separate colliding definitions and model constructors` | **deviates from my proposal**; later fixed again for scope, macros, forward declarations, delegation |
| 6 | `parser(julia): extract struct fields declared with a default` | as proposed |
| 7 | `parser(julia): follow Julia's own docstring attachment rule` | as proposed, + 3 follow-up commits |
| 9 | `parser(julia): drop the collected-but-discarded type parameters` | as proposed |

## Where I was wrong

### Finding 4 — I had the persistence backwards

I wrote that `graph.Edge.Alias` is "the canonical persisted alias field". It is canonical, but it is **not persisted**: the `edges` table has no `alias` column — not in the DDL, not in `edgeGeneratedColumns`, not in `edgePromotedColumns` — `insertEdgeLocked` never binds it and `scanEdgeWithMeta` never sets it. `Edge.Meta` is the durable half. So setting `Edge.Alias`, as I asked, would have changed nothing on its own, and `Meta["alias"]` / `Meta["names"]` have no reader either, which was the real complaint.

What actually makes `F.process(x)` reach `Foo` is the trick the Python extractor already uses: resolve the alias **at extraction time**. The patch collects module aliases in a pre-pass and rewrites the qualified callee, so the aliased and unaliased spellings produce an identical graph. `TestJuliaImports_AliasedCallMatchesUnaliasedCall` drives real extraction through `ResolveAll` and asserts exactly that, instead of inspecting a Meta map. Selective imports also emit one edge per binding, JS/TS style — with the same cap that convention carries, and with the rename on both `Edge.Alias` and Meta so it survives the store.

One thing that did **not** improve: `unresolved::Foo.process` still doesn't bind to the definition. I checked, and the *unaliased* spelling has the same gap, so it isn't alias-specific — it's resolver-side. I narrowed the docs claim rather than pretend otherwise, which was the alternative my own review offered.

### Finding 5 — module-qualified ids would have broken owner derivation

I asked you to fold the module path into node ids. That is wrong here. `graph.EnclosingFromID` recovers a method's or field's owner by cutting the id at the **last dot**, so `file::Outer.Inner.f` would report `Outer.Inner` as the owning type. Every extractor in the tree keeps flat ids and puts the scope on Meta — `scope_ns` in C#/C++/PHP, `scope_mod` in Rust, whose `mod` handling documents the rule for exactly this shape.

So the patch keeps flat ids, stamps `Meta["scope_mod"]`, and separates genuine collisions with the shared `disambiguateID` helper (`f` and `f_L10`) — the treatment `TestRsExtractor_SameNameAcrossModulesBothSurvive` already pins for Rust. Constructors take the cross-language `<Type>.<init>` spelling Java, C# and Swift emit.

I checked the obvious risk — does `<init>` cost a call binding? — by running a two-file fixture through `ResolveAll` at this branch's head and at the patch. `Box(v)` at the call site was **already** `unresolved::Box`, so nothing is lost. What changes is that `make_box` is now attributed to the constructor instead of to `struct Box` itself, which is what the finding was about.

### Finding 7 — adjacency confirmed, with one thing I under-specified

Your instinct was right, and I verified it against the parser rather than assuming: Julia's lexer allows exactly one newline between a docstring and the documented object, and the manual says twice that no blank line or comment may intervene.

Two refinements. A *trailing* comment on the docstring's own line, or a block comment opening there, does **not** detach it — comment rows are discounted from the distance rather than breaking it. And removing the body-docstring inference isn't enough on its own: the walk recurses into bodies, so a string at the top of one still attached to a definition nested below. Docstrings are now gated to the blocks that can hold one, and `begin`/`quote` count only outside a definition — `quote ... end` is how nearly every macro body is written.

## Three things I broke and went back for

These came out of a second adversarial pass over my own diff, and they are the parts I'd most want a second opinion on:

- **The staleness fix put a permanent banner on every file read, in every repo.** Comparing against the baseline every language implicitly carries means julia reads as behind in *any* repo indexed by an older binary — a Go-only one included — so the MCP freshness rider attached `extractor_stale_langs:["julia"]` to every `read_file` / `get_symbol_source` / `get_editing_context` response and told the reader to reindex a language they have no file of. The advisory is now scoped to the language of the file being described, which is already how the hint beside it was gated.
- **The alias map was file-global.** `import Plots as P` in one module rewrote `P.plot(x)` in a sibling module that had bound `P` to something else. Aliases are keyed by their module now. The same commit fixes the constructor lookup, which walked *out* of a submodule — Julia submodules inherit nothing from their parent, so that was inventing a `member_of` into an unrelated type.
- **A definition's own signature was being counted as a call.** The head is a `call_expression` in this grammar, so every definition emitted a call edge to itself; the recursion guard had been swallowing it invisibly, and the moment constructors stopped applying that guard, `Rep(n) = Rep(n, false)` reported two delegations. Definitions now walk the signature's children (so a call in a default argument survives) without treating the head as a call site.

Also fixed while in there: an indented docstring lost its summary because the signature-block test ran before dedenting; a docstring above `Base.@kwdef struct` was dropped at the macro boundary; selected macros (`using Base: @time`) and renamed operators (`import Base: + as plus`) vanished from import lists; a macro sharing a type's name became that type's constructor; a constructor written *above* its struct stole the type's canonical id; and a same-line twin (`g(x) = h(x); g(y) = k(y)`) lost its call edges.

## Verification

- `go test -race` green on all four touched packages (`internal/parser/languages`, `internal/resolver`, `internal/indexer`, `internal/mcp`).
- Full `go test ./...`: 11319 passed across 185 packages.
- `golangci-lint` clean on the touched packages.
- Every new test was checked by reverting the production change and confirming it fails — including the three the review found couldn't fail, which are now reordered or strengthened so they can.

## Left for you (finding 8), with notes

- `Vector{Int}(xs)` — the callee is a `parametrized_type_expression`; `juliaCalleeName` has no case for it.
- `Base.@time f(x)` — the macro identifier nests under a `field_expression`, and `handleMacroCall` matches only a direct `macro_identifier` child.
- `export @m, ⊗` — `macro_identifier` and `operator` nodes; the export scan takes identifiers only. (Import lists now handle both, so there's a pattern to copy.)
- `const X = 1` and a nested `module Inner` still get no `member_of` to their enclosing module. They now carry `Meta["scope_mod"]`, so if you'd rather record containment that way than with an edge, half of it is already there.
- One that is a bug rather than coverage, in the same function: `juliaCalleeName` splits `n.Content(src)` on the last `.`, so `get(cfg).run(x)` emits the edge target `unresolved::get(cfg).run`, and a chain broken across lines emits a target containing a newline. Decoding the `field_expression`'s children instead of its text fixes both. Left to you since it's the function your work will touch — say the word if you'd rather I took it.
- `Base.:(==)` currently yields the name `(==)` while `Base.:+` yields `+`; same decoder.

Before touching the docs cell: `docs/languages.md` has no legend, and the house style for bounded coverage is an enumeration in the cell plus an explicit exclusion sentence in the language section — the Rust row is the model.

## Follow-ups I deliberately did not take

- `Edge.Alias` is dropped at the SQLite boundary for **every** language, not just Julia; JS/TS re-export barrels depend on it.
- The non-Merkle extractor-staleness gate is only consulted on a full-root incremental pass, which the daemon's warm-start route never takes — a clean repo short-circuits, a churned one goes scoped. Pre-existing for every bumped language (`php@2`, `csharp@12`, `go@3`, …), so widening it belongs in its own change rather than riding in on the Julia bump. Same for the re-stamp being blocked while any file in the repo fails to parse.
- `persistExtractorVersion` can in principle write a one-key row, which now reads as "every other language is behind". Unreachable today — the only writers of that row marshal the full snapshot — so I left it rather than invent behaviour for a state nothing produces.
- Julia qualified calls into another file's module don't bind, and constructor call sites don't bind to `<Type>.<init>` (Java gets that by spelling the call site too). Both resolver-side.
- Two same-named methods on one physical line still collapse to one node — the shared line-suffix helper can't separate them. Their call edges are preserved; the second node is not.
- Still dropped: `const X::Int = 1`, parametric constructors `Box{T}(x) where T = …`, callable objects `(f::Box)(x) = …`, and calls inside string interpolation.

Thanks again for this — it's a large and genuinely useful piece of work, and the tree-sitter direction was right from the start.

zzet (approved):

PR #678, opened 2026-08-26
graph: file-scoped evictions describe themselves to mutation receipts exactly

pbednarcik:
Every EvictFile and EvictFiles call voids any active mutation receipt, even
though a file-scoped eviction knows its exact node set. The structural
incremental batch commits every reindex through evict-then-readd, so every
enrichment apply and watcher reindex poisons its receipt and
resolveDeferredMutations takes the fail-closed whole-graph fallback. On my
production store that fallback admits the full pending set to convert a
sliver of it: the post-enrichment pass scanned 1.1M pending edges to
produce 42.7k conversions (a 3.9 percent hit rate, about 64 seconds), with
the daemon's own incomplete_reason pointing at file_batch_evict.go. The
receipt-exact path that resolves only the touched files already exists;
it just never ran.

The fix records the eviction to active receipts instead of voiding them.
An evicted referenceable definition is resolution-relevant the same way an
added one is: pending references naming it elsewhere may resolve
differently once it is gone, and in the evict-then-readd flow the re-add
records the successor identity. So the doomed nodes' files join the
definition frontier and their names join the target set, mirroring how an
identity-changing upsert already records a vanished old identity while
staying complete.

Backend mechanics: the SQLite store reads the doomed identities inside the
eviction transaction, and only while a receipt is active, following the
existing pattern that receipt-active windows may pay bounded extra reads
while the steady path pays nothing. The in-memory graph records the node
set it already collects. Repo-scoped evictions keep failing closed: their
unbounded identity set is exactly the Go frontier the predicate-based
eviction design refuses to materialize.

The second commit closes the one case where the exact path would have
under-resolved relative to the fallback it replaces: a pending reference
in a file outside every receipt frontier that names an evicted
definition. The evicted file no longer declares the name, so the
file-scoped incoming pass never enumerates its unresolved stub; only the
whole-graph fallback ever retried those edges (the identity-changing
upsert receipt has carried the same latent shape, and TargetNames was
recorded but consumed by nothing). ResolveIncomingForNames rebinds the
pending references parked under the recorded names' stubs, bare and
repo-prefixed forms both, with resolveEdge's gates unchanged so a
still-ambiguous name binds no differently. All three receipt consumers
feed it the receipt's TargetNames after their file-frontier pass.

One deliberate asymmetry: on the in-memory Graph a single-file edit with
resolved incoming references still fails its receipt closed (its
ReindexEdges path does not describe itself yet), so it keeps the
conservative fallback; the SQLite backend composes the whole
restub-evict-readd cycle receipt-exact, and an end-to-end test pins
exactly that shape.

Two existing tests pinned the fail-closed behavior for file evictions and
were flipped deliberately to pin the exact frontier instead. New tests,
all watched fail first: exact frontiers for both backends (failing with
the production incomplete_reason, file_batch_evict.go), the
non-referenceable-only eviction staying neutral, the rollback path
leaving receipts untouched, the out-of-frontier pending reference
rebinding through TargetNames (failing as a stuck unresolved edge with
the wiring removed), both stub forms at the resolver level, and two
end-to-end indexer pins - the structural incremental reindex completing
its receipt, and the SQLite single-file-edit composition above (verified
failing against the unfixed store).

Full test suite on Windows matches my recorded baseline, no new failures.


zzet (changes requested):
Thanks for the detailed performance investigation and for taking on a difficult resolver path. The optimization direction makes sense, but I found reproducible correctness regressions at the current head (`c606f593`), so I don't think this is safe to merge yet.

### 1. [P1] Exact eviction receipts regress existing indexer behavior

The change at `internal/graph/file_batch_evict.go:34` certifies file eviction receipts as complete, but restubbed incoming callers are still represented as ordinary unresolved-file writes. The exact consumer then forward-resolves unchanged callers before provenance restoration.

I reproduced six existing tests failing at the PR head and passing at the exact base (`06939fce`), including under the race detector:

```bash
go test -race -count=1 -timeout=15m ./internal/indexer \
  -run '^(TestAffectedBy_MinifiedSkip_PreservesFactsNoFanout|TestPatchGraph_ExternalAdd_InPlaceWrite_KeepsCrossFileCaller|TestPatchGraph_ExternalRevert_RenameOver_KeepsCrossFileCaller|TestPatchGraph_ExternalRevert_UnlinkRecreate_KeepsCrossFileCaller|TestIncrementalReindex_DoubleCycle_Sqlite_IntraFileCaller|TestMultiWatcherThreeChunksRunEveryCatchupTailOnce)$'
```

The failures show three observable regressions:

- a transient minified-file parse skip deletes the persisted `b.js` reference facts;
- surviving cross-file edges are demoted from `lsp_resolved` to `ast_resolved`, which can make `find_usages` return zero;
- the exact watcher frontier includes an unchanged `caller.go` (67 files instead of 66).

All six already fail at the first commit in this PR, so this is caused by the eviction-receipt change rather than the later name pass.

**Proposed correction:** the safest immediate fix is to keep these eviction receipts incomplete. For the optimized path, distinguish restubbed incoming edges from genuinely new unresolved writes, keep unchanged callers out of `UnresolvedFiles`, and preserve their LSP provenance/reference facts through the reverse/name frontier. Please retain the six existing tests unchanged as an acceptance gate.

### 2. [P1] The receipt records nodes but omits touching edges

Both recorders capture doomed node identities, while `EvictFile(s)` also deletes every touching edge and still reports the receipt as complete. This is visible around `internal/graph/mutation_receipt.go:304` and `internal/graph/store_sqlite/file_batch_evict.go:84-146`.

Reproduced on both Graph and SQLite:

```go
// Keep is in unchanged keep.go; A is in a.go.
// Existing edge: Keep --calls--> A

token := store.BeginMutationReceipt()
store.EvictFile("a.go")
store.AddNode(newA)
receipt := store.EndMutationReceipt(token)
```

Observed:

```text
Complete=true
ResolutionFiles=[a.go]
ChangedFiles=[a.go]
TargetNames=[A pkg.A]
edges_after_exact_resolve=0
edges_after_ResolveAll=0
```

The surviving caller edge is deleted rather than parked under an unresolved stub, so neither resolution path can reconstruct it. A non-referenceable `KindPackage` with an external import edge is similarly classified `ResolutionRelevant=false` while the import edge is deleted.

**Proposed correction:** before deletion, detect incoming edges whose source survives. Either mark the receipt incomplete, or transactionally restub those edges and record their source files. Please add Graph and SQLite tests for resolved incoming edges, a pre-restubbed control, external imports to non-referenceable nodes, and overlapping receipt windows.

### 3. [P1] `ResolveIncomingForNames` omits wildcard member stubs

`internal/resolver/resolver.go:2813` emits only:

```text
unresolved::Target
repo::unresolved::Target
```

However, `graph.UnresolvedNameCandidateIDs` defines four name-owned forms, also including:

```text
unresolved::*.Target
repo::unresolved::*.Target
```

An exact-head test with both wildcard forms leaves both edges unresolved. Therefore method/field references outside every receipt file frontier can remain pending after the full fallback is removed.

**Proposed correction:** centralize the candidate-ID expansion and use it from both paths. Add tests for bare and repo-prefixed wildcard forms, plus the ambiguous case that must remain unresolved.

### 4. [P2] The new name pass rebuilds graph-wide indexes before proving there is work

`ResolveIncomingForNames` performs a second `buildPassIndexes()` before checking whether any requested name has a pending incoming edge. This happens immediately after the file pass has already built and cleared resolver indexes.

On a 50k-node no-hit fixture at this head:

```text
ResolveIncomingForNames:    40.25 ms/op, 43.8 MB/op, 251,349 allocs/op
ResolveFilesAndIncoming:     3.75 µs/op, 144 B/op, 3 allocs/op
```

**Proposed correction:** probe the named pending buckets first and return when empty, then use pending-scoped index preparation. Ideally, combine the file and name frontiers under one lock/index lifecycle and build the scoped indexes once. A no-pending benchmark/regression test would protect the intended performance improvement.

The SQLite transaction/rollback handling otherwise looks sound. I found no security or introduced dead-code concern, and the focused new tests pass; they just don't cover the provenance, persisted-fact, wildcard, and no-work cases above.

I'm happy to help validate the next revision or prepare focused regression tests if useful.

pbednarcik:
All four findings are addressed, one commit per finding on top of the
reviewed head. Point by point:

1. Restubbed callers are now distinguished from genuinely new unresolved
writes at the recorder level: a stub write that carries the restub
provenance stash stays out of UnresolvedFiles, while the receipt stays
resolution-relevant and the stub's name still joins TargetNames, so the
name frontier that restores the stashed tier is the one that rebinds it.
The in-memory reindexEdge also stops voiding receipts outright and mirrors
the SQLite reindex recorder (only newly unresolved writes create work).
Your six tests pass unchanged under the race detector and stay in place as
the acceptance gate. On the "pre-existing failures" note in my original
verification: that claim was wrong, these were this PR's regressions and my
base-tree check was contaminated. I regenerated my Windows baseline from a
clean worktree at the base commit so that class of mistake cannot repeat.

2. Both backends now probe for incoming edges from surviving sources before
recording an eviction: the in-memory gate scans the doomed nodes' in-edge
buckets under the write lock, the SQLite gate runs one EXISTS inside the
eviction transaction, both only while a receipt is active, and any hit
fails the receipt closed on every active window. That covers your repro,
the import-edge-to-a-package case, and overlapping receipt windows, with
Graph and SQLite tests for each plus a pre-restubbed control. I chose
fail-closed over transactionally restubbing in the store: which edge kinds
restub and how provenance is stashed is indexer policy, and duplicating it
below the layer that owns it seemed worse than keeping the bare path
conservative. The reindex composition keeps exactness because
restubIncomingRefs parks those edges first; the flipped pins now seed that
composition explicitly.

3. Both stub-key builders now go through the centralized expansion:
UnresolvedNameCandidateIDs delegates to a name-based core and the names
pass and the file pass's incoming leg both consume it, so the four forms
are one contract with no hand-built subsets. Tests cover the bare and
repo-prefixed wildcard rebinds plus the ambiguous member case that must
stay parked (two same-name method candidates; resolveEdge's gates hold).

4. The names pass now expands its stub keys first and probes their pending
buckets with one bounded read before touching the pass indexes; a no-work
call returns early. On the new benchmark fixture the no-pending call drops
from 53 microseconds and 24 KB to 0.7 microseconds and 0.8 KB per op, and
the cost no longer scales with the store. The benchmark stays as the
regression guard. Combining the file and name frontiers under one lock and
index lifecycle is a good further step; I left it out of this revision to
keep the diff reviewable and would take it as a follow-up.

Full test suite on Windows matches my baseline regenerated from a clean
worktree at the base commit, no new failures.

zzet (changes requested):
Thanks for addressing the previous review. I re-reviewed exact head `025792d7cea63fd2500e292e47effe2d8204c8f9`.

The earlier provenance, wildcard-stub, surviving-edge, and no-work performance findings appear fixed. However, I still need to request changes because two correctness/availability blockers and one material hit-path regression remain.

## [P1] Package eviction can falsely certify an unresolved import as unaffected

Locations:

- [`internal/graph/mutation_receipt.go:304-320`](https://github.com/zzet/gortex/blob/025792d7cea63fd2500e292e47effe2d8204c8f9/internal/graph/mutation_receipt.go#L304-L320)
- [`internal/graph/store_sqlite/file_batch_evict.go:164-188`](https://github.com/zzet/gortex/blob/025792d7cea63fd2500e292e47effe2d8204c8f9/internal/graph/store_sqlite/file_batch_evict.go#L164-L188)

Both receipt implementations skip evicted nodes when `!IsReferenceableSymbol(n.Kind)`. Import resolution, however, does consider `KindPackage` nodes as candidates.

### Reproduction

1. Add two `KindPackage` nodes, `one::pkg` and `two::pkg`, both with `QualName: "example/pkg"`.
2. Add an `EdgeImports` edge targeting `unresolved::import::example/pkg`.
3. Run `ResolveAll`; the edge correctly remains unresolved because two candidates exist.
4. Begin a mutation receipt, evict the file containing `one::pkg`, and end the receipt.
5. Consume the receipt using the production exact-receipt path.

Actual result on both Graph and SQLite:

```text
receipt={
  Complete:           true
  ResolutionRelevant: false
  TargetNames:        []
  ResolutionFiles:    []
}

target_after_exact="unresolved::import::example/pkg"
target_after_fallback="two::pkg"
```

The exact path skips the necessary retry, while `ResolveAll` immediately binds the remaining package. The surviving-incoming-edge check cannot detect this because the edge still targets the unresolved stub rather than either package candidate.

### Proposed fix

Centralize the mapping from resolver candidate nodes to their pending stub keys. For an evicted package candidate, record `import::<QualName>` in `TargetNames` and mark the receipt resolution-relevant.

Until every candidate kind has an exact mapping, conservatively mark the receipt incomplete when evicting a non-referenceable kind that participates in resolver candidate lookup.

Please add Graph and SQLite end-to-end tests for:

- two ambiguous packages → evict one → exact consumption binds the survivor;
- three ambiguous packages → evict one → remains unresolved;
- bare and repository-prefixed import stubs.

Suggested command:

```bash
go test -race -count=20 ./internal/graph ./internal/graph/store_sqlite \
  -run 'TestMutationReceiptEvictAmbiguousImportPackageCandidate'
```

## [P1] `runMasterResolveNames` races repository removal

Location:

[`internal/indexer/multi.go:958-959`](https://github.com/zzet/gortex/blob/025792d7cea63fd2500e292e47effe2d8204c8f9/internal/indexer/multi.go#L958-L959)

The new helper reads and iterates `mi.indexers` without holding `mi.mu`:

```go
prefixes := make([]string, 0, len(mi.indexers))
for prefix := range mi.indexers {
```

### Reproduction

An exact-head test using public methods runs the deferred receipt consumer concurrently with repository removal:

```go
go mi.RunDeferredPassesAllResult(context.Background())

for _, prefix := range trackedPrefixes {
    mi.UntrackRepo(prefix)
}
```

The deferred receipt drives this path:

```text
RunDeferredPassesAllResult
  -> FinishTailResult
  -> resolveDeferredMutations
  -> runMasterResolveNames
```

Under `go test -race`, the failure appears in approximately 0.05 seconds:

```text
read:  MultiIndexer.runMasterResolveNames  multi.go:958/959
write: MultiIndexer.UntrackRepo            multi.go:3243
```

This is not only a detector warning: concurrent map iteration and mutation can crash the daemon.

### Proposed fix

Snapshot the prefixes under the registry read lock, then release the lock before resolver work:

```go
mi.mu.RLock()
prefixes := make([]string, 0, len(mi.indexers))
for prefix := range mi.indexers {
    if prefix != "" {
        prefixes = append(prefixes, prefix)
    }
}
mi.mu.RUnlock()

sort.Strings(prefixes)
```

Please add the public deferred-tail-versus-`UntrackRepo` race test. `TrackRepoCtx` and repository replacement/reconciliation are sibling writers worth covering as well.

## [P2] Successful name resolution materializes the same pending frontier twice

Locations:

- [Initial probe at `resolver.go:2833`](https://github.com/zzet/gortex/blob/025792d7cea63fd2500e292e47effe2d8204c8f9/internal/resolver/resolver.go#L2833)
- [Second read at `resolver.go:2890`](https://github.com/zzet/gortex/blob/025792d7cea63fd2500e292e47effe2d8204c8f9/internal/resolver/resolver.go#L2890)

The no-work fast path is fixed, but the successful path calls `GetInEdgesByNodeIDs(stubKeys)` twice:

1. The probe materializes all matching edges to determine whether work exists.
2. `resolveIncomingStubKeysLocked` performs the identical read again.

A counting store records two identical four-key batches:

```text
unresolved::Target
unresolved::*.Target
repo::unresolved::Target
repo::unresolved::*.Target
```

With 42,700 pending edges:

| Store | One read | Current two reads |
|---|---:|---:|
| Graph | 18.8–26.8 µs, 344 KB | 34.5–57.5 µs, 689 KB |
| SQLite | 87.5–99 ms, 33.1 MB | 160–208 ms, 66.2 MB |

SQLite allocations double from approximately 640,247 to 1,280,494.

### Proposed fix

Retain the probe’s `inByStub` map and pass it into the resolution helper. Alternatively, add an existence-only store operation for the probe and retain one full materializing read after index construction.

Please add a hit-path benchmark; the current benchmark protects only the no-pending path.

## Validation

The following prior findings are fixed:

- all six existing indexer regression tests pass unchanged under `-race`;
- restub provenance and persisted reference facts are preserved;
- surviving resolved incoming edges fail receipts closed;
- all four unresolved stub forms are enumerated;
- the no-pending path avoids graph-wide index construction.

All hosted checks are green. Focused Graph, SQLite, resolver, and indexer tests pass, including affected race tests and `go vet`.

I found no hardcoded secret, authentication issue, or introduced dead code. The unlocked registry access is an availability concern because it can crash the daemon.

pbednarcik:
All three findings are addressed at exact head, each RED-first with your
reproduction as the fixture.

1. Package eviction receipts. ReceiptNamesForEvictedSymbol is now the
single authority on both backends mapping an evicted node to the stub
names its pending references may be parked under: referenceable
definitions keep name plus qualified name; a package - an import
candidate via the resolver's qualified-name tier - records the
import-prefixed stub of its qualified name (and of its bare name, for
stem-form specifiers) and joins the definition frontier like any other
candidate; any other kind carrying a qualified name participates in
candidate lookup without an exact mapping and fails the receipt closed
(extractors set QualName almost exclusively on referenceable symbols
and packages, so that branch is rare in practice). End-to-end tests
drive the production exact consumption over real evictions on both
Graph and SQLite: two ambiguous packages bind the survivor, three keep
the remaining ambiguity refused, and the repo-prefixed COPY-rewrite
stub form resolves through the prefixed key enumeration. Your suggested
command passes at -race -count=20 on both packages.

2. The registry race. The prefixes are snapshotted under mi.mu exactly
as you proposed, with the lock released before resolver work. The
public regression (FinishTailResult against UntrackRepo, with
TrackRepoCtx interleaved as the sibling writer) then exposed a second
unlocked reader on the same tail: collectCrossWorkspaceRules ranged
over the live GlobalConfig.Repos slice that UntrackRepo edits in place.
GlobalConfig gains ReposSnapshot (a copy under the same mutex that
serializes AddRepo / RemoveRepo / Save) and the rule collector uses it.
The test is clean at -race -count=10 with both fixes; either one alone
still trips the detector.

3. The double materialization. The probe's read is retained and handed
to the resolution helper - resolveIncomingStubEdgesLocked now takes the
caller-supplied batch, and the file-scoped path keeps its single fetch
through a thin wrapper. A counting-store regression pins the stub-key
frontier to exactly one store read on the hit path, and a
pending-frontier benchmark joins the no-pending guard: it uses a
two-candidate ambiguity so every iteration repeats the same
read-and-refuse work instead of draining the frontier on the first
pass.

Full race suite on Windows: the fail-name set matches my baseline except
one known timing flake in each direction (a Swift tight-window test that
reproduces standalone on a clean checkout of the base, and one baseline
flake that happened to pass this run). The receipt, resolver, and
indexer packages are clean under -race.

zzet (changes requested):
The design is sound and the mechanics check out. Build, `go vet` and `golangci-lint` are clean; `internal/graph`, `internal/graph/store_sqlite`, `internal/resolver`, `internal/config` and `internal/indexer` all pass (indexer also under `-race`). I neutered three of the fixes in turn — the eviction gate, the `runMasterResolveNames` call, the `HasRestubProvenance` skip — and each time the corresponding new test failed, so the tests bind to their fixes rather than decorating them. I also confirmed the SQLite probe's `arg, arg` binding is correct (both file predicates carry exactly one `?`), that `recordReindexedEdgeForReceipts` really does mirror `sqliteReindexReceipt.recordInserted`, and that the names pass cannot stamp terminal (stamping stays gated to an unscoped `ResolveAll`).

What I can't sign off on is that the change delivers the win it is built for. Two things, plus two smaller ones.

## 1. The new gate blocks the exact path on realistic trees

`recordFileEvictionForReceipts` (and the SQLite `EXISTS` probe) fails the receipt closed whenever a surviving file holds a resolved edge into the doomed node set. But `restubIncomingRefs` only parks `IsResolvableRefEdge` kinds, so `accesses_field`, `arg_of`, `tests`, `imports` and `contains` all trip the gate.

Measured on 130 flat-copied `internal/graph/*.go` files in a SQLite store, editing one file at a time through `incrementalReindexPathsWithReceiptMode`:

| | complete | `incomplete_reason` |
|---|---|---|
| base `3eb51c6d` | 0 / 25 | `file_batch_evict.go:110` |
| this branch | 1 / 25 (24/25 incomplete with `_test.go` excluded) | `file_batch_evict.go:159` |

Instrumenting the probe to dump the blocking rows, the offenders are `accesses_field` (54, 49, 45 … per evicted file), `arg_of`, and `tests`.

Every fixture in the PR has a `calls` edge as its only surviving incoming edge — the one kind restubbing does park — which is why they are green. On a real Go package the shape almost never holds. The production symptom the description cites, `incomplete_reason` naming `file_batch_evict.go`, is unchanged; only the line number moves.

## 2. The gate's own rationale argues against failing closed

The gate's comment says the deleted edge is reconstructible by neither the exact frontier nor the whole-graph fallback. I verified that half directly: an `accesses_field` edge from a surviving `a.go` into `b.go::T.F` is destroyed by reindexing `b.go` and is **not** restored by a subsequent `ResolveAll`.

So the gate detects unrecoverable damage and responds by forcing a whole-graph resolve that provably cannot repair it. The fallback is not more correct here — it is the same graph, minus the perf. If the eviction's *resolution* delta is still exactly describable (and I think it is: the lost edge is lost either way), the receipt should stay exact and the edge loss should be tracked as its own problem.

That edge loss looks like a real pre-existing bug worth splitting out: capability and structural edges from surviving files are silently dropped when a neighbour file is reindexed, and nothing re-creates them until the *source* file is reparsed.

## 3. `runMasterResolveNames` scales with the wrong quantity

`ResolveIncomingForNames` expands every name into four stub forms per repo prefix, and it is handed the receipt's whole `TargetNames` — which `recordAddedNodeForReceipts` fills with every added node's `Name` and `QualName`. A 130-file reindex produces 3,931 target names (≈30 per file).

Measured probe cost on SQLite (60k nodes / 60k pending edges), all of it paid *before* the "no pending" early-out:

```
names=1      prefixes=1    stubKeys=4          0.97ms
names=1000   prefixes=20   stubKeys=42000      469ms
names=10000  prefixes=20   stubKeys=420000     1.89s
names=40000  prefixes=20   stubKeys=1680000    8.08s
```

`BenchmarkResolveIncomingForNamesNoPending` uses one name, so it doesn't cover this. Only *evicted* names need the names pass — an added definition's pending references are already reached by the file frontier's incoming half, since its file is in `DefinitionFiles` and declares the name. A dedicated `EvictedNames` receipt field would bound the cost to the evictions that motivated it.

## 4. The `ReposSnapshot` race fix is incomplete

`workspace_resolve.go:574` now goes through `ReposSnapshot()`, but `:341` (`backfillWorkspaceSlugsWithImpact`, reached from `RunPreEnrichResolve` and `RunPreEnrichResolveFiles`) still iterates `Global().Repos` unlocked. Driving it concurrently with `UntrackRepo`, in the same shape as the new `TestRunDeferredPassesRacesUntrackRepo`, reports a real race on this branch:

```
WARNING: DATA RACE
  internal/config/global.go:506            (RemoveRepo)
  internal/indexer/multi.go:3306
  internal/indexer/workspace_resolve.go:341
  internal/indexer/workspace_resolve.go:317
```

Same one-line fix.

## Smaller notes

- `ReceiptNamesForEvictedSymbol` maps a `KindFile` node with an empty `QualName` to `(nil, exact=true)`, i.e. certified resolution-neutral. Import stubs can bind to file nodes (relative imports, Lua requires, Godot `res://`). It is narrow — a *resolved* import from a surviving file trips the gate anyway, so only a pre-existing ambiguity is exposed — but it is worth a comment saying so rather than leaving it implicit in the `qualName != ""` fall-through.
- Missing blank line between `recordReindexedEdgeForReceipts` and the `recordFileEvictionForReceipts` doc comment in `internal/graph/mutation_receipt.go`.

None of this is a correctness objection to the receipt model — the eviction-describes-itself idea is right, and the `TargetNames` consumption gap you found and closed in the second commit is a genuinely good catch. It is that (1) and (2) together mean the branch currently pays for the machinery without collecting the win on the shape it was written for, and I'd like to see the gate's cost/benefit resolved before this lands.

pbednarcik:
Thanks for measuring this rather than reading it. Findings 1 and 2 are right, and I have taken the direction you argued for: the gate is gone. Finding 3 needed a prerequisite fix before it was safe, which is the one place I did more than you asked. I also put the result through an adversarial pass of my own, which found two defects in my own revision; those are fixed and described at the end.

## 1 and 2: the gate is removed, not narrowed

I reproduced your measurement before deciding. Flat copy of `internal/graph` into a SQLite store, editing the first 25 files in sorted order one at a time through `incrementalReindexPathsWithReceiptMode`, `_test.go` excluded, 69 non-test files rather than your 130:

| | complete | `incomplete_reason` |
|---|---|---|
| `ad53dd4e` | 1 / 25 | `file_batch_evict.go:159` x24 |
| this revision | 25 / 25 | - |

Same ratio and the same line you saw. In fairness to the method: that line also fires if the identity read errors, so it names the call site rather than proving the probe was the cause on its own - what proves it is that head reaches 25 of 25 with only the probe removed. Happy to hand over the harness.

Your second point is what decided it. I verified the premise directly rather than reasoning about it. `TestResolveAllDoesNotRestoreAnEvictionDestroyedCapabilityEdge` in `internal/resolver` builds a resolved `accesses_field` edge from a surviving `a.go` into `b.go::T.F`, reindexes `b.go`, and asserts the edge is absent both immediately and after a full `ResolveAll`. It is. Resolution retargets edges that still exist and are parked under a stub, and a deleted edge is neither.

One correction to how I first wrote this up. The two paths are not equivalent in general - `ResolveAll` also runs graph-wide passes the file frontier never runs, `resolveRelativeImports` and the Lua / Godot / Razor ones, and I have a fixture where the exact path leaves `unresolved::pyrel::pkg/b` while the fallback binds `pkg/b.py`. They agree only about the destroyed edge. That is the whole question here, but the narrower claim is the true one and it is what the test pins.

I should also correct your cost framing slightly in my own favour's opposite direction. On the watcher path an incomplete receipt does not force a whole-graph resolve: `incrementalResolutionFrontier` hands back `DerivedInvalidation.Files` and `resolveIncrementalRepoMutationMode` runs a scoped pass over it, skipping the name pass. The whole-graph resolve is reserved for the deferred tail and for the corner where no frontier survives. So the gate's cost on the path where it actually fired was a wider scoped pass, not a full resolve - real, but smaller than my first draft of this reply claimed.

Two more things about the gate. `contains` never tripped it, contrary to what I first wrote: a file node carries `FilePath` equal to its own path, so `EvictFile` takes it too, and `contains` is only emitted file-node to import-node within one file, which makes both endpoints doomed. And the exact-frontier fixtures that had a surviving incoming edge carried a `calls` edge, one of the ten kinds restubbing parks, which is why none of them could reach this.

The two tests that asserted fail-closed are inverted rather than deleted, and they now assert the destruction explicitly - the surviving caller's edge list must be empty afterwards. I would rather the loss be pinned by a test that says so than implied by a gate that mentions it in a comment. It is filed as its own issue.

## 3: evicted names, plus the bug that made the naive version unsafe

`EvictedNames` exists and the name pass consumes it. It is filled by node eviction on both backends and by the old half of a SQLite identity change, since those names vanished too.

Before wiring it up I checked your safety argument - that an added definition's pending references are already reached by the file frontier's incoming half - and it was not true yet. `collectIncrementalFileFrontierMode` hand-built two of the four name-owned stub forms, so both wildcard member forms were missing. Narrowing the name pass on top of that would have stranded every member reference parked under `unresolved::*.Name` whose definition was added rather than evicted, until the next whole-graph resolve. A recall regression, and a quiet one.

This is my own miss. An earlier commit on the branch routed `ResolveIncomingForNames` and `resolveIncomingLocked` through `UnresolvedNameCandidateIDs` and its message claimed the contract had no hand-built subsets left. There was a third builder, and it owns the incoming leg of every batched incremental resolve - the path this PR's performance story runs through. `UnresolvedNameCandidateIDsForName`'s own doc says a path that hand-builds a subset silently strands the missing forms, which is what happened.

So the frontier fix lands first, with a rebind test and an ambiguity control mirroring the pair you already have on the names pass, and the narrowing sits on top of it.

That widening is not free, and its commit message does not say so. The frontier's `stubKeys` doubles. On a 10,000-node frontier with nothing pending, in-memory backend, it costs +66% time and +142% allocations, 5.77ms to 9.59ms and 28,274 to 68,535 allocs. Paid on every incremental resolve including single-file saves, where it is negligible in absolute terms, and clearly net-positive against what the narrowing saves on the batch shape - but you would have found it, so here it is.

On the narrowing's own cost, the new benchmark covers what the single-name one could not. No pending edges, in-memory graph:

```
BenchmarkResolveIncomingForNamesNoPending-32                 710.0 ns/op        816 B/op         13 allocs/op
BenchmarkResolveIncomingForNamesNoPendingManyNames-32     16433340 ns/op   27824752 B/op     204800 allocs/op
```

That second row is 2000 names over 20 prefixes, the shape `TargetNames` produced, and it is the in-memory analogue of your 8.08s SQLite number.

One loose end for you to rule on. `TargetNames` now has no consumer in any resolution path; it survives as the receipt's description of every name the mutation touched, alongside `TargetIDs`, which is also descriptive. It is not free - two map inserts per added node per active receipt, plus a sort at `End`, on exactly the batch-sized path this PR exists to shrink - and its meaning is backend-divergent, since the graph backend records non-referenceable added nodes' names and SQLite does not. I left it because removing it changes the receipt's public shape and a lot of assertions. Say the word and it goes.

## 4: the second unlocked reader

Fixed the same way, and the new race test drives the public `BackfillWorkspaceSlugsWithImpact` against both registry writers. Before the fix the detector reports the read at `workspace_resolve.go:341` against `AddRepo` at `global.go:488` - the `TrackRepoCtx` side rather than the `RemoveRepo` side you traced, which is the sibling writer you said was worth covering. The reader loop runs until the writes finish rather than for a fixed count; a fixed count raced past them and reported nothing.

## Smaller notes

Both done. The `ReceiptNamesForEvictedSymbol` fall-through now carries a comment, though not the one you suggested - your justification for it being narrow was that a resolved import from a surviving file trips the gate anyway, and with the gate gone that no longer holds. It says instead that failing closed there is not available at all, because every file reindex evicts its own file node, so the branch would retire the exact path entirely.

The blank line went in, then the gate change removed the function it preceded, so there is nothing left to look at.

## What my own review pass found

I ran three adversarial passes over this revision before sending it. Neither the gate removal nor the narrowing was shown to drop resolution work, and the four mutants bind: restore the gate, stop populating `evictedNames`, revert the frontier to the bare pair, or point the consumers back at `TargetNames`, and a specific test fails each time. But two real defects turned up, both mine, both fixed here.

The first is a consequence of removing the gate. An evicted node that maps to no stub name was taken as proof of resolution neutrality, so an eviction touching only such nodes produced `ResolutionRelevant=false` - the one verdict `incrementalResolutionFrontier` reads as "no catch-up work", so no pass ran at all. That is wrong for file nodes: an import parks under the specifier as written, not the file path, so there is no stub key to offer, but relative imports, Lua requires and Godot `res://` paths all bind to file nodes, and removing one can collapse an ambiguity for a pending import elsewhere. The gate had been masking this, because it fired whenever a surviving file held an edge into the doomed node. `EvictedNodeNeedsResolutionFrontier` names the single exception; every other nameless kind stays neutral and the existing neutrality tests pin that.

The second: the evicted-name skip on the SQLite identity-change path was keyed on string equality against either final name, and both arms were unsound. The file frontier enumerates a node's `Name` only, and only for referenceable kinds. So matching `final.QualName` is never grounds to skip, and matching `final.Name` is not either once the final kind has stopped being referenceable. A referenceable definition becoming a param or an import would have had its old name enumerated by neither pass.

One thing worth naming that is not a defect. Removing the gate widens how often the exact path is taken; narrowing to `EvictedNames` reduces what that path sweeps. Each is argued on its own, but together they cut the incidental self-healing of pre-existing stranded pendings that the fallback was doing on nearly every reindex. Pendings this mutation creates or unblocks stay covered - added definitions through `DefinitionFiles`, removed ones through `EvictedNames` - so it is not a regression against the receipt's contract. It is the one place observable recall on a live daemon could drift, and it seems worth saying out loud rather than leaving for someone to find.

## Verification

`go build ./...` clean. `go vet ./internal/...` reports only the pre-existing `possible misuse of unsafe.Pointer` at `internal/parser/languages/grammar_load_windows.go:33`, in a file this branch does not touch.

`internal/graph` and `internal/graph/store_sqlite` pass. `internal/resolver` carries the same 14 failures before and after this revision, pre-existing JS/TS import-resolution and workspace cases. `internal/indexer` under `-race` is unchanged against the branch head baseline: the same 32 pre-existing failures at `ad53dd4e` and after, no new ones and no data races. These are Windows runs; the Linux side is CI's.

zzet (approved):
All four findings are closed, and I could reproduce every number you quoted. Build, `go vet`, `golangci-lint` clean; `internal/graph`, `internal/graph/store_sqlite`, `internal/resolver`, `internal/config`, `internal/indexer` all pass, and all five pass under `-race` (macOS; `store_sqlite -race` takes 610s, indexer 253s). Approving.

## The win now lands

Same harness as before — flat copy of `internal/graph` into a SQLite store, 25 single-file edits in sorted order through `incrementalReindexPathsWithReceiptMode`, `_test.go` excluded:

| | complete | `incomplete_reason` |
|---|---|---|
| `ad53dd4e` | 1 / 25 | `file_batch_evict.go:159` ×24 |
| `5ca780d0` | **25 / 25** | — |

Your ratio and your line number, exactly. And with the probe gone, `file_batch_evict.go:159` can now only mean a genuine identity-read error, which makes the diagnostic sharper than it was.

The narrowing shows up in the receipts too: on a single-file edit `EvictedNames` runs 3–72 entries against `TargetNames`' 29–75, and your `BenchmarkResolveIncomingForNamesNoPendingManyNames` reproduces at 27.3 ms/op vs 1.1 µs for the single-name case here — a 25,000× spread on the shape the consumers used to be handed.

## What I checked beyond the four findings

**The recall drift you flagged does not materialise.** This was the one thing in your reply I did not want to take on argument. I first ran the drift measurement with the corpus read from the checked-out tree and got a clean +26 pending edges on the branch — which turned out to be my own harness fault, since the branch adds nodes to the very `internal/graph` files it copies. Re-run against a corpus frozen at `3eb51c6d`, 130 files, 25 incremental edits through the full `incrementalWatcherPaths`, base / `ad53dd4e` / head are byte-identical at every stage:

```
AFTER_INDEX       nodes=5395 edges=24029 pending=2814
AFTER_RESOLVEALL  nodes=5395 edges=24029 pending=2792
after 25 edits    pending=3107 -> full resolve 3075   (healed_by_full=32)
```

Same 32 on all three, deterministic across repeated runs. Whatever the exact path leaves for a later whole-graph resolve, the fallback left exactly the same.

**The relative-import divergence does not widen either.** Your `unresolved::pyrel::pkg/b` fixture is the shape where the exact path and `ResolveAll` genuinely disagree, and gate removal makes the exact path far more common — so I built it end-to-end: a `pkg/` with `from . import b`, edit `a.py`, reindex through `incrementalWatcherPaths`. Head, `ad53dd4e` and base all leave `unresolved::pyrel::pkg/b` parked after the incremental pass and all heal it on the next `ResolveAll`. Your own correction is why: `incrementalResolutionFrontier` hands back a non-empty file set either way, so the watcher took the scoped branch on base too. The divergence is real at the resolver level and irrelevant on the path this PR changes.

**The frontier fix is a recall improvement in its own right, not just a prerequisite.** Worth saying because your commit message frames it as clearing the way for the narrowing. Add `ZZTarget()` and `h.ZZMember()` to a caller, then create the definitions in a new file and reindex only that file:

```
base 3eb51c6d   AFTER_INCREMENTAL  b.go::ZZTarget + unresolved::*.ZZMember   <- still parked
head 5ca780d0   AFTER_INCREMENTAL  b.go::ZZTarget + b.go::Holder.ZZMember
```

On main a wildcard member reference is never rebound by an incremental reindex — it waits for the next whole-graph resolve. `ad53dd4e` also binds it, but only because the names pass was being fed `TargetNames` and masked the bug. That is a user-visible fix on the daemon's hot path and deserves to be in the commit message.

Cost of the widening, on the repo's own benchmarks: `BenchmarkResolveFileAndIncomingNoPending100K` 1993 → 3912 ns/op (+96%, a bigger multiplier than your +66% but 2 µs absolute), `BenchmarkResolveFileAndIncomingCSharpVisibility100K` 28.4 → 30.4 ms (+7%). Fine for what it buys.

**Mutants.** Restore the gate on `EvictFiles` → `TestMutationReceiptEvictImportToNonreferenceableStaysExact` fails. Stop populating `evictedNames` in `recordEvictedNodesForReceipts` → four tests fail. Make `EvictedNodeNeedsResolutionFrontier` return false → `TestMutationReceiptEvictNamelessNodeStaysResolutionRelevant` fails. Point `backfillWorkspaceSlugsWithImpact` back at the live slice → `TestBackfillWorkspaceSlugsRacesUntrackRepo` reports the race against both writers, `RemoveRepo` at `global.go:506` and `AddRepo` at `:488`, so your reader-until-writes-finish loop does cover the side I originally traced and the sibling.

**Coverage of the narrowing.** All three consumers (`resolveReceiptNamePendings`, both `resolveIncrementalRepoMutationMode` arms, `resolveDeferredMutations`) reach the file pass through `ResolveFilesAndIncoming`, so the fixed builder covers them uniformly — the narrowing's safety argument does not have a hole where one consumer runs a different frontier. And `#690` exists, so "tracked separately" is real rather than a promise.

## The `TargetNames` ruling you asked for

Keep it, and document it. Two reasons:

The cost argument does not survive measurement. A 130-file batch produces ~3.9k names, so maintaining the set is a few thousand map inserts and one sort against a multi-second reindex. That is noise. What made the old wiring expensive was handing those names to the probe (4 stub forms × 20 prefixes), and that is exactly what you already fixed.

The divergence argument is the real one, and removal is not the only fix for it. `TargetNames` and `TargetIDs` are now both descriptive-only — neither has a production consumer outside the storetest conformance assertion. The hazard is not their cost, it is that a future consumer wires one up and inherits behaviour that differs by backend (the graph accumulator records non-referenceable added nodes' names, SQLite does not). So put that on the field: say it is descriptive, say `EvictedNames` is what resolution consumes, and name the divergence. Removing it would also churn a lot of assertions to buy microseconds.

## Smaller notes

`internal/graph/stub.go`: `EvictedNodeNeedsResolutionFrontier` landed *inside* `ReceiptNamesForEvictedSymbol`'s doc comment. The block that starts "ReceiptNamesForEvictedSymbol maps one evicted node to…" at :359 now runs unbroken into the new function's own doc at :390 and attaches to `EvictedNodeNeedsResolutionFrontier`, which leaves `ReceiptNamesForEvictedSymbol` at :407 undocumented. Needs a blank line before :390 and the new function moved out of the block — the fall-through paragraph at :385-389 belongs with the function it describes.

`ResolutionRelevant=false` is now unreachable for any real eviction, since every file reindex evicts its own file node and `EvictedNodeNeedsResolutionFrontier` makes that relevant. That is the correct outcome, but the doc's "Every other nameless kind really is neutral… creates no resolution work" reads as if the neutral branch is still live for evictions, and both `…NonreferenceableOnlyStaysNeutral` tests pin it with a graph that has no file node — a shape production never produces. One sentence saying the branch survives as a per-node skip rather than a receipt-level verdict would stop the next reader from trusting it.

Out of scope, but `ReposSnapshot` now exists and three more readers still iterate the live slice: `internal/mcp/tools_multi.go:539`, `cmd/gortex/daemon_controller.go:558`, `cmd/gortex/daemon_state.go:320`. `config.FindRepoByPrefix` is the worse one — it iterates unlocked *and* returns a pointer into the slice. Same one-line fix, separate PR.

PR #679, opened 2026-08-27
resolver: derived tests edges never re-resolve, and test-linkage skips unresolved calls

pbednarcik:
While validating other work I hit a counted-fixture tripwire: an
extension-stamped edge appeared where the binder provably refuses to
bind. The artifact was odd enough to chase down: kind `tests`, origin
ast_inferred, confidence 0.75, Meta `resolution: extension_method`,
pointing from a plain method to the one extension method sharing its
name, while the calls edge at the same call site stayed honestly
unresolved.

Root cause, in three steps:

1. The test-linkage pass clones a test caller's calls edges into
EdgeTests, and it clones unresolved calls too. The clone carries none of
the original's Meta, so the receiver evidence (`receiver_shape` and
friends) is stripped.

2. `resolveEdge` has no route for EdgeTests, so an unresolved clone
falls into the member-call case like any other `*.` target.

3. On the real calls edge the extension shape guard refuses (the
receiver is a `List<int>`, the candidate is `this List<string>`). The
naked clone has no receiver evidence at all, so nothing can refuse, and
the untyped pool-unique fallback binds it at 0.75.

The result is a tests edge asserting the exact bind the guarded calls
edge exists to prevent. The exposure is not small: my stores carry ~29k
tests edges pointing at unresolved targets (assertion-framework calls
from test files dominate), each one an evidence-stripped clone a future
incremental re-resolve can convert the same way. Cold builds emit test
edges after resolution settles, which is why the shape only surfaces on
incremental paths.

Two commits, one per layer:

1. `resolveEdge` gains an EdgeTests gate, the same shape as the existing
type-position gate (extends/implements/returns/typed_as): a derived edge
never routes through call resolution. The tests layer follows its calls
edge; the resolver never binds it independently. This half also protects
the clones already sitting in existing stores.

2. The test-linkage emitters skip unresolved calls entirely. A tests
edge to an unresolved stub can never say what is tested, so the clones
were dead weight at scale even before the wrong-bind hazard; a call that
resolves later links on the next reconciliation of its caller.

Both fixes are RED-first: the resolver test reproduces the field
artifact end to end (a C# fixture through the real extractor, the tests
clone added at the same site, ResolveAll binds it pre-fix while the
calls-edge control stays refused), and the indexer test pins that only
resolved subjects clone.

Verification: the full race suite's fail-name set is identical to my
Windows baseline at the merge base (one known-flaky baseline name
happened to pass). The branch is based on 3eb51c6d and merges cleanly
onto current main.

### DIFF
--- a/internal/indexer/multi.go
+++ b/internal/indexer/multi.go
@@ -1687,7 +1687,9 @@ func (mi *MultiIndexer) runGlobalGraphPassesTopologyHeld(
 	// calls parked for those repos are already served - discard them, or
 	// the first later incremental save drains the entire test corpus into
 	// a scoped re-projection under ResolveMutex.
-	mi.discardRetargetedTestCallFiles(scanPrefixes)
+	if scanPrefixes == nil {
+		mi.discardRetargetedTestCallFiles(nil)
+    }
 	passStart("entrypoint_hierarchy")
 	ctrlStart := time.Now()
 	// Seeds from already-stamped entry points, so cost is O(seed
--- a/internal/indexer/multi.go
+++ b/internal/indexer/multi.go
@@ -924,6 +924,7 @@ func (mi *MultiIndexer) runMasterResolveHookedContext(ctx context.Context, scope
 		zap.Int("pending_scanned", stats.PendingBefore),
 		zap.Int("pending_admitted", stats.PendingAfter),
 		zap.Error(err))
+	mi.reconcileRetargetedTestCalls(master.TakeRetargetedTestCallFiles())
 	return err
 }
 
--- a/internal/indexer/multi.go
+++ b/internal/indexer/multi.go
@@ -936,6 +936,22 @@ func (mi *MultiIndexer) runMasterResolveFiles(files []string, useLSP bool) {
 		zap.Int("files", len(files)),
 		zap.Int("pending_scanned", stats.PendingBefore),
 		zap.Int("pending_admitted", stats.PendingAfter))
+	mi.reconcileRetargetedTestCalls(master.TakeRetargetedTestCallFiles())
+}
+
+// reconcileRetargetedTestCalls re-runs the scoped test projection over the
+// caller files of test-classified calls a resolution pass just bound. The
+// receipt-exact catch-up lanes run with no global test-edges pass behind
+// them, so a call that binds later than its caller's projection would
+// otherwise never gain its EdgeTests. No-op on an empty frontier.
+func (mi *MultiIndexer) reconcileRetargetedTestCalls(files []string) {
+	if len(files) == 0 {
+		return
+	}
+	_, emitted := markTestSymbolsAndEmitEdgesScoped(mi.graph, nil, files...)
+	mi.logger.Info("DEFERRED-TIMING test-edges reconcile for retargeted callers",
+		zap.Int("files", len(files)),
+		zap.Int("edges", emitted))
 }
 
 // RunPreEnrichResolve runs the resolution stage that makes references queryable
--- a/internal/indexer/multi.go
+++ b/internal/indexer/multi.go
@@ -1682,6 +1682,12 @@ func (mi *MultiIndexer) runGlobalGraphPassesTopologyHeld(
 		zap.Int("test_symbols", marked),
 		zap.Int("edges", emitted),
 		zap.Duration("elapsed", time.Since(testStart)))
+	// This projection covers every test caller in scanPrefixes (nil =
+	// whole graph), so the retarget frontiers the per-repo ResolveAll
+	// calls parked for those repos are already served - discard them, or
+	// the first later incremental save drains the entire test corpus into
+	// a scoped re-projection under ResolveMutex.
+	mi.discardRetargetedTestCallFiles(scanPrefixes)
 	passStart("entrypoint_hierarchy")
 	ctrlStart := time.Now()
 	// Seeds from already-stamped entry points, so cost is O(seed
--- a/internal/graph/store_sqlite/schema_version.go
+++ b/internal/graph/store_sqlite/schema_version.go
@@ -32,7 +32,7 @@ import (
 // index changes in a way an old on-disk DB would not already have, and append a
 // matching schemaMigrations entry describing how to bring an older store
 // forward (in place, or by rebuild).
-const currentSchemaVersion = 12
+const currentSchemaVersion = 13
 
 // schemaMigration is one forward step. Exactly one strategy applies:
 //   - rebuild=true: the change introduces structure/data that can only come
@@ -80,6 +80,23 @@ var schemaMigrations = []schemaMigration{
 	{version: 10, name: "rebuild vector corpus ownership and parents", inPlace: rebuildVectorCorpusSchema},
 	{version: 11, name: "add symbol FTS normalization state", inPlace: createSymbolFTSNormalizationStateTable},
 	{version: 12, name: "normalize dir column separators", inPlace: normalizeDirColumnSeparators},
+	{version: 13, name: "purge unresolved derived tests edges", inPlace: purgeUnresolvedTestsEdges},
+}
+
+// purgeUnresolvedTestsEdges removes derived EdgeTests rows whose target is
+// an unresolved stub, in both spellings (`unresolved::X` and the multi-repo
+// `<repo>::unresolved::X` COPY-rewrite form). The test-linkage pass cloned
+// them from unresolved calls before the emission guard existed; stripped of
+// the call's receiver evidence they are naked stubs the resolver now
+// refuses to bind, new emission never re-creates them, and warm startup may
+// skip file-scoped reconciliation entirely — so an old store keeps paying
+// their resolver-scan cost forever without this explicit purge. Idempotent
+// and bounded to the tests kind: pending calls and resolved projections are
+// untouched.
+func purgeUnresolvedTestsEdges(tx *sql.Tx) error {
+	_, err := tx.Exec(`DELETE FROM edges WHERE kind = 'tests'
+		AND (to_id LIKE 'unresolved::%' OR to_id LIKE '%::unresolved::%')`)
+	return err
 }
 
 // normalizeDirColumnSeparators rebuilds the two generated dir columns whose
--- a/internal/graph/store_sqlite/tests_edge_purge_test.go
+++ b/internal/graph/store_sqlite/tests_edge_purge_test.go
@@ -1,0 +1,62 @@
+package store_sqlite
+
+import (
+	"database/sql"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/zzet/gortex/internal/graph"
+)
+
+// TestOpenPurgesLegacyUnresolvedTestsEdges is the upgrade proof for the
+// unresolved-tests-edge purge: stores written before the emission guard
+// hold derived EdgeTests rows cloned from unresolved calls (naked stubs
+// nothing may ever bind). New emission never re-creates them, and warm
+// startup may skip file-scoped reconciliation entirely, so an explicit
+// versioned migration removes them. Every other row — resolved tests
+// projections and ordinary unresolved calls — must survive untouched.
+func TestOpenPurgesLegacyUnresolvedTestsEdges(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "store.sqlite")
+
+	s, err := Open(path)
+	require.NoError(t, err)
+	s.AddNode(&graph.Node{ID: "r/a_test.go::TestFoo", Kind: graph.KindFunction, Name: "TestFoo", FilePath: "r/a_test.go", RepoPrefix: "r"})
+	s.AddNode(&graph.Node{ID: "r/b.go::Foo", Kind: graph.KindFunction, Name: "Foo", FilePath: "r/b.go", RepoPrefix: "r"})
+	s.AddBatch(nil, []*graph.Edge{
+		// A healthy resolved projection: must survive.
+		{From: "r/a_test.go::TestFoo", To: "r/b.go::Foo", Kind: graph.EdgeTests, FilePath: "r/a_test.go", Line: 5},
+		// Legacy unresolved clones in both stub spellings: must be purged.
+		{From: "r/a_test.go::TestFoo", To: graph.UnresolvedMarker + "Gone", Kind: graph.EdgeTests, FilePath: "r/a_test.go", Line: 6},
+		{From: "r/a_test.go::TestFoo", To: "r::" + graph.UnresolvedMarker + "*.Gone", Kind: graph.EdgeTests, FilePath: "r/a_test.go", Line: 7},
+		// An ordinary pending call: NOT a tests edge, must survive.
+		{From: "r/a_test.go::TestFoo", To: graph.UnresolvedMarker + "Gone", Kind: graph.EdgeCalls, FilePath: "r/a_test.go", Line: 6},
+	})
+	require.NoError(t, s.Close())
+
+	// Simulate a store written before the purge shipped.
+	withRawDB(t, path, func(db *sql.DB) {
+		_, err := db.Exec(`PRAGMA user_version = 12`)
+		require.NoError(t, err, "reset to the pre-purge version")
+	})
+
+	s2, err := Open(path)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = s2.Close() })
+
+	var kinds []string
+	for _, e := range s2.GetOutEdges("r/a_test.go::TestFoo") {
+		if e == nil {
+			continue
+		}
+		if e.Kind == graph.EdgeTests {
+			require.False(t, graph.IsUnresolvedTarget(e.To),
+				"legacy unresolved tests edge survived the upgrade: %+v", e)
+		}
+		kinds = append(kinds, string(e.Kind)+"->"+e.To)
+	}
+	require.Contains(t, kinds, "tests->r/b.go::Foo", "the resolved projection must survive")
+	require.Contains(t, kinds, "calls->"+graph.UnresolvedMarker+"Gone", "the pending call must survive")
+	require.Len(t, kinds, 2, "exactly the two healthy edges remain")
+}
--- a/internal/indexer/incremental_derived.go
+++ b/internal/indexer/incremental_derived.go
@@ -95,6 +95,25 @@ func (mi *MultiIndexer) drainRetargetedTestCallFiles(prefixSet map[string]struct
 	return appendUniqueSorted(nil, files...)
 }
 
+// discardRetargetedTestCallFiles drops the per-repo retarget frontiers a
+// graph-wide (or repo-scoped) test projection has just superseded. A nil
+// prefix set discards every tracked repo's frontier. Destructive by
+// design: the callers noted there are already covered by the projection
+// that just ran, so draining them later would only re-project them.
+func (mi *MultiIndexer) discardRetargetedTestCallFiles(prefixes map[string]bool) {
+	mi.mu.RLock()
+	defer mi.mu.RUnlock()
+	for prefix, idx := range mi.indexers {
+		if idx == nil || idx.resolver == nil {
+			continue
+		}
+		if prefixes != nil && !prefixes[prefix] {
+			continue
+		}
+		idx.resolver.TakeRetargetedTestCallFiles()
+	}
+}
+
 // RunIncrementalDerivedPasses executes only the derived families invalidated
 // by the exact per-file plans. A legacy database without persisted fingerprints
 // takes the old scoped-global path once; ordinary body/metadata edits never do.
--- a/internal/indexer/incremental_derived.go
+++ b/internal/indexer/incremental_derived.go
@@ -72,6 +72,29 @@ func (idx *Indexer) runStandaloneIncrementalDerivedPasses(plan DerivedInvalidati
 	})
 }
 
+// drainRetargetedTestCallFiles collects, from each named repo's resolver,
+// the caller files of test-classified calls the resolution step just bound
+// (see resolver.Resolver.TakeRetargetedTestCallFiles). Destructive: the
+// per-resolver frontier is cleared.
+func (mi *MultiIndexer) drainRetargetedTestCallFiles(prefixSet map[string]struct{}) []string {
+	mi.mu.RLock()
+	resolvers := make([]*resolver.Resolver, 0, len(prefixSet))
+	for prefix := range prefixSet {
+		if idx := mi.indexers[prefix]; idx != nil && idx.resolver != nil {
+			resolvers = append(resolvers, idx.resolver)
+		}
+	}
+	mi.mu.RUnlock()
+	var files []string
+	for _, r := range resolvers {
+		files = append(files, r.TakeRetargetedTestCallFiles()...)
+	}
+	if len(files) == 0 {
+		return nil
+	}
+	return appendUniqueSorted(nil, files...)
+}
+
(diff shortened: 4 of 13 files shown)


zzet (changes requested):
Thanks for tackling this — the underlying receiver-evidence bug is real, and preventing unresolved calls from being projected is the right direction. I found several production paths that the current tests do not cover, though, so I think these need to be addressed before merge.

## 1. P1: enforce the `EdgeTests` guard before every resolver path

The new guard in `resolver.go` runs after inline LSP resolution. Deferred LSP also collects `EdgeTests` before reaching this switch, and `CrossRepoResolver` has an independent `resolveEdge` with no equivalent guard.

Exact-head reproductions:

```text
ResolveFile + fake LSP: EdgeTests bound to target.ts::Foo
ResolveAll + fake LSP:  EdgeTests bound to target.ts::Foo
CrossRepo ResolveAll:   EdgeTests bound to repoB/target.ts::Foo
```

This contradicts the PR's central invariant and means persisted unresolved test edges can still be independently misbound.

Suggested fix: use a shared resolver-eligibility predicate before inline/deferred LSP and in `CrossRepoResolver`. Please add `ResolveFile` + LSP, `ResolveAll` + LSP, and cross-repository regression tests.

Relevant locations:
- `internal/resolver/resolver.go:3093-3098`
- `internal/resolver/lsp_resolve.go:277`
- `internal/resolver/cross_repo.go:1044`

## 2. P1: reconcile calls that become resolved later

Skipping an unresolved call in `internal/indexer/test_edges.go:716` can permanently omit a valid test projection.

Reproduction:

1. Seed `TestFoo -> unresolved::Foo` as an `EdgeCalls` and run test projection; no `EdgeTests` is emitted.
2. Add `pkg/foo.go::Foo`.
3. Run `ResolveIncomingForFile("pkg/foo.go")`; the call now resolves to `Foo`.
4. Run a declaration-only derived plan for `pkg/foo.go`.
5. The expected `EdgeTests(TestFoo, Foo)` is still absent.

The exact-head regression fails with:

```text
resolved test call has no EdgeTests projection
```

A declaration-only plan does not reconcile test edges, and the caller file is not necessarily revisited later. Suggested fix: have resolution report the source files of retargeted `EdgeCalls`, merge that frontier into `DerivedInvalidatesTests`, and reconcile those callers after resolution. Please cover add, rename/delete/rebind, and cross-repository cases.

## 3. P2: clean up legacy unresolved `EdgeTests` on upgrade

The new filter prevents new unresolved clones but does not remove rows persisted by older versions. Whole-graph emission only adds current projections, while warm/no-change startup may skip file-scoped reconciliation entirely. An exact-head test seeded with a legacy unresolved `EdgeTests` row retains it after reconciliation.

For 29,000 legacy rows, the measured `ResolveAll` overhead was approximately:

| Store | Time | Allocated |
|---|---:|---:|
| In-memory graph | 102.7 ms | 69.3 MB |
| SQLite | 915.5 ms | 50.3 MB |

Suggested fix: add a versioned migration that deletes unresolved `EdgeTests`, or evict and rebuild the applicable projection set. Please include an SQLite seed/close/reopen upgrade test.

The two tests added by this PR pass, as do the hosted checks and `go vet`, but they cover only immediate emission and the no-LSP heuristic resolver path. I found no security, hardcoded-secret, off-by-one, or dead-code issue in the change.

pbednarcik:
All three findings are addressed at exact head, each RED-first with your
reproductions as fixtures.

1. The refusal now covers every resolution path through one shared
predicate (resolutionExempt): resolveEdge consults it first, ahead of
the inline LSP hot-path; lspDeferTarget refuses exempt edges at
collection time, since the bulk batch is gathered before resolveEdge
sees the clone; and the cross-repository resolveEdge refuses them ahead
of its tiers. Your three reproductions are the regression tests -
ResolveFile with an answering LSP helper, ResolveAll through the
deferred batch, and the cross-repository same-repo name tier - and the
old dispatch-internal case is deleted in favor of the shared predicate.

2. Later-resolved calls now reconcile. Resolution passes report their
retargeted test-call frontier: the core apply sites (the parallel
ResolveAll batch, the incremental and file-scoped applies, the deferred
LSP batch, and the cross-repository pass) note the caller file of every
calls edge they bind whose caller is test-classified - by test file
path, or by an is_test stamp on the source symbol, so annotation-marked
tests outside test-named paths are covered. The frontier is
instance-scoped and drained destructively: the derived coordinator
drains the per-repo resolvers and reconciles those callers regardless
of plan flags (your declaration-only shape is the regression, failing
with your exact message before the fix), and the catch-up lanes with no
global test pass behind them - the master file resolve and both
cross-repository wrappers - drain their own instances and reconcile
inline. Cold warmup stays unchanged: its throwaway resolver instances
leak nothing, and the whole-graph emission that follows warmup remains
the single owner of that scale. End-to-end regressions cover the add
and rebind flows on real files (rebind also pins that the stale
projection to the evicted subject disappears) and the cross-repository
inbound test call.

3. Legacy rows purge on upgrade. Schema v13 deletes tests-kind edges
whose target is an unresolved stub, in both the bare and the
repo-prefixed COPY-rewrite spellings, leaving pending calls and
resolved projections untouched. The seed/close/reopen upgrade test you
asked for is included; the in-memory backend needs no counterpart since
it always loads from a migrated store or a fresh index.

Full race suite on Windows: the fail-name set matches my baseline except
one known timing flake in each direction (a Swift tight-window test that
reproduces standalone on a clean checkout of the base, and one baseline
flake that happened to pass this run). The resolver, indexer, and
store_sqlite packages are clean under -race.

zzet (changes requested):
Re-review at `5910a1de`. P1-A is properly fixed, and the store data backs the premise more strongly than the PR claims. Two things still need work: one production lane where **P1-B still reproduces**, and the test integrity of the reconcile commit.

Measured read-only against a 20-repo store (9.5 GB, 7,479,444 edges):

| | count |
|---|---:|
| `tests` edges | 104,972 |
| → unresolved target (purged by v13) | **42,962 (41% of the whole tests layer)** |
| migration predicate cost | **1.35 s** |

`EXPLAIN QUERY PLAN` gives `SEARCH edges USING INDEX edges_by_kind`, so the leading-wildcard `LIKE` never scans the 7.5M-row table. **The startup-cost concern from the last round is settled** — and 42,962 is well above the ~29k the description estimates.

---

## 1. BLOCKING — P1-B still reproduces on the whole-graph master lane

`runMasterResolveFiles` drains the frontier (`multi.go:939`). `runMasterResolveHookedContext` does not: it runs `master.ResolveAllContext` at `multi.go:910` and returns at `:923` with no drain. Every whole-graph master lane funnels through it — `runMasterResolve` (`:875`), `runMasterResolveHooked` (`:881`), and `:988`. `newMasterResolver` returns a fresh resolver per call, so that frontier is garbage-collected unread.

This is reachable on a routine path: `resolveDeferredMutations`' `fullFallback` (`multi.go:783`, calling through at `:793`) is the fail-closed whole-graph catch-up taken whenever a mutation receipt comes back incomplete. `runCrossRepoResolveContext` runs right after and *does* drain — but only its own `CrossRepoResolver` frontier; it never sees what the master bound.

### Reproduce

Drop this in `internal/indexer/` and run `GOWORK=off go test -run TestReproMasterResolveAllDropsRetargetFrontier ./internal/indexer/ -v`:

```go
func TestReproMasterResolveAllDropsRetargetFrontier(t *testing.T) {
	build := func() (*MultiIndexer, *graph.Edge, graph.Store) {
		g := graph.New()
		g.AddNode(&graph.Node{ID: "repoA/pkg/foo_test.go", Kind: graph.KindFile, Name: "foo_test.go", FilePath: "repoA/pkg/foo_test.go", Language: "go", RepoPrefix: "repoA", WorkspaceID: "ws"})
		g.AddNode(&graph.Node{ID: "repoA/pkg/foo_test.go::TestFoo", Kind: graph.KindFunction, Name: "TestFoo", FilePath: "repoA/pkg/foo_test.go", Language: "go", RepoPrefix: "repoA", WorkspaceID: "ws"})
		g.AddNode(&graph.Node{ID: "repoA/pkg/foo.go", Kind: graph.KindFile, Name: "foo.go", FilePath: "repoA/pkg/foo.go", Language: "go", RepoPrefix: "repoA", WorkspaceID: "ws"})
		g.AddNode(&graph.Node{ID: "repoA/pkg/foo.go::Foo", Kind: graph.KindFunction, Name: "Foo", FilePath: "repoA/pkg/foo.go", Language: "go", RepoPrefix: "repoA", WorkspaceID: "ws"})
		call := &graph.Edge{From: "repoA/pkg/foo_test.go::TestFoo", To: graph.UnresolvedMarker + "Foo", Kind: graph.EdgeCalls, FilePath: "repoA/pkg/foo_test.go", Line: 5}
		g.AddEdge(call)
		if _, emitted := markTestSymbolsAndEmitEdges(g); emitted != 0 {
			t.Fatalf("fixture: unresolved call must not project, emitted %d", emitted)
		}
		return NewMultiIndexer(g, newTestRegistry(), search.NewNull(), nil, zap.NewNop()), call, g
	}

	t.Run("control_runMasterResolveFiles", func(t *testing.T) {
		mi, call, g := build()
		mi.runMasterResolveFiles([]string{"repoA/pkg/foo.go"}, false)
		if call.To != "repoA/pkg/foo.go::Foo" {
			t.Fatalf("fixture: call must bind, got %q", call.To)
		}
		if e := findEdge(g, graph.EdgeTests, "repoA/pkg/foo_test.go::TestFoo", "repoA/pkg/foo.go::Foo"); e == nil {
			t.Fatalf("control: expected EdgeTests projection")
		}
	})

	t.Run("subject_runMasterResolve_wholegraph", func(t *testing.T) {
		mi, call, g := build()
		mi.runMasterResolve(nil, false) // the fullFallback lane (multi.go:793)
		if call.To != "repoA/pkg/foo.go::Foo" {
			t.Fatalf("fixture: call must bind, got %q", call.To)
		}
		if e := findEdge(g, graph.EdgeTests, "repoA/pkg/foo_test.go::TestFoo", "repoA/pkg/foo.go::Foo"); e == nil {
			t.Fatalf("BUG: master ResolveAll bound the test call but no EdgeTests was reconciled")
		}
	})
}
```

At `5910a1de`:

```
--- PASS: TestReproMasterResolveAllDropsRetargetFrontier/control_runMasterResolveFiles
--- FAIL: TestReproMasterResolveAllDropsRetargetFrontier/subject_runMasterResolve_wholegraph
    zz_repro_test.go:48: BUG: master ResolveAll bound the test call but no EdgeTests was reconciled
```

### Fix

One line in `runMasterResolveHookedContext`, mirroring `:939`:

```diff
 		zap.Int("pending_admitted", stats.PendingAfter),
 		zap.Error(err))
+	mi.reconcileRetargetedTestCalls(master.TakeRetargetedTestCallFiles())
 	return err
 }
```

Verified: the repro goes green, and the full `internal/indexer` suite stays green with it applied (`ok ... 150.625s`). Worth landing the repro above as the regression test so the lane stays pinned.

---

## 2. BLOCKING — two of the four new reconcile tests pass with the entire PR reverted

Neuter all three fixes (`resolutionExempt` → `return false`; both `TakeRetargetedTestCallFiles` → `return nil`) and run the four reconcile tests:

```
--- PASS: TestIncrementalReindex_LaterResolvedCallGainsTestsProjection   <- vacuous
--- PASS: TestIncrementalReindex_RebindMovesTestsProjection             <- vacuous
--- FAIL: TestDeclarationOnlyPlanReconcilesLaterResolvedTestCall        <- genuinely pins the fix
--- FAIL: TestCrossRepoResolveReconcilesRetargetedTestCall              <- genuinely pins the fix
```

Both passers drive the change through `IncrementalReindexPaths`, whose derived plan *names the caller file* — so the pre-existing evict-and-re-emit sweep in `markTestSymbolsAndEmitEdgesForFilesLocked` (`test_edges.go:335`) produces the expected result whether or not the retarget frontier exists. This includes the stale-edge assertion in the rebind test, and it includes the headline P1-B test.

The two that fail are good; keep them. For the other two: either drop them as redundant, or re-point them at a plan shape that does not name the caller file (the declaration-only shape the third test already uses), or assert directly that the drained frontier contains `foo_test.go`.

---

## 3. BLOCKING (small) — the warm-path emission guard has no coverage

`test_edges.go:319`, the unresolved-call skip inside `markTestSymbolsAndEmitEdgesForFilesLocked`, is the guard on the incremental path taken on **every warm save**. Replace its condition with `if false` and the whole suite still passes:

```
ok  github.com/zzet/gortex/internal/indexer  107.712s
```

The only emission test, `TestMarkTestSymbolsAndEmitEdges_SkipsUnresolvedCallTargets`, calls `markTestSymbolsAndEmitEdges(g)` with no files, which routes to `emitTestEdgesAndPersistLocked` and pins only the `:712` guard.

**Fix:** a sibling test calling `markTestSymbolsAndEmitEdgesScoped(g, nil, "pkg/foo_test.go")` on the same two-call fixture, asserting `emitted == 1` and no unresolved `EdgeTests`.

---

## Non-blocking

### a. Already-bound legacy clones survive the purge

The `DELETE` matches only an unresolved `to_id`, so a clone the *old* resolver already bound — the exact motivating artifact — survives forever. Real, but small: 22 same-repo `tests` rows with a resolved target and no sibling `calls` edge in that 20-repo store. They do include genuine cross-language impossibilities, e.g.

```
torture.rb::Documented.wire             -> react-native-bridge.test.ts::method
CustomerMgt.Codeunit.al::CreateCustomer -> SharedPointer.h::TSharedPtr.Get
```

**Caution — do not widen the predicate with a `NOT EXISTS` sibling-`calls` probe.** I tried exactly that; it deletes **9,724 legitimate cross-repo projections** (9,724 of the 9,746 orphan rows are cross-repo) and it fails your own `TestOpenPurgesLegacyUnresolvedTestsEdges`, whose healthy fixture row has no sibling `calls` edge either. If you want the residue, gate on same-repo **and** cross-language; otherwise accepting it is defensible at this scale.

### b. Cold index parks the whole test corpus on the frontier

`indexer.go:3768` (`idx.resolver.ResolveAll()`) notes every bound test caller, and no cold path drains it. The first later incremental save drains all of them into `markTestSymbolsAndEmitEdgesScoped` under `ResolveMutex` — on this repo that is ~2.4k test files re-projected inside one save. Cheap fix: discard the frontier at the end of the cold whole-graph path, which already runs its own graph-wide projection.

---

## Verified good

- `resolutionExempt` being blanket on `Kind == EdgeTests` is correct, not over-broad: `internal/indexer/test_edges.go` is the **only** producer of that kind in the tree; every other reference is a consumer.
- The two `LIKE` patterns are exactly equivalent to `graph.IsUnresolvedTarget` (`HasPrefix("unresolved::") || Contains("::unresolved::")`), with no LIKE metacharacters.
- Merges cleanly onto current `main` (`217b60ca`); 656 `-race` tests pass across `resolver` / `indexer` / `store_sqlite` on the merged state.
- Two constructs that look wrong are right: `mi.mu.RLock()` in `drainRetargetedTestCallFiles` is not reentrant-unsafe (the same function already RLocks earlier), and the `len(files) > 0 &&` guard is correct — zero files means a whole-graph pass, so merging the retargeted list in would *downgrade* it to scoped.

Heads-up unrelated to correctness: #681 also bumps `currentSchemaVersion` 12 -> 13. Not silent — `validateSchemaMigrations` enforces strictly-ascending versions and `highest == current`, so whoever lands second just renumbers to 14.

pbednarcik:
All three blocking items are addressed, plus the cold-index frontier note; one commit per finding on top of a merge of current main.

1. The whole-graph master lane now drains: `runMasterResolveHookedContext` mirrors the file-scoped lane's drain before returning, so the fullFallback catch-up (and every other whole-graph master pass funnelling through it) reconciles what it bound. Your reproduction is landed as the regression test, control subtest included, and failed exactly as you showed before the one-line fix.

2. Test integrity. You were right that both `IncrementalReindexPaths`-shaped tests were vacuous - that lane's derived plan names the caller file, so the pre-existing scoped sweep produced the expected result with the frontier removed entirely. The gain scenario was already genuinely pinned by the declaration-only test, so the vacuous copy is dropped. The rebind scenario is re-pointed at the same declaration-only plan shape (the plan names only the new subject file, with a stale projection to the old location still on the books), where both the moved-projection and stale-eviction assertions require the frontier. Verified the way you did: with both `TakeRetargetedTestCallFiles` implementations neutered to `return nil`, every reconcile test in the file now fails - the master-lane subject, the declaration-only gain, the declaration-only rebind, and the cross-repo test.

3. The warm-path guard has its own pin now: a sibling of the emission test drives the same two-call fixture through `markTestSymbolsAndEmitEdgesScoped` with the caller file named, asserting one emitted edge and no unresolved clone. Replacing the scoped skip's condition with `if false` fails it.

On the non-blocking items:

- Cold-index frontier: took the cheap fix at both cold projection sites. The inline single-repo path discards the resolver's frontier right after its graph-wide projection, and the batch global passes discard the per-repo frontiers scoped to the same `scanPrefixes` set the projection covered (nil = all). A regression test pins the inline path (it failed with the frontier parked, one caller, before the fix).

- Already-bound legacy clones: accepting the residue, per your analysis. The purge stays keyed on an unresolved `to_id` only - your measurement that a sibling-`calls` probe deletes 9,724 legitimate cross-repo projections settles that widening the predicate trades 22 stale rows for real data loss, and a same-repo-and-cross-language gate does not seem worth its own migration risk at that scale.

Housekeeping: the branch is merged with current main; the schema migration renumbers to v14 behind main's v13, as you predicted, and the upgrade test seeds the pre-purge store at v13 accordingly.

Suite state: `go build` and `go vet` clean. The `-race` suite for `internal/graph/store_sqlite` passes in full on the merged head; for `internal/indexer` and `internal/resolver` the failing-name set is identical to a clean current-main control run in the same environment (46 names, the usual Windows environmental set), with zero new names and zero data races.

zzet (changes requested):
All three blocking items are genuinely closed, and I verified each the way I found them.

- **Master-lane drain.** `runMasterResolveHookedContext` now drains before returning; my reproduction is landed as `TestMasterResolveWholeGraphReconcilesRetargetedTestCalls`, control subtest included. Deleting the drain line fails it.
- **Test integrity.** The vacuous gain test is dropped and the rebind test is re-pointed at the declaration-only plan shape. With both `TakeRetargetedTestCallFiles` implementations short-circuited to `return nil`, all four reconcile tests now fail — the two `IncrementalReindexPaths`-shaped passers are gone.
- **Warm-path guard.** `TestMarkTestSymbolsAndEmitEdgesScoped_SkipsUnresolvedCallTargets` pins `test_edges.go:320`; replacing the condition with `false` fails it.

Purge residue accepted per the measurement, predicate unchanged, schema renumbered to 14 with the upgrade test seeding v13 — all correct. Build, `go vet`, `golangci-lint` clean across `indexer` / `resolver` / `store_sqlite`; those three suites pass here (148s / 10s / 44s), and `indexer` + `resolver` are also fully green under `-race` (397s / 40s, no data races) — none of the 46 Windows failures you saw reproduce on macOS.

The cold-index frontier note is the problem. I raised it as non-blocking and the fix overshot: the batch half of `68152a70` turns a *deferred* projection into a *lost* one.

## Blocking — the scoped batch discard permanently drops projections in unchanged repos

`runGlobalGraphPassesTopologyHeld` discards with the same set it projected with:

```go
marked, emitted := markTestSymbolsAndEmitEdgesScoped(mi.graph, scanPrefixes)
// "This projection covers every test caller in scanPrefixes"
mi.discardRetargetedTestCallFiles(scanPrefixes)
```

The comment is true of the projection and false of the discard, because the two are keyed on different things. The projection is keyed on the **repo of the calls edge**; the discard is keyed on the **repo of the indexer whose frontier it drains**, and it then drops that frontier *whole*, without looking at the file paths in it. Those sets are not the same, and I measured both halves.

**A per-repo resolver parks other repos' caller files.** Per-repo resolvers are never `SetScope`d — only the master lane is (`multi.go:911`) — so `idx.resolver.ResolveAll()` walks the whole shared graph. Two repos in one graph, repo-b holding the only pending test call, running repo-a's resolver:

```
call target after repoA ResolveAll: "repoB/pkg/bar.go::Bar"
repoA resolver frontier:            [repoB/pkg/bar_test.go]
```

**The scoped projection does not cover them.** With `changedPrefixes != nil`, `emitTestEdgesAndPersistLocked` reads only `graph.ReadRepoEdgesByKinds(g, prefixes, [EdgeCalls])` (`test_edges.go:755`), so a caller outside the set is never walked. Same two-repo fixture, both calls already resolved:

```
scanPrefixes={repoA}          emitted=1   repoB projection: MISSING
scanPrefixes=nil              emitted=2
```

**End to end it is a permanent loss, not a deferral.** Two real repos through `IndexAll`, repo-b's call put back to a stub with its projection evicted, nothing in repo-b changed, repo-a's resolver rebinds it, then `runGlobalGraphPasses(ctx, {repo-a}, false)`. (I drive `mi.indexers["repo-a"].resolver.ResolveAll()` directly — the same call `indexCtxRaw:3790` and `RunDeferredPasses:975` make — rather than staging a full warm-restart sequence.)

```
head                       after scoped batch: projected=false frontier=[]
batch discard removed      after scoped batch: projected=false frontier=[repo-b/main_test.go]
```

At head nothing will ever recreate that edge — no projection ran for it and the frontier that would have carried it to the next save is gone. Without the discard it is simply deferred to the next save, which is the pre-existing (and correct) behaviour. This is the same failure class the PR exists to fix, reintroduced by the optimisation.

It slipped in because that call site has no coverage. The only frontier assertion in the suite is `test_edges_test.go:587`, and it drives the inline `indexCtxRaw` path through a single-repo `idx.Index(dir)`; nothing references `discardRetargetedTestCallFiles` or checks frontier state after a global-pass batch.

**Fix — one line.** The perf motivation was entirely the *cold, full-coverage* path, where the projection genuinely is graph-wide. Gate on that:

```diff
-	mi.discardRetargetedTestCallFiles(scanPrefixes)
+	if scanPrefixes == nil {
+		mi.discardRetargetedTestCallFiles(nil)
+	}
```

Verified: the reproduction above goes from `LOST` to `RECOVERABLE`, `TestColdIndexDiscardsRetargetedTestCallFrontier` still passes (the inline `indexCtxRaw` site is untouched and is safe — its projection really is graph-wide), and the full `internal/indexer` suite stays green (132.9s).

Filtering the drained list by repo instead would also work in principle, but frontier entries are `Edge.FilePath` values, which are not reliably prefix-qualified — so the cheap correct move is to only discard when the projection was graph-wide.

Please add a pin for the surviving discard too, or this recurs: a full-coverage batch (`scanPrefixes == nil`) should leave every per-repo frontier empty afterwards. That pins the optimisation without pinning the unsafe scoped case.

Everything else on this branch looks right to me — with that one line I'm happy to approve.

internal/indexer/multi.go
```diff
@@ -1665,6 +1682,12 @@ func (mi *MultiIndexer) runGlobalGraphPassesTopologyHeld(
 		zap.Int("test_symbols", marked),
 		zap.Int("edges", emitted),
 		zap.Duration("elapsed", time.Since(testStart)))
+	// This projection covers every test caller in scanPrefixes (nil =
+	// whole graph), so the retarget frontiers the per-repo ResolveAll
+	// calls parked for those repos are already served - discard them, or
+	// the first later incremental save drains the entire test corpus into
+	// a scoped re-projection under ResolveMutex.
+	mi.discardRetargetedTestCallFiles(scanPrefixes)
```
zzet:
```suggestion
	if scanPrefixes == nil {
		mi.discardRetargetedTestCallFiles(nil)
    }
```

zzet (approved):

pbednarcik:
The pin for the surviving discard is up as #699, together with a gofmt retab of the suggestion-applied brace.

PR #680, opened 2026-08-27
csharp: EF Core ORM model attribution (models_table for the seventh ecosystem)

pbednarcik:
models_table ships for six ecosystems (go, java, python, ruby,
typescript, elixir) and C# had none: my production store holds zero
models_table edges, zero orm table nodes. EF Core is the reason a
straight java_orm port does not work - the table name can come from
three places, and two of them live in files other than the entity's:

1. `[Table("name")]` / `[Table("name", Schema = "s")]` on the entity
itself (the attribute is shared by Dapper.Contrib, so this form pays
for itself beyond EF).

2. Fluent configuration: a class implementing
`IEntityTypeConfiguration<T>` calling `builder.ToTable(...)`, or an
`OnModelCreating` body calling `modelBuilder.Entity<T>().ToTable(...)`.

3. The DbSet convention: with neither of the above, EF names the table
after the DbSet PROPERTY on the context - verbatim, not a
pluralization of the class name. `DbSet<Widget> StockWidgets` maps
`StockWidgets`.

The split follows from that: the extractor decides only the in-file
form (the attribute, java_orm's model - same `db::orm::` table nodes,
same edge meta shape, schema-qualified IDs as `db::orm::schema.table`)
and stamps the cross-file forms as facts: `ef_config_entity` /
`ef_config_table` / `ef_config_schema` / `ef_config_relation` on
config classes, `ef_fluent` entries ("entity|table|schema|relation")
on files with an OnModelCreating body. A resolver pass
(SynthCSharpEFCoreModels, registered beside the other dotnet
synthesizers) joins the facts to entity classes by unique
same-boundary class name and applies precedence fluent > attribute >
dbset: a fluent fact REWIRES the attribute edge in place (go_orm's
override model, via ReindexEdges) rather than adding a second edge,
and an entity the fluent tier claimed never falls through to the DbSet
convention. Within the fluent tier an inline OnModelCreating entry
beats a config class, matching EF's application order.

Everything fails open, refusing over guessing:

- An ambiguous entity name (two classes sharing it) joins nothing.
- Two facts in the winning fluent tier that disagree drop the entity;
the merge is two-phase and fact lists are sorted first, so the verdict
and the retained evidence site never depend on map iteration order.
- A non-literal table name (nameof, a constant, the lambda-only
ToTable overloads) stamps nothing.
- A ToTable reached through an owned-type or navigation builder names
a DIFFERENT entity's table (`builder.OwnsOne(...).ToTable(...)`, the
OwnsOne-lambda form, and the inline chained spellings, generic
included) - both the config scan and the inline chain walk refuse
these: the config side requires a plain-identifier receiver outside
any lambda, the inline side stops at subject-changing links.
- A class implementing IEntityTypeConfiguration for two entities
refuses entirely - a single-entity stamp cannot say whose ToTable the
body scan found.
- A DbSet whose type argument is a nested generic is not an entity
registration.
- OnModelCreating scanning stays inside the method body - a helper
method reached from there is a cross-method chase the extractor does
not attempt.
- The pass is idempotent: an entity already carrying a models_table
edge at its final target is left alone.

`ToView` rides as `relation: view` on the edge, so view-backed keyless
entities attribute without pretending to be tables.

Two fixes came out of validating the branch against a live store (an
isolated daemon, cold-indexing my production C# codebase plus my
acceptance fixtures) - both invisible to unit tests because they live
in the ingest pipeline:

1. The sql coverage domain defaults off, and stripSQLArtifacts then
dropped EVERY KindTable node without origin=migration, plus all edges
touching them. That silently empties the models_table layer for every
ORM ecosystem under the default config, not just this one - the six
older *_orm extractors mint the same nodes. The gate exists for the
noisy code-side string-literal SQL matching; ORM attribution is
declaration-anchored, so dialect=orm table nodes now ride through the
gate the same way migration-origin DDL already does (RED-tested).

2. In a workspace store, ingest prefixes extraction IDs with the repo -
the extractor-minted db::orm:: nodes included - while the resolver
pass minted bare IDs, so one logical table split into two identities
depending on which binding path named it. Resolver-minted table IDs
now derive from the joined entity's RepoPrefix (empty in single-repo
stores, where extraction IDs are bare too - one rule, one spelling).

With those in, the live run attributes 482 model-to-table edges on my
production codebase (456 attribute-bound, 26 DbSet-convention) where
the store previously had zero, and every acceptance scenario behaves:
attribute-only, schema-qualified, convention-pins-property-name,
separate-file fluent rewire, inline fluent, ToView, and the
ambiguous-name guard binding nothing.

On scoped (incremental) invocations the pass runs through the scoped
store with no bespoke scopedFn: a cross-file join whose other half is
outside the changed frontier joins nothing - fail-open, never wrong.
The one transient window is a fluent fact whose file is out of scope:
it cannot shadow the DbSet convention until the next full settle
rewires the edge. I judged that acceptable for a first cut and
documented it on the pass; a scopedFn that pulls the fact files into
the frontier is the follow-up if it matters in practice. I also swept
the existing models_table consumers: analyze models needs no changes
(the orm filter doc gains ecto and efcore), and the Rails
resolve pass's model-class set is gated behind a .rb-caller check
before it consults membership, so cross-language entries are inert.

Extractor version: csharp bumps to 13, because a store extracted
before the stamps has no facts for the pass to join and no content
change would mint them. Note the number collides with my other two
open branches' bumps (the PR 677 revision and the virtual-dispatch
branch each claim the next slot from their own base); whichever merges
first is fine and I will renumber the others on rebase.

Named non-goals for this round, so the review has edges: `[Column]`
property mapping, owned types (OwnsOne/OwnsMany), inheritance
strategies (TPH/TPT/TPC), and FromSqlRaw/Dapper query-text edges.

Verification: RED-first throughout (extractor fixtures for each
attribute/config/fluent shape including a mutation check on the
OnModelCreating scope gate; resolver e2e tests through the real
extractor covering the property-name-verbatim pin, the
attribute-rewire, fluent-beats-dbset precedence, ToView, ambiguity
skip, and double-run idempotency). My counted acceptance fixtures add
a round covering the general surface: attribute-only, schema form,
convention pin with a property name that matches no pluralization,
separate-file fluent override, inline fluent, ToView, and an
ambiguous-name guard cell that must never bind. Full race suite:
fail-name set identical to my Windows baseline at the merge base,
except one Swift localization timing test that fails identically on an
unmodified checkout (this branch touches no Swift or localization
code).

### DIFF
--- a/internal/indexer/coverage_strip_test.go
+++ b/internal/indexer/coverage_strip_test.go
@@ -66,3 +66,58 @@ func TestStripFunctionShape(t *testing.T) {
 		t.Error("defines edge missing after strip")
 	}
 }
+
+// TestStripSQLArtifacts_PreservesORMTableNodes: the sql domain gates
+// the NOISY code-side string-literal SQL extraction. ORM model
+// attribution (models_table) is declaration-anchored — attributes,
+// DbSet properties, schema classes — and must survive the gate the
+// same way migration-origin DDL does, or the models_table layer is
+// silently empty for every ORM ecosystem under the default config.
+func TestStripSQLArtifacts_PreservesORMTableNodes(t *testing.T) {
+	result := &parser.ExtractionResult{
+		Nodes: []*graph.Node{
+			{ID: "Models/StockCrate.cs::StockCrate", Kind: graph.KindType, Name: "StockCrate"},
+			{ID: "db::orm::stock_crates", Kind: graph.KindTable, Name: "stock_crates",
+				Meta: map[string]any{"dialect": "orm", "schema": "", "source": "csharp-orm"}},
+			{ID: "db::sqlite::guessed", Kind: graph.KindTable, Name: "guessed"},
+			{ID: "db::sqlite::from_ddl", Kind: graph.KindTable, Name: "from_ddl",
+				Meta: map[string]any{"origin": "migration"}},
+			{ID: "str::sql1", Kind: graph.KindString, Name: "q",
+				Meta: map[string]any{"context": "sql"}},
+		},
+		Edges: []*graph.Edge{
+			{From: "Models/StockCrate.cs::StockCrate", To: "db::orm::stock_crates", Kind: graph.EdgeModelsTable},
+			{From: "svc::repo.Load", To: "db::sqlite::guessed", Kind: graph.EdgeQueries},
+		},
+	}
+	stripSQLArtifacts(result)
+
+	kept := map[string]bool{}
+	for _, n := range result.Nodes {
+		kept[n.ID] = true
+	}
+	if !kept["db::orm::stock_crates"] {
+		t.Errorf("ORM table node stripped — models_table layer dies under the default config")
+	}
+	if !kept["db::sqlite::from_ddl"] {
+		t.Errorf("migration-origin table node should survive as before")
+	}
+	if kept["db::sqlite::guessed"] {
+		t.Errorf("code-side SQL table guess should still be stripped")
+	}
+	if kept["str::sql1"] {
+		t.Errorf("sql-context string registry node should still be stripped")
+	}
+	hasModels := false
+	for _, e := range result.Edges {
+		if e.Kind == graph.EdgeModelsTable {
+			hasModels = true
+		}
+		if e.Kind == graph.EdgeQueries {
+			t.Errorf("queries edge should still be stripped: %+v", e)
+		}
+	}
+	if !hasModels {
+		t.Errorf("models_table edge to the surviving ORM table node should be kept")
+	}
+}
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -32,7 +32,7 @@ var extractorVersions = map[string]int{
 	//   "go": 2,
 	"c":      generatedParserProjectionPolicyVersion, // generated parser projection covers all strictly detected table sizes
 	"php":    2,                                      // class/interface inheritance now emits typed structural edges
-	"csharp": 12,                                     // receiverless calls carry arg_count / type_arg_count for #559 (was: params parameters emit complete shape and arity evidence)
+	"csharp": 13,                                     // EF Core ORM facts: [Table] models_table edges, ef_config_* and ef_fluent stamps (was: receiverless calls carry arg_count / type_arg_count for #559)
 	"scala":  2,                                      // explicitly instantiated generic calls emit call edges
 	"go":     3,                                      // generic instantiations are marked so indexing a func value cannot bind (was: generic calls emit call edges)
 	"cpp":    2,                                      // templated and namespace-qualified calls emit call edges
--- a/internal/indexer/extractor_version_test.go
+++ b/internal/indexer/extractor_version_test.go
@@ -62,9 +62,13 @@ func TestStaleLangsDetection(t *testing.T) {
 		if got := ExtractorVersionStaleLangs(`{"csharp":10}`); !reflect.DeepEqual(got, []string{"csharp"}) {
 			t.Errorf("stored pre-params C# version = %v, want [csharp]", got)
 		}
+		// A store extracted before the EF Core ORM stamps must re-extract
+		// unchanged .cs files: the models_table pass reads facts ([Table]
+		// edges, ef_config_* and ef_fluent stamps) that only re-extraction
+		// can mint.
 		for _, path := range []string{"src/Handler.cs", "Views/Page.razor", "Views/Page.cshtml"} {
-			if got := merkleSaltFor(path); got != "csharp@12" {
-				t.Errorf("C# extractor salt for %s = %q, want csharp@12", path, got)
+			if got := merkleSaltFor(path); got != "csharp@13" {
+				t.Errorf("C# extractor salt for %s = %q, want csharp@13", path, got)
 			}
 		}
 		if got := merkleSaltFor("src/Handler.php"); got != "php@2" {
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -1809,11 +1809,21 @@ func applyMigrationExtraction(relPath string, src []byte, result *parser.Extract
 // defaults off because string-literal pattern matching against
 // db.Get / db.Query / db.Exec produces false positives when
 // domain code shares method names (cache.Get, etc.).
+//
+// Two table-node origins survive the gate, because neither is the
+// noisy code-side matching the gate exists for: migration-origin DDL
+// (unambiguous CREATE TABLE — see applyMigrationExtraction), and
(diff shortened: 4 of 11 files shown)


zzet (changes requested):
Thanks for the substantial work here — EF Core model attribution is valuable, the PR documents its intended contract unusually well, and the immediate extraction cases are well tested. I found several exact-head failures outside those tests, though, so I think the following need to be addressed before merge.

## 1. P1: only apply `IEntityTypeConfiguration<T>` mappings that EF actually activates

`ResolveCSharpEFCoreModels` consumes every stamped configuration class at `internal/resolver/csharp_efcore_models.go:73`, even when no DbContext applies it.

Reproduction:

```csharp
class UnusedWidgetConfig : IEntityTypeConfiguration<Widget>
{
    public void Configure(EntityTypeBuilder<Widget> b)
        => b.ToTable("unused_widgets");
}

class Ctx : DbContext
{
    public DbSet<Widget> Widgets { get; set; }
    // No ApplyConfiguration / ApplyConfigurationsFromAssembly
}
```

Actual graph: `Widget -> db::orm::unused_widgets`.
EF result: the DbSet convention, `Widget -> Widgets`.

Microsoft's `IEntityTypeConfiguration<T>` documentation explicitly says the configuration must then be applied in `OnModelCreating`: https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.ientitytypeconfiguration-1?view=efcore-10.0

Suggested fix: extract application-site facts for `ApplyConfiguration` and `ApplyConfigurationsFromAssembly`, activate only reachable configurations per DbContext, and retain application order relative to inline calls. If activation cannot be proven, fail open.

## 2. P1: make the EF projection a real scoped reconciliation

The synthesizer is registered without a `scopedFn` at `internal/resolver/framework_synth.go:951`. A config-only incremental run cannot see the entity-file-owned attribute edge and adds a second fluent edge. A later full settle rewires the first edge but leaves both duplicates:

```text
before:
  Widget -> attribute_widgets

scoped config change:
  Widget -> attribute_widgets
  Widget -> fluent_widgets

full settle:
  Widget -> fluent_widgets
  Widget -> fluent_widgets
```

Related exact-head failures from the same add/rewire-only design:

- Removing `ToTable` after an attribute edge was rewired leaves the stale fluent mapping forever.
- At `csharp_efcore_models.go:140`, target-ID equality returns early. `[Table("widgets")]` overridden by `ToView("widgets")` therefore keeps `binding=attribute` and loses `relation=view`.

Suggested fix: add a dedicated scoped implementation that expands changed fact sites to referenced entities, loads all existing `models_table` edges for each affected entity, and replaces the complete per-entity projection atomically. Preserve reversible base attribution, coalesce legacy duplicates, and compare complete desired edge semantics rather than only `To`.

## 3. P1: reduce facts within the repository/entity boundary and preserve rejected claims

`csharpEFMergeFluentFacts` groups facts globally by bare entity name at `csharp_efcore_models.go:233-239`, before `sameBoundaryCandidates` runs.

Reproduction:

```text
repo-a: Widget -> widgets_a
repo-b: Widget -> widgets_b
```

Actual: zero mappings because the two independent repository facts conflict globally. Expected: one mapping in each repository.

Also, when same-boundary fluent facts conflict, the reducer emits no winner but the entity falls through to the DbSet loop at line 177, producing a convention edge despite the documented “drop rather than guess” contract.

Suggested fix: resolve/group by boundary plus entity identity, and return an explicit winner/rejected/claimed decision so a rejected fluent tier cannot fall through to DbSet.

## 4. P1: restrict configuration scanning to the real Configure method and parse C# arguments

`csharpEFConfigTableCall` (`csharp_orm.go:182-210`) scans the whole class and stops at the first plain-identifier `ToTable` call.

Reproduction:

```csharp
void Render(Grid grid) {
    grid.ToTable("not_an_ef_mapping");
}

void Configure(EntityTypeBuilder<Widget> builder) {
    builder.ToTable("widgets");
}
```

Actual mapping: `not_an_ef_mapping`; expected: `widgets`. Two legitimate `builder.ToTable` calls also retain the first, although EF applies fluent configuration in call order and the latest conflicting call wins.

The argument parser at `csharp_orm.go:238-255` also drops the standard named schema spelling:

```csharp
builder.ToTable("blogs", schema: "blogging");
```

Actual target: `db::orm::blogs`; expected: `db::orm::blogging.blogs`. With both arguments named, no fact is emitted. This is the syntax used in Microsoft's EF documentation: https://learn.microsoft.com/en-us/ef/core/modeling/entity-types

Suggested fix: locate `Configure(EntityTypeBuilder<T> parameter)`, scan only that body, require the receiver to be that parameter, process calls in source order, and parse named `name`/`schema` CST arguments explicitly. The same parsing should cover `ToView` and inline mappings.

## 5. P1: invalidate every ecosystem affected by the new strip policy

`stripSQLArtifacts` now retains `dialect=orm` table nodes for all ORM extractors, but `extractor_version.go:35` bumps only C#.

Using the exact base extractor-version snapshot, head reports only:

```text
stale languages: [csharp]
```

Go, Java, Python, Ruby, TypeScript, and Elixir stores created before this fix therefore keep their already-stripped ORM nodes and edges until each source file happens to change.

Suggested fix: preferably add a global post-extraction/admission-policy version to the file-cache fingerprint. With the current infrastructure, bump every ORM-emitting language (including JavaScript if its TypeORM path applies) and add a persisted warm-store upgrade regression.

## 6. P2: replace the unescaped fact protocol and tighten syntax validation

Additional exact-head failures:

- `ef_fluent` serializes arbitrary identifiers as `entity|table|schema|relation|line` at `csharp_orm.go:353`. `ToTable("widget|rows")` decodes as `db::orm::rows.widget`, not `db::orm::widget|rows`.
- Any unrelated method merely named `OnModelCreating` is accepted without validating a DbContext, ModelBuilder parameter, or receiver.
- C# escaped/Unicode/verbatim literal source spellings are stored rather than their runtime values.
- The interface regex accepts names such as `NotIEntityTypeConfiguration<Widget>`.

Suggested fix: use structured fact metadata rather than a delimiter protocol; validate EF methods/interfaces through CST structure; and decode supported C# string literal forms correctly, or reject forms that cannot yet be decoded.

## Validation and missing coverage

The hosted checks, affected-package race suite, and `go vet` all pass. The exact-head regression tests above fail, confirming coverage gaps around application sites, scoped/full reconciliation, removal/reversion, multi-repository facts, named arguments/call order, structured-fact persistence, and warm-store upgrades.

I found no auth, secret, code-execution, off-by-one, or introduced dead-code issue in the change.

zzet:
Thanks again, @pbednarcik. I carried this work forward in #688 and addressed the six requested changes there, while preserving your original EF Core attribution commits. I left this PR open so the relationship and review history remain visible.

PR #683, opened 2026-08-27
fix(sqlite): retain bounded WAL read pool

slikts:
## Review question

Is an opt-in stable SQLite read pool an acceptable recovery mode for hosts where reader-connection churn makes WAL-index shared-memory mappings unreliable?

## Problem

On the affected host, both stock v0.63.3 and v0.63.8 reproduced a `SIGBUS` while the WAL index expanded at frame 4063. The failure was not caused by one unusually large repository: it recurred during normal multi-repository indexing as idle read connections were shed and reopened.

Constraining the store to one reader avoided that churn but is not viable. Store iterators can invoke callbacks that perform nested reads, and the single-reader variant deadlocked deterministically.

## Change

- Add the opt-in `GORTEX_SQLITE_STABLE_READ_POOL=1` recovery mode.
- Keep the existing read-pool bound (at most four connections), but retain opened readers by setting the idle bound to the open bound.
- Preserve the dedicated writer connection and the concurrency required for nested reads.
- Log when the mode is active, document its resource tradeoff, and test that all bounded readers remain open after use.

The default remains unchanged. When enabled, the process retains up to four reader page caches and shared-memory mappings instead of shedding idle readers.

## Validation

- `go test ./internal/graph/store_sqlite`
- A forced retrack of three repositories completed all resolver, enrichment, and global passes, processing 472,585 pending resolver edges and 351,966 rebinds.
- During that run, the WAL-index shared-memory file grew from 32 KiB to 819,200 bytes without a `SIGBUS`, panic, or SQLite error.
- A fresh graph query succeeded after the completed retrack.

The exact fix revision used for the stress run is retained in this branch's ancestry; the branch head is a conflict-free projection onto current `main`.


slikts:
Follow-up validation invalidated this mitigation. Stable-pool mode was confirmed active at 18:15:59; at 19:55:29 the daemon again crashed with `SIGBUS` in `AtomicLoadNUint16` → `_walFindFrame`. This shows retaining the bounded read pool does not prevent the fault. It does not establish whether it changes failure frequency. Closing because the proposed recovery-mode claim is unsupported.

PR #689, opened 2026-08-28
`batch_symbols` JSON callers/callees cap at 9 and did not say so

zorionarrillaga:
`handleBatchSymbols` walks the 1-hop neighbourhood with a hard-coded
`Limit: 10` (`tools_coding.go`). BFS counts the seed against that limit,
so the id lists cap at nine. `SubGraph.Truncated` was already computed
(`engine.go:1636`) and then discarded — the JSON response was just
`callers: [...]` / `callees: [...]`.

The cap stays. The JSON entry now carries `callers_truncated` /
`callees_truncated` when the cut binds. Complete under-cap pages (8
callers, 8 callees) grow no new key.

No `*_total`. On this path `TotalNodes` is `len(allNodes)` after the cap
(`engine.go:1634`), so a `TotalNodes - 1` figure would report 9 on a
15-caller graph — the page size, not the neighbourhood.

Full callers/callees remain `get_callers` / `get_call_chain`. GCX
(`encodeBatchSymbols`) still omits the lists; this stamp is the JSON
wire.

```
HEAD:    15 callers → 9 ids, no flag
this PR: 15 callers → 9 ids, callers_truncated=true
         8 callers  → 8 ids, no flag
```

`go test ./internal/mcp -run TestBatchSymbols_CallersTruncatedFlag -v`

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -281,7 +281,7 @@ over a very large tree, or `ask` against a slow local model.
 | Tool | Description |
 |------|-------------|
 | `get_symbol_source` | Source code of a single symbol (80% fewer tokens than Read). Returns `tokens_saved` per call. `compress_bodies` stubs bodies (with an optional `keep` subset); `max_lines` salience-truncates to a control-flow skeleton |
-| `batch_symbols` | Multiple symbols with source/callers/callees in one call |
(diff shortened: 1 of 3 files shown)


zzet:
@zorionarrillaga thank you for your contribution!

Issue #692, opened 2026-08-29
Failed graph-ingest mutation receipt fail-closes change.detect repo-wide with no reconcile path; reindex does not clear it


tiendungdev:
## Summary

A file mutation whose **disk write succeeded** but whose **graph ingest failed** (`context deadline exceeded`) leaves a `graph_status: failed` head in the mutation ledger. That head fail-closes `change(operation="detect")` **for the entire repo, from every MCP session**, and there is currently no operation that clears or reconciles it:

- full reindex and scoped `workspace_admin.reindex(paths=[...])` complete but do **not** clear the failed head;
- daemon restart does not clear it (the ledger persists in the store);
- a superseding mutation on the same path replaces the head, but its own ingest can fail again under load, re-locking detect;
- a sweep of the MCP facade (`change`, `edit`, `workspace_admin`, `session` capabilities) and the CLI (`gortex --help`, `gortex edit --help`) found no clear/reconcile verb.

Related: #691 (deadline/cancellation asks overlap; this issue is about recovery, not latency).

## Environment

- Gortex `v0.63.8+2a33f3d`, Windows 11, native MCP.
- Single-repo workspace at the time of the detect refusals: 3,058 files, 148,117 nodes, 1,398,981 edges, `health_score=100`, clean path liveness.
- Daemon under load during the re-failure: 15 connected MCP sessions, FTS doc count fluctuating (post-shrink store churn).

## Sequence observed

1. `edit_file` mutation on a 15.7 KB bash script: `disk_status: committed`, `graph_status: failed`, receipt `commit-3`, caller `mutation_id` preserved. The bytes on disk are correct (verified by blob SHA).
2. Any `change.detect` (any scope, any session) now refuses:

   ```text
   change detection refused a stale graph: graph freshness unavailable:
   failed receipt=mutation-3 repo="..." path="...\scripts\dev\....sh"
   generation=6 error="context deadline exceeded"
   ```

   Reproduced from a second, freshly oriented MCP session — the state is daemon-owned, not session-owned.
3. Full reindex (from the original session) and scoped `workspace_admin.reindex(paths=[<file>])` both completed; `change.receipt` for the path still reports `graph_status: failed`.
[...]

PR #693, opened 2026-08-29
fix(mcp): resolve superseded failed mutation receipts on later success

tiendungdev:
## Summary

Fixes the rolling repo-wide `change.detect` lockout described in #692.

A mutation whose disk write succeeded but whose graph ingest failed leaves a terminally failed receipt in `Server.mutationReceipts`. `awaitMutationFreshnessForRepos` then refuses `change.detect`/`change.impact` for the whole repository until the receipt's 10-minute retention lapses — even though waiting can never heal a terminal failure. Under load, each failed ingest of an actively edited file re-arms the barrier for another retention window, so agents observe a seemingly permanent, uncleareable lockout.

## Change

- `trackMutationTicket`: when a ticket completes successfully (`Err == nil && Reindexed`), call the new `resolveSupersededFailedReceipts`, which drops **terminally failed** receipts for the **same path** with a **lower generation**. The graph then reflects newer bytes than the failed generation ever wrote, so the stale failure no longer describes a real freshness gap. Pending receipts and failures at or above the succeeded generation are untouched, as are failures on other paths.
- Barrier error message: when a terminal failure is reported, append that terminally failed generations do not recover by waiting and clear on a later successful mutation of the path or on retention expiry. (Existing tests assert substrings, so this is additive.)

## Tests

- `TestMutationFreshnessSuccessResolvesSupersededFailures`: success resolves an older failed receipt for the same path; failures on another path and failures newer than the success survive and still fail the barrier; the superseded receipt no longer appears in the error.
- `TestTrackMutationTicketResolvesSupersededFailures`: end-to-end through `trackMutationTicket` — a successful ticket for the path removes the stale failed receipt and keeps its own receipt for retention.
- Deleting the resolve call makes the end-to-end test fail (verified), so the tests bind the behavior.

## Verification

- `go vet ./internal/mcp/`: clean.
- Focused freshness tests pass on Windows and on Linux (`golang:1.26` container).
- Full `./internal/mcp` suite in the Linux container: `TestDetectChanges_EchoesMeasuredScope` and `TestFileMutationsReindexSynchronouslyAcrossConsecutiveEdits` fail **identically with and without this patch** (both pass in isolation), so they appear to be pre-existing load-sensitive failures, not regressions from this change.

## Scope (explicit, so #692 stays open)

- `workspace_admin.reindex` goes through `multiIndexer.IncrementalReindexRepo` and never touches `s.mutationReceipts`, so a scoped reindex of the file still does not resolve a failed receipt. Not addressed here.
- Under sustained load where every successive ingest also fails, nothing here clears anything — the resolve only runs on a success. This converts "blocked until retention" into "blocked until the next success", not the full rolling-lockout fix. Retry-with-backoff and degrade-instead-of-refuse remain open in #692.
- As of the review follow-up, superseded failures are resolved **in place** (stamped with the superseding apply) rather than deleted, so `change.receipt` keeps reporting truthfully.

## Notes

The reproduction and measurements behind this are in #692 (including the correction comment with the retention/in-memory analysis). The other asks in that issue — an explicit reconcile operation, degrade-instead-of-refuse, and printing a queryable id in the refusal — are intentionally left out of this PR since they involve API-surface decisions.

### DIFF
--- a/internal/mcp/edit_serialization.go
+++ b/internal/mcp/edit_serialization.go
@@ -171,12 +171,42 @@ func (s *Server) trackMutationTicket(ticket *indexer.MutationTicket) *mutationRe
 		receipt.result = result
 		receipt.completed = true
 		receipt.mu.Unlock()
+		if result.Err == nil && result.Reindexed {
+			s.resolveSupersededFailedReceipts(receipt)
+		}
 		close(receipt.done)
 		time.AfterFunc(mutationReceiptRetention, func() {
 			s.mutationReceipts.Delete(receipt.id)
 		})
 	}()
 	return receipt
+}
+
+// resolveSupersededFailedReceipts drops terminally failed receipts for a path
+// once a later generation of the same path has been applied successfully. The
+// graph then reflects newer bytes than the failed generation ever wrote, so
+// the stale failure no longer describes a real freshness gap — keeping it
+// would only fail freshness barriers that waiting cannot heal, because a
+// terminal error never completes differently. Pending receipts and failures
+// at or above the succeeded generation are left untouched.
+func (s *Server) resolveSupersededFailedReceipts(succeeded *mutationReceipt) {
+	succeededPath := filepath.Clean(succeeded.path)
+	s.mutationReceipts.Range(func(key, value any) bool {
+		other, ok := value.(*mutationReceipt)
+		if !ok || other == succeeded {
+			return true
+		}
+		if other.generation >= succeeded.generation || filepath.Clean(other.path) != succeededPath {
+			return true
+		}
+		other.mu.RLock()
+		terminalFailure := other.completed && (other.result.Err != nil || !other.result.Reindexed)
+		other.mu.RUnlock()
+		if terminalFailure {
+			s.mutationReceipts.Delete(key)
+		}
+		return true
+	})
 }
 
 func (r *mutationReceipt) outcome(pending bool) mutationReindexOutcome {
@@ -313,16 +343,19 @@ waitLoop:
 	}
 
 	issues := make([]string, 0, len(receipts))
+	hasTerminalFailure := false
 	for _, receipt := range receipts {
 		select {
 		case <-receipt.done:
 			outcome := receipt.outcome(false)
 			switch {
 			case outcome.Err != nil:
+				hasTerminalFailure = true
 				issues = append(issues, fmt.Sprintf(
 					"failed receipt=%s repo=%q path=%q generation=%d error=%q",
 					receipt.id, receipt.repo, receipt.path, receipt.generation, outcome.Err.Error()))
 			case !outcome.Reindexed:
+				hasTerminalFailure = true
 				issues = append(issues, fmt.Sprintf(
 					"failed receipt=%s repo=%q path=%q generation=%d error=%q",
 					receipt.id, receipt.repo, receipt.path, receipt.generation, "reindex not confirmed"))
@@ -351,6 +384,11 @@ waitLoop:
 			message += "; "
 		}
 		message += issue
+	}
+	if hasTerminalFailure {
+		message += "; terminally failed generations do not recover by waiting — " +
+			"they clear when a later mutation of the same path succeeds or when " +
+			"the receipt retention lapses"
 	}
 	return fmt.Errorf("%s", message)
 }
--- a/internal/mcp/edit_serialization.go
+++ b/internal/mcp/edit_serialization.go
@@ -172,7 +172,7 @@ func (s *Server) trackMutationTicket(ticket *indexer.MutationTicket) *mutationRe
 		receipt.completed = true
 		receipt.mu.Unlock()
 		if result.Err == nil && result.Reindexed {
-			s.resolveSupersededFailedReceipts(receipt)
+			s.resolveSupersededFailedReceipts(receipt.path, receipt.generation, result)
 		}
 		close(receipt.done)
 		time.AfterFunc(mutationReceiptRetention, func() {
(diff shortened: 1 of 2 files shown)


zzet (changes requested):
Thanks for this — the diagnosis in #692 and the follow-up correction are the good kind of bug report, and the core of this fix is right. I verified the load-bearing argument independently rather than reading it: a completed waiter always has `RequestedGeneration <= AppliedGeneration`, because `completeMutationWaiters` only completes waiters whose `generation <= appliedGeneration` (`watcher.go:1774`, and the storm path at `:2042`). So a success at requested generation G guarantees the graph reflects at least G's bytes, hence strictly newer than any failure below G — and thresholding on the *requested* rather than the applied generation is the conservative side of that. The claim holds.

Build, `go vet`, `golangci-lint ./internal/mcp/...` clean. The full `./internal/mcp` suite passes here on macOS, and again under `-race` (198s, no data races) — including `TestDetectChanges_EchoesMeasuredScope` and `TestFileMutationsReindexSynchronouslyAcrossConsecutiveEdits`, which you saw fail in your Linux container. So those look environmental rather than pre-existing-and-known; worth a second look on your side, but not this PR's problem.

Mutants: removing the `resolveSupersededFailedReceipts` call fails `TestTrackMutationTicketResolvesSupersededFailures`, dropping the generation guard fails `TestMutationFreshnessSuccessResolvesSupersededFailures`, and dropping the path guard fails it too. Three of the four guards bind. The fourth is finding 2 below.

One regression I need fixed before this lands, and one gap.

## 1. `change.receipt` reports a superseded failure as `pending`, permanently

Deleting the receipt removes the only thing that could refresh the mutation-commit ledger's graph half. `mutationStatusPayload` refreshes a record's status via `mutationReceiptState(record.pendingReindexReceipt())` and simply skips the refresh when the lookup misses — so a record whose edit response timed out stays `graph_status: "pending"` forever once its receipt is dropped.

Same probe, natural sequence (edit response times out → record is `pending` with receipt `mutation-3`; the ingest fails terminally; nobody polls in that window; a later generation of the same path succeeds), asserting only what `change.receipt` would print:

```
origin/main   after supersede   graph_status="failed"
this branch   after supersede   graph_status="pending"
```

`failed` is at least true of that generation and actionable. `pending` is neither — and `mutationStatusGuidance` for a committed write says *"if graph_status is not \"fresh\" the graph is still catching up"*, so the caller is told to keep waiting for a generation that will never complete. That is exactly the failure `mutationStatusPayload`'s own doc comment says the refresh exists to prevent ("every pending receipt would read as pending forever, which is precisely the stale answer this tool exists to avoid").

The window is not exotic — this PR widens it. Today an agent learns about the failed ingest from the detect refusal and then goes and reads the receipt. After this change the refusal is gone, so there is less reason to poll in the interval where the status is still truthful.

Worth contrasting with the batch path, which already models a vanishing receipt correctly: `refreshBatchGraph` clears `file.ReindexReceipt` on a lookup miss and re-admits the file (`batch_transaction.go:909-922`). `mutationStatusPayload` has no such fallback.

**Suggested fix — don't delete, resolve.** Stamp the superseding outcome onto the failed receipt instead of removing it. Prototyped locally; it is a few lines in place of the `Delete`:

```go
other.mu.Lock()
if other.completed && (other.result.Err != nil || !other.result.Reindexed) {
    other.result = indexer.MutationResult{
        RequestedGeneration: other.generation,
        AppliedGeneration:   applied.AppliedGeneration,
        Reindexed:           true,
    }
}
other.mu.Unlock()
```

With that, the barrier clears exactly as it does now (I ran a barrier assertion against both variants — both pass), the receipt stays queryable, and `change.receipt` reports:

```
prototype     after supersede   graph_status="fresh"   (reindexed=true, applied=9)
```

which is not just better than this branch, it is better than `main` — the path genuinely *is* fresh once a later generation lands, and `failed` was already the wrong answer there. It also mirrors what the watcher itself does: `completeMutationWaiters` resolves every earlier waiter with the later apply's result rather than discarding it, so the MCP ledger would end up describing supersession the same way the layer beneath it does.

One caveat if you take this: read the succeeded result *before* the `Range` and pass it in (you already have `result` in hand at the call site in `trackMutationTicket`) rather than taking `succeeded.mu.RLock()` inside the loop while holding nothing — two concurrent resolves each holding their own read lock and reaching for the other's write lock is a deadlock shape that does not need to exist.

## 2. The "pending receipts are untouched" guard is the one that isn't tested

`terminalFailure := true` — i.e. delete every earlier-generation receipt for the path, in flight or not — leaves the entire `./internal/mcp` suite green. Every receipt in `TestMutationFreshnessSuccessResolvesSupersededFailures` is already completed, so nothing exercises it.

That is the guard whose failure direction is unsafe: dropping a *pending* receipt removes a live freshness gap from the barrier rather than a dead one. Test that fails under that mutant and passes at head:

```go
func TestMutationFreshnessSuccessKeepsPendingSamePathReceipts(t *testing.T) {
	s := &Server{mutationSafetyWait: time.Millisecond}
	pendingFreshnessReceipt(s, "receipt-inflight", "repo-a", "/repo-a/file.go", 4)

	succeeded := pendingFreshnessReceipt(s, "receipt-success", "repo-a", "/repo-a/file.go", 9)
	completeFreshnessReceipt(succeeded, indexer.MutationResult{
		RequestedGeneration: 9, AppliedGeneration: 9, Reindexed: true,
	})
	s.resolveSupersededFailedReceipts(succeeded)

	if _, loaded := s.mutationReceipts.Load("receipt-inflight"); !loaded {
		t.Fatal("a still-pending receipt for the same path was dropped by the resolve")
	}
	err := s.awaitMutationFreshnessForRepos(context.Background(), "repo-a")
	if err == nil || !strings.Contains(err.Error(), "receipt-inflight") {
		t.Fatalf("barrier no longer reports the in-flight receipt: %v", err)
	}
}
```

(Adjust the receipt-identity assertion if you take the resolve-in-place route.)

## Scope, so #692 does not get closed by accident

`trackMutationTicket` is reached only from `mutationReindexState` and the batch transaction. `reindex_repository` goes through `multiIndexer.IncrementalReindexRepo` and never touches `s.mutationReceipts`, so a scoped `workspace_admin.reindex(paths=[<file>])` still leaves the failed receipt blocking detect — request 1 in the issue, and the half of your own issue comment that said "and when a direct reindex of that path succeeds". Your PR body doesn't claim it and the new barrier text is careful not to promise it, which I appreciate; I'd just say so explicitly in the PR so the issue stays open for it.

Same for the shape you actually hit: under sustained load where each successive ingest also fails, nothing here clears anything, because the resolve only runs on a success. That is fine as a scoped first step — it converts "blocked until retention" into "blocked until the next success", which is a real improvement — but it is not the rolling-lockout fix, and the issue should keep tracking retry-with-backoff and degrade-instead-of-refuse separately.

The message addition is good and I'd keep it as is.

tiendungdev:
Both findings addressed in 027959d6 — thank you for the independent verification of the generation invariant and for the prototype.

1. **Resolve in place, not delete.** Took your prototype nearly verbatim: the stamp happens under the failed receipt's write lock only, and the succeeded result is passed **by value** from the call site in `trackMutationTicket`, so the sweep holds no other lock — the succeeded receipt skips itself via the generation guard rather than pointer identity. `change.receipt` for a superseded failure now stays queryable and reports `fresh` with the superseding `AppliedGeneration`, matching your prototype's output and `completeMutationWaiters` semantics one layer down.

2. **Pending-receipt guard test added.** `TestMutationFreshnessSuccessKeepsPendingSamePathReceipts` per your sketch, plus an assertion that the in-flight receipt is not marked completed by the resolve. All four mutants now bind: call removed → end-to-end test fails; generation guard dropped and path guard dropped → resolve test fails; stamping pending receipts (with `completed` forced) → the new test fails. Restored tree: the freshness family and the `Mutation|Batch|Receipt|FileOps` test families pass, `go vet` clean.

**Scope** — agreed, and now stated explicitly in the PR body: this change does not touch the `workspace_admin.reindex` path (`IncrementalReindexRepo` never sees `s.mutationReceipts`), and under sustained ingest failure nothing here clears anything until the next success. #692 should stay open for reindex-resolves-receipts, retry-with-backoff, and degrade-instead-of-refuse.

On the two container failures passing on macOS and under `-race` for you: I'll chase the Linux-container environment on my side separately — agreed it's not this PR's problem.

zzet (approved):

PR #694, opened 2026-08-29
Feat/julia extractor followups

NilsWildt:
## Summary

Follow-up to #675, addressing point 8 ("Cover the advertised 'Full' forms or narrow
the documentation") and the decoding bug zzet left for the
Julia AST work. Both done: the extractor now handles the
previously-missing call, macro, and export/containment shapes, and `docs/languages.md`
is narrowed to enumerate exactly what is and isn't covered. Bumps the Julia extractor
version to 3 so existing graphs re-extract.

## Changes

- **Chained/qualified callees decoded from CST children, not source text**
  (`juliaCalleeName` / new `juliaUnwrappedName`): `get(cfg).run(x)` now records
  `unresolved::run` instead of `unresolved::get(cfg).run`, and a chain broken across
  lines no longer leaks a newline into the target. Quoted operators normalise —
  `Base.:+` → `unresolved::Base.+`, `Base.:(==)` → `unresolved::Base.==`.
- **Parametric-constructor callees** (`Vector{Int}(xs)` → `unresolved::Vector{Int}`)
  via new `juliaParametrizedCallee`, rebuilding the `{…}` list from children so a
  multi-line parameter list can't leak a newline into the target.
- **Module-qualified macro invocations** (`Base.@time f(x)` → `unresolved::Base.time`,
  `Meta["macro"]`), including alias resolution (`import Foo as F; F.@spawn` names `Foo`).
- **Macro and operator exports** recorded verbatim (`export @m, ⊗` records `@m` and `⊗`),
  plus the Julia 1.11 `public` list in `Meta["public"]`
  (`handleExport`/`handlePublic` refactored onto a shared `recordModuleNames`).
- **Containment edges** `member_of` for module-level constants and nested modules
  (previously only functions/types carried them; constants carried `scope_mod` on Meta
  only, which is not traversable).
- **External-receiver `member_of` honesty**: `function Base.show` inside a module points
  its `member_of` at `unresolved::Base` rather than a node-shaped id for a type this file
  doesn't declare.
- **Explicit `@doc "text" object` docstrings** attach the same way triple-quoted ones do,
  and call sites inside documented macro arguments keep their edges.
- **`docs/languages.md`**: Julia table row enumerates the newly-covered forms; the Julia
  section gains a "What is **not** covered" list (parametric constructor *definitions*,
  callable objects, `const X::Int`, string-interpolation calls, same-line twins) following
  the Rust-row house style.

## Testing

- [x] All tests pass (`go test -race ./...`) — branch related packages green
  (`internal/parser/languages`, `internal/indexer`); 
  (`internal/gitcmd`, `internal/agents/opencode`) fail only under full-suite parallel load and pass on isolated rerun, same as noted in #675. Neither touches Julia.
- [x] Benchmarks run if performon-correctness change

## Checklist

- [x] Code follows existing patds + `scope_mod`, `<Type>.<init>`
  constructor spelling, JS/TS per-binding import cap, python `base_path` convention)
- [x] No unnecessary abstractio
- [ ] Language extractor includes `Meta["methods"]` for interfaces (N/A — Julia has none)
- [x] Methods have `EdgeMemberOtype (qualified methods → receiver; constructors → type; constants/nested modules → module)

Cheers

### DIFF
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -154,7 +154,10 @@ but only when the string sits immediately above the documented object —
 the adjacency Julia itself enforces, where a blank line or an own-line
 comment detaches the string and leaves the definition undocumented. A
 string at the top of a function body is executable code, not
-documentation. The stored text is the first PROSE paragraph, skipping the
+documentation. The explicit `@doc "text" object` form (which is what
+Julia lowers every docstring to) attaches the same way, with the text
+taken from inside the macro call. The stored text is the first PROSE
+paragraph, skipping the
 indented signature block Julia's convention opens a docstring with.
 
 What is **not** covered:
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -84,7 +84,7 @@ on interface nodes stores the expected method set for implementation matching.
 | Dart | Full | Full | Classes/Enums/Mixins/Extensions | Abstract interface | Full | Full | Full |
 | OCaml | Full | Full (class) | Types/Modules | Module types | open | Full | Full |
 | Lua | Full | Full (M.func/M:method) | - | - | require() | Full | Full |
-| Julia | Full (long + short form, `where` syntax) | Full (qualified `Base.show`, operators) | Structs/abstract/primitive + fields | - | Full (`using`/`import`/`include`, selective lists) | Full (incl. broadcast, macro calls) | `const` |
+| Julia | Full (long + short form, `where` syntax) | Full (qualified `Base.show`, operators) | Structs/abstract/primitive + fields | - | Full (`using`/`import`/`include`, selective lists incl. macros/operators) | Full (incl. broadcast, macro calls, `Vector{Int}` constructors) | `const` (with `member_of`) |
 
 ### Rust specifics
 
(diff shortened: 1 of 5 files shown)


zzet (changes requested):
Requesting changes for six confirmed issues:

1. **P1 — CI is blocked by an ineffectual assignment.** `internal/parser/languages/julia.go:895` assigns `ownerTarget = ownerID`, but every path immediately overwrites it. This reproduces the current `ineffassign` failure locally and in CI.

2. **P1 — Same-file module methods get an unresolved owner.** At `julia.go:896`, qualified receivers are resolved only through `lookupType`. Modules live in `st.nodes`, not `st.types`, so `module M ... end; function M.f() ... end` emits `member_of -> unresolved::M` although `file::M` exists. Track lexical modules and add an assertion for `M.f -> file::M`.

3. **P2 — Explicit documentation drops constants.** `julia.go:320` does not dispatch a direct `const_statement`, so `@doc "text" const X = 1` reaches assignment handling without `isConst`, dropping the constant and its documentation.

4. **P2 — Multi-segment qualified macros lose their call edge.** `julia.go:1361` requires the receiver to be a single identifier. Valid calls such as `Base.Threads.@threads` and `A.B.@m` therefore disappear. Decode dotted receivers recursively and cover a three-segment qualifier.

5. **P2 — Arbitrary qualified `@doc` macros are misclassified.** `julia.go:280` checks only the property name. `Foo.@doc` consequently attaches `Meta["doc"]`; restrict recognition to the supported bare/Core/Base documentation macros and add a negative test.

6. **P2 — Nested parametric callees retain raw source formatting.** `julia.go:722` canonicalizes only the outer parameter list. A valid nested type such as `Vector{Tuple{Int,\nString}}(x)` can still create a newline-dependent unresolved target. Canonicalize nested type expressions recursively.

The focused parser and indexer tests pass, but lint fails and none of the semantic cases above currently has regression coverage.

zzet (approved):

Issue #696, opened 2026-08-29
Remote-proxied tool calls don't enforce the origin session's tool-surface policy or workspace boundary


timkjr:
Follow-up from your latest review on PR #649 (findings #2 and #3's remote half) — opening this separately since it's a design question about the multi-server/federation protocol, not a bug in that PR's diff.

## The gap

When a tool call routes to a remote gortex daemon (`Router.RouteToolCall` → proxy path, reached from `internal/server/handler.go`'s `handleToolCall`, `internal/mcp/streamable/transport.go`'s `tryRouteToolCall`, and the daemon dispatcher's `tryProxyToolCall`), two things about the *origin* session are not currently transmitted to, or enforced by, the remote:

1. **The origin session's tool-surface policy.** A tool hidden by the calling session's `facade-v1`/hide-mode policy is checked locally by `checkToolGate` inside the wrapped handler — but a proxied request never reaches that handler locally. `Router.RouteToolCall` only gates on `remoteEnabledIn`/`IsEffectful` (whether the remote is enabled at all, and whether the tool is read-only) before forwarding — nothing checks whether *this session* is even allowed to call the tool in the first place. The remote executes under its own default policy, whatever that happens to be.

2. **A trusted, enforceable workspace/cwd ceiling.** The origin resolves a `cwd`/workspace for routing purposes (`peekRouteContext`, `RouteInputs.Cwd`), but that's used only to *pick* a remote destination — it isn't carried as an authoritative boundary the remote must honor once it receives the request. A remote hosting multiple repositories could, in principle, answer using a different repo than the one the caller's resolved cwd pointed to.

Neither of these is a bug in a single line — the origin-side gate simply doesn't exist on the remote-proxy branch at all, by construction: the local branch relies entirely on the downstream handler enforcing policy, and there is no equivalent downstream handler on the remote-proxy branch to rely on.

## Why I'm asking rather than guessing at a fix

This needs an actual protocol decision, not just a code patch:

- Where does the origin-policy check belong? The natural shared choke point is `ProxyDecision.Decide` (used by all three call sites above), but it doesn't currently have access to the MCP `Server`'s session-policy machinery (`IsToolEnabledForSession` etc.) — that'd mean threading a callback/dependency into `daemon.Router`/`ProxyDecision` the way `localExecute` is already wired in.
[...]

PR #697, opened 2026-08-30
fix(watcher): detect NFS mounts and make WatchConfig.Enabled actually disable watching

timkjr:
This fixes two related bugs that together explain why repos on NFS/SMB mounts silently stop being watched, with no error surfaced anywhere obvious.

**1. `slowWatchMount` never recognized NFS**

The slow-mount gate only checked for WSL2's 9p/drvfs magic number and (within a WSL context) CIFS. A native Linux host with a repo on a plain NFS mount fell through every safety net — `NFS_SUPER_MAGIC` (`0x6969`) was never checked. Native fsnotify was then attempted regardless, reliably failing `confirmWatchActive`'s 5s readiness window, leaving the repo with neither fsnotify nor the poller fallback until a manual untrack+track.

**2. `WatchConfig.Enabled` didn't actually disable watching**

`config.Default()` ships `Watch.Enabled: false` — checked into gortex's own `.gortex.yaml` too. But `Watcher.Start` only gated the *safe* branches (the slow-mount skip-to-poller path from fix #1, and launching the poller as a fallback) behind `Enabled`. It never gated the actual attempt to start native fsnotify, which ran unconditionally regardless of the flag. So with the shipped default, every repo blindly attempted raw fsnotify, racing the same 5s timeout with no safety net and no fallback — meaning fix #1 was effectively dead code unless a user manually set `watch.enabled: true`, which is undocumented anywhere. This PR adds an early return when `Enabled` is false, matching what the surrounding comments already claimed the flag did.

**Known related gap, not fixed here:** `slow_mount_other.go` (the non-Linux build) still returns `false` unconditionally, so a macOS host with a repo on an NFS mount gets no slow-mount protection either — the same class of bug this PR closes for Linux. I don't have a Mac to test against, so I've left that alone rather than guess at the right `statfs`-equivalent check; happy to take a pass at it if a maintainer can confirm the right approach (or point at existing platform-detection code I should mirror).

Verified: `go test -race ./...` and `golangci-lint run` both clean. Manually confirmed on a live NFS-mounted multi-repo daemon — all repos went from erratic per-boot watcher-attach failures to 100% success (fsnotify where safe, poller fallback on slow mounts).

Separately flagging for maintainer discussion, not changed here: is `Enabled: false` the right *default*? A code-intelligence tool whose core value is live indexing arguably shouldn't ship watching off by default with no documentation pointing at the flag.

### DIFF
--- a/internal/indexer/poller_test.go
+++ b/internal/indexer/poller_test.go
@@ -209,8 +209,10 @@ func TestPoller_DetectsGitHeadMoveMissedByFsnotify(t *testing.T) {
 
 // TestPoller_RespectsWatcherDisableKnob verifies the poller honours
 // the per-repo watcher-disable knob: when WatchConfig.Enabled is
-// false, Start must not create a poller — the disabled repo gets no
-// fallback either.
+// false, Start must not create the alongside-fsnotify poller. This
+// is not the whole story on a slow mount, where the degraded-path
+// pollers still start unconditionally regardless of Enabled — see
+// TestWatcher_ShippedDefaultStillWatches and the slow-mount tests.
 func TestPoller_RespectsWatcherDisableKnob(t *testing.T) {
 	dir := t.TempDir()
 	writeTestFile(t, filepath.Join(dir, "main.go"), "package main\n\nfunc Main() {}\n")
--- a/internal/indexer/slow_mount_linux.go
+++ b/internal/indexer/slow_mount_linux.go
@@ -4,42 +4,40 @@ package indexer
 
 import (
 	"os"
-	"strings"
 	"syscall"
 )
 
 // slowWatchMount reports whether path lives on a filesystem where native
-// fsnotify is unreliable or prohibitively slow — notably a Windows drive
-// surfaced into WSL2 via 9p/drvfs (inotify events arrive late or never), or
-// an SMB/CIFS share. On such a mount the watcher disables fsnotify and
-// relies on the adaptive poller + git hooks. GORTEX_FORCE_FSNOTIFY=1 forces
-// native fsnotify on regardless.
+// fsnotify is unreliable or prohibitively slow — a Windows drive surfaced
+// into WSL2 via 9p/drvfs (inotify events arrive late or never), an SMB/CIFS
+// share (whether mounted directly or via WSL2), or an NFS mount (the kernel
+// inotify backend does not reliably notice changes made by another NFS
+// client, and even same-client notifications can arrive late enough to miss
+// the watcher's readiness window entirely — see confirmWatchActive's 5s
+// timeout). On such a mount the watcher disables fsnotify and relies on the
+// adaptive poller + git hooks, which are mount-agnostic.
+// GORTEX_FORCE_FSNOTIFY=1 forces native fsnotify on regardless.
 func slowWatchMount(path string) bool {
 	if path == "" || os.Getenv("GORTEX_FORCE_FSNOTIFY") == "1" {
-		return false
-	}
-	if !runningUnderWSL() {
 		return false
 	}
 	var st syscall.Statfs_t
 	if err := syscall.Statfs(path, &st); err != nil {
 		return false
 	}
-	switch int64(st.Type) {
+	return isSlowMountFSType(int64(st.Type))
+}
+
+// isSlowMountFSType is the magic-number check slowWatchMount applies to a
+// statfs result. Factored out so it can be unit-tested directly against
+// known-bad magic numbers without needing a live WSL2, SMB, or NFS mount in
+// the test environment.
+func isSlowMountFSType(fsType int64) bool {
+	switch fsType {
 	case 0x01021997, // V9FS_MAGIC — 9p, WSL2's drvfs transport for Windows drives
-		0xFF534D42: // CIFS_MAGIC — SMB/CIFS share
+		0xFF534D42, // CIFS_MAGIC — SMB/CIFS share
+		0x6969:     // NFS_SUPER_MAGIC — NFS v3/v4
 		return true
 	}
 	return false
-}
-
-// runningUnderWSL reports whether the process is inside the Windows
-// Subsystem for Linux, probed from /proc/version's microsoft/WSL marker.
-func runningUnderWSL() bool {
-	b, err := os.ReadFile("/proc/version")
-	if err != nil {
-		return false
-	}
-	v := strings.ToLower(string(b))
-	return strings.Contains(v, "microsoft") || strings.Contains(v, "wsl")
 }
--- a/internal/indexer/slow_mount_test.go
+++ b/internal/indexer/slow_mount_test.go
@@ -1,18 +1,8 @@
+//go:build linux
+
 package indexer
 
 import "testing"
-
-// TestSlowWatchMount_NormalMountNotDegraded guards the safe default: a
-// normal local filesystem (the test temp dir) must never be flagged as a
-// slow mount, so fsnotify is only disabled on a genuine slow-mount type.
-func TestSlowWatchMount_NormalMountNotDegraded(t *testing.T) {
-	if slowWatchMount(t.TempDir()) {
-		t.Error("a normal local mount must not be flagged slow (fsnotify must stay enabled)")
-	}
-	if slowWatchMount("") {
-		t.Error("an empty path must not be flagged slow")
-	}
-}
 
 // TestIsSlowMountFSType pins the magic-number check slowWatchMount applies
 // to a statfs result, independent of the host actually having a WSL2, SMB,
--- a/internal/indexer/slow_mount_test.go
+++ b/internal/indexer/slow_mount_test.go
@@ -4,12 +4,42 @@ import "testing"
 
 // TestSlowWatchMount_NormalMountNotDegraded guards the safe default: a
 // normal local filesystem (the test temp dir) must never be flagged as a
-// slow mount, so fsnotify is only disabled on a genuine WSL2 9p/SMB mount.
+// slow mount, so fsnotify is only disabled on a genuine slow-mount type.
 func TestSlowWatchMount_NormalMountNotDegraded(t *testing.T) {
 	if slowWatchMount(t.TempDir()) {
 		t.Error("a normal local mount must not be flagged slow (fsnotify must stay enabled)")
 	}
 	if slowWatchMount("") {
 		t.Error("an empty path must not be flagged slow")
+	}
+}
+
+// TestIsSlowMountFSType pins the magic-number check slowWatchMount applies
+// to a statfs result, independent of the host actually having a WSL2, SMB,
+// or NFS mount available to probe. NFS (0x6969, NFS_SUPER_MAGIC) previously
+// went undetected on any non-WSL2 host: slowWatchMount's own filesystem-type
+// check was gated behind a WSL-only probe, so a native Linux host with a
+// repo on an NFS mount fell through every safety net — the fsnotify backend
+// then reliably failed its 5s readiness wait, and because that failure path
+// returns before the adaptive poller is ever constructed, the repo ended up
+// with neither fsnotify nor the poller: no update mechanism at all until a
+// manual untrack+track.
+func TestIsSlowMountFSType(t *testing.T) {
+	cases := []struct {
+		name string
+		typ  int64
+		want bool
+	}{
+		{"ext4/xfs/local (typical, unlisted)", 0xEF53, false},
+		{"V9FS (WSL2 9p/drvfs)", 0x01021997, true},
+		{"CIFS/SMB", 0xFF534D42, true},
+		{"NFS_SUPER_MAGIC", 0x6969, true},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			if got := isSlowMountFSType(c.typ); got != c.want {
+				t.Errorf("isSlowMountFSType(%#x) = %v, want %v", c.typ, got, c.want)
+			}
+		})
 	}
 }
--- a/internal/indexer/slow_watch_mount_test.go
+++ b/internal/indexer/slow_watch_mount_test.go
@@ -1,17 +1,31 @@
 package indexer
 
-import "testing"
+import (
+	"runtime"
+	"testing"
+)
 
 // TestSlowWatchMount_NormalMountNotDegraded guards the safe default: a
-// normal local filesystem (the test temp dir) must never be flagged as a
-// slow mount, so fsnotify is only disabled on a genuine slow-mount type.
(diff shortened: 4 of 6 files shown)


zzet (changes requested):
Thanks for this — the investigation behind it is genuinely good, and you found **two** real bugs, not one. The NFS detection half is correct and I want it. But the `WatchConfig.Enabled` half has to change before this can merge: as written it disables live indexing for every repo the daemon tracks, under the shipped default.

## Blocker: the `Enabled` early return turns watching off everywhere

The premise that `Enabled` gates watching isn't quite right, and the daemon is the case that proves it. `cmd/gortex/daemon_state.go:987` builds the per-repo watch configs straight from repo config, with no override:

```go
watchCfgs[prefix] = state.configManager.GetRepoConfig(prefix).Watch
```

The only place in the tree that sets `Watch.Enabled = true` is `cmd/gortex/mcp.go:507`, the `gortex mcp --watch` path. Since `config.Default()` ships `enabled: false`, the new early return fires for every repo under `gortex daemon`: no fsnotify, no poller, no live indexing at all.

I verified this rather than reading it. A test that drives `config.Default().Watch` through `Start()` fails on this branch and passes on `main`:

```
this branch:  file change never reached the graph — repo has neither fsnotify nor poller
main:         change detected, live indexing works
```

It also silently defeats the staleness alarm. The early return yields `nil`, so `MultiWatcher.Start` records `started = true`, and `WatchedRepos()` then reports `live == configured` — which means the warning at `daemon_state.go:1013`, added for exactly this symptom, never fires:

```go
// Reporting the configured count here made an install where every watcher had
// failed to start look fully watched, so the one signal that the graph was
// going stale never fired.
```

The daemon would log `daemon: watching repos=N configured=N` and publish `watched_repos: N` while nothing is watching anything.

The underlying reason for the misread is fair: **`Enabled` has never meant "watch at all"** — it gates the adaptive poller only, and fsnotify has always run regardless. `TestPoller_RespectsWatcherDisableKnob` (`internal/indexer/poller_test.go:214`) is the contract. The flag is badly named, which is the actual defect you walked into.

## What this PR gets right — please keep both

1. **NFS detection.** `slowWatchMount` being gated behind `runningUnderWSL()` meant `NFS_SUPER_MAGIC` (`0x6969`) and native-Linux CIFS were never checked. Correct diagnosis, correct magic numbers, and factoring out `isSlowMountFSType` for direct testing is the right call.
2. **Ungating the degraded-path pollers.** Both the slow-mount branch and the `confirmWatchActive` readiness-failure branch built the fallback poller only `if w.config.Enabled` — so under the default, a repo whose fsnotify died got *neither* mechanism. That is a second genuine dead-repo bug and your removal of those two gates is exactly right. It's the part of your report I'd most want fixed.

## Suggested fix (verified)

Drop the early return, keep both degraded-path pollers ungated as you have them, and re-gate only the final alongside-healthy-fsnotify poller:

```go
// Enabled is the opt-in only where fsnotify is LIVE — there the poller is a
// belt-and-braces extra for what fsnotify misses, and a repo may decline it.
// The degraded paths above start it unconditionally: fsnotify is dead there,
// so declining it means the repo silently goes stale.
if w.config.Enabled {
    w.poller = newPoller(w, w.indexer, w.logger)
    w.poller.Start()
}
```

With that, `TestPoller_RespectsWatcherDisableKnob` and a new daemon-default regression test both pass, and the full indexer suite is green (1573 passed).

## Why the suite didn't catch it

Your `go test -race ./...` result is accurate — all 1649 indexer + server tests pass on this branch. The gap is that every watcher test hardcodes `Enabled: true` (see `setupWatcher`), so the daemon's actual default is never exercised. Please add a test that drives `config.Default().Watch` through `Start()` and asserts a file change reaches the graph; that's the guard that would have caught this.

## Please split out the tool-promotion commits

Four of the eight commits and seven of the twelve files here are the deferred-MCP-tool promotion work (`SetToolPromoter` / `getToolOrPromote`), which the description doesn't mention. It's substantively sound — `EnsureToolPromoted` already exists, `Promote` is mutex-guarded, and hidden tools are never deferred so the hide gate isn't bypassed — but it's unrelated to the watcher fix and needs to be reviewed on its own. Two things to carry into that PR:

- **Concurrency:** `lazyToolRegistry.Promote` marks `promoted[name]` under the lock but calls `promoteFn` outside it, so a concurrent second first-call for the same tool gets `false` back and skips the retry lookup — a spurious "not registered". The established idiom at `cmd/gortex/daemon_mcp.go:155` uses that boolean for bookkeeping only and proceeds regardless; worth matching.
- **Session surface:** `daemon_mcp.go` deliberately checks `IsToolEnabledForSession` before touching the process-global lazy registry. The new `newLocalToolExecutor` path in `server_router.go` calls `EnsureToolPromoted` directly, skipping that guard.

## On your two open questions

**Is `Enabled: false` the right default?** Given it actually gates the poller, yes — fsnotify on, poller off, and the poller auto-engaging when fsnotify is dead once your fix #2 lands. What should change is the name, not the default. A follow-up renaming it to something like `FallbackPoller` (with the old key accepted for a release) would be very welcome.

**`slow_mount_other.go` on macOS:** leaving it alone was the right instinct, and no, don't mirror the Linux approach — Darwin's `statfs` carries `f_fstypename` as a string, so the check there is a name comparison against `nfs` / `smbfs` / `webdav` rather than a magic number. Happy to take that as a separate PR from you if you want it; it doesn't need to block this one.

One thing I checked and am *not* asking you to change: `int64(st.Type)` would truncate `CIFS_MAGIC` on a 32-bit host, but only amd64 and arm64 ship, and both have `Statfs_t.Type int64`. Not a problem.

Thanks again for chasing this down to the actual mechanism — the report quality made it fast to verify.

timkjr:
Thanks for the detailed review — you were right on all counts, and it exposed a real gap in the test suite that I've closed.

**The blocker (`Enabled` disabling everything)** — fixed exactly as suggested. Dropped the `!Enabled` early return entirely. Both degraded-path pollers (slow-mount fallback, inotify/FD-exhaustion fallback) now start unconditionally, as they did before my regression. Only the final poller — the one running alongside healthy fsnotify — stays gated on `w.config.Enabled`:

```go
if w.config.Enabled {
    w.poller = newPoller(w, w.indexer, w.logger)
    w.poller.Start()
}
```

**New test** — added `TestWatcher_ShippedDefaultStillWatches`, which drives the real `config.Default().Watch` (asserting `Enabled == false` rather than hardcoding it, so this can't drift silently) through `Start()`, confirms the poller is nil, and asserts a file change still reaches the graph — proving it arrives via fsnotify, not the poller. `TestPoller_RespectsWatcherDisableKnob` and the full `internal/indexer` suite pass under `-race`.

**Split the tool-promotion commits out** — done. This branch is now rebased to 4 commits against current `main`, touching only `internal/indexer/{watcher.go, slow_mount_linux.go, slow_mount_test.go, slow_watch_mount_test.go}`. The deferred-MCP-tool-promotion work belongs with #649, which is already open and further along (it's since picked up a session-identity fix this branch's copy didn't have) — I'll bring your two flagged bugs (the `Promote` concurrency issue and the `newLocalToolExecutor` session-policy bypass) over there rather than duplicating/regressing that work here.

**On the two open questions** — agreed on both. `Enabled: false` is the right default; I'll open the `FallbackPoller` rename as a follow-up rather than bundling it here. And leaving `slow_mount_other.go` alone was intentional — happy to take the Darwin `f_fstypename` follow-up as a separate PR if useful, just let me know.

timkjr:
Following up on the two bugs from your split-out ask — both are already fixed, no new PR needed.

`lazyToolRegistry.Promote`'s concurrency issue and `newLocalToolExecutor`'s session-surface bypass turned out to be exactly round-1 findings #1 and #3 from your own review on #649 (that branch carried a stale pre-review snapshot of the same code). #649 merged with both fixed — the lock now wraps mark-and-register in one critical section (`TestPromote_ConcurrentCallersNeverFalse404` pins it), and the dispatch path calls `EnsureToolPromotedForSession` instead of the bare `EnsureToolPromoted`, so a hidden-session tool is never promoted on its behalf (`TestLocalExecutor_HiddenSessionDeniedWithoutPromotion`/`...EvenWhenAlreadyLive`). Verified both under `-race` on merged `main`.

For the record, since it's adjacent: #649's round 2 flagged the bigger issue (remote-proxied calls skipping the origin session's policy/workspace-boundary enforcement) as a "potential security issue," and the fix that landed only closes the local half by design — the remote/federation half is tracked separately in #696, still open. Not something this PR needs to touch, just flagging so it doesn't look dropped.

zzet:
Round 2 — the branch doesn't have the fix described above.

I went to verify it and found `89e360b2` (current head) unchanged from what I reviewed. Checking the timeline, the force-push landed at `21:15:55`, nine seconds before the comment at `21:16:04`, and it carries the same four commits. Best guess is a local commit that never got pushed, or a push to a different branch — worth a `git log origin/fix-nfs-watch-detection` on your side.

What's actually on the head:

- `internal/indexer/watcher.go:436` — the `if !w.config.Enabled { … return nil }` early return is still there
- `TestWatcher_ShippedDefaultStillWatches` — not present anywhere on the branch
- The tool-promotion split **did** land — the PR is now four files, all under `internal/indexer/`. Thank you for that, and for chasing the two bugs into #649 rather than duplicating them.

## The blocker, re-verified against current `main` (`217391f8`)

I wrote the regression test you described and ran it three ways:

| tree | result |
|---|---|
| `main` | PASS — the change reaches the graph |
| this branch as submitted | FAIL — `file change never reached the graph — repo has neither fsnotify nor poller` |
| this branch + the fix you described | PASS |

The wiring is unchanged: `internal/config/config.go:1782` ships `Enabled: false`, `.gortex.yaml:24` repeats it, `cmd/gortex/daemon_state.go:987` passes it through verbatim, and `cmd/gortex/mcp.go:506` is still the only `Enabled = true` in the tree.

**The fix you described is the right one** — I applied it locally (drop the early return, keep both degraded-path pollers ungated, re-gate only the final alongside-fsnotify poller at `watcher.go:612`) and the full `internal/indexer` suite is green under `-race`, `TestPoller_RespectsWatcherDisableKnob` included. It just needs to land on the branch.

The test, for reference:

```go
wcfg := config.Default().Watch
require.False(t, wcfg.Enabled, "precondition: the shipped default is Enabled=false")
wcfg.Paths, wcfg.DebounceMs = []string{dir}, 50

w, err := NewWatcher(idx, wcfg, zap.NewNop())
require.NoError(t, err)
require.NoError(t, w.Start([]string{dir}))
t.Cleanup(func() { _ = w.Stop() })

// Enabled gates the adaptive poller only; fsnotify must still run.
assert.Nil(t, w.poller, "Enabled=false must not run an adaptive poller")

require.NoError(t, os.WriteFile(filepath.Join(dir, "main.go"), []byte(changed), 0o644))
select {
case <-w.Events():
case <-time.After(15 * time.Second):
	t.Fatal("file change never reached the graph — live indexing is dead under the shipped daemon default")
}
```

Worth noting why CI can't catch this: `setupWatcher` (`watcher_test.go:42`) hardcodes `Enabled: true`, so every existing watcher test opts out of the daemon's actual default. The branch as submitted passes all of `TestWatcher|TestPoller|TestSlowWatch|TestMultiWatcher`.

## Three smaller things on what is on the branch

1. `TestPoller_RespectsWatcherDisableKnob`'s doc comment says "the disabled repo gets no fallback either". Ungating the degraded-path pollers — which is correct — makes that untrue on a slow mount. Please update the comment along with the code, so the next reader doesn't take it as the contract.

2. `slow_watch_mount_test.go` now depends on the host filesystem in a way it didn't before. On a non-WSL host the old code returned `false` unconditionally, so the test could not fail; now a runner whose `TMPDIR` sits on NFS would fail it. Not worth blocking on, but asserting through `isSlowMountFSType` against a known-local magic would be sturdier than probing `t.TempDir()`.

3. On `slow_mount_other.go` — don't bother with the Darwin `f_fstypename` check. There's a better fix that covers every platform at once: the plain 5s-timeout arm at `watcher.go:551` still returns an error with no poller, which is exactly the failure you described from live NFS. Giving that arm the same poller fallback the inotify/FD-exhaustion arm just got makes the whole class self-healing regardless of magic-number coverage, macOS included. I'll file that separately; it's out of scope here.

## Checked and dismissed

- Deleting `runningUnderWSL` — no remaining references, and `slow_mount_linux.go` plus its test cross-compile clean for `linux/amd64` and `linux/arm64`.
- `int64(st.Type)` truncating `CIFS_MAGIC` on a 32-bit arch — only amd64/arm64 ship, and `Statfs_t.Type` is `int64` on both.
- `paths[0]` being the only probed path — `MultiWatcher` always passes a single-element `[]string{rootPath}`, so there's no multi-path blind spot.
- `0x6969` is `NFS_SUPER_MAGIC`, no collision.

You were right that `watch.enabled` is undocumented — `docs/multi-repo.md:79` documents the `watch:` block and omits `enabled:`. Worth a line there whenever the rename follow-up happens.

The NFS detection is good work and I want it in. Push the `Enabled` fix plus that test and I'll take it.

timkjr:
Confirmed — the fix I described in the last comment never actually made it into a commit. Sorry for the noise; that force-push only re-landed the same 4 commits, nothing new. It's pushed now, head is `30878fbe`.

**The blocker** — fixed as you diagnosed and I described. Dropped the `!Enabled` early return entirely. Both degraded-path pollers (slow-mount fallback, inotify/FD-exhaustion fallback) start unconditionally. Only the poller alongside a healthy fsnotify backend stays gated on `w.config.Enabled`. `TestWatcher_ShippedDefaultStillWatches` is in this push — it's your regression test, adapted to drive `config.Default().Watch` directly rather than through `setupWatcher` (see below on why).

**Three smaller things:**
1. Fixed `TestPoller_RespectsWatcherDisableKnob`'s doc comment — no longer claims a disabled repo gets no fallback at all.
2. Hardened `TestSlowWatchMount_NormalMountNotDegraded`: on Linux it now skips the live `t.TempDir()` statfs probe and relies on `TestIsSlowMountFSType`'s known-magic-number assertions instead, per your suggestion. Non-Linux platforms keep the probe since `slowWatchMount` is a static stub there and can't flake.
3. Left the Darwin/generic-5s-timeout poller fallback alone, since you said you're filing that separately.

**On why CI didn't catch this** — you're right that `setupWatcher` hardcodes `Enabled: true`, so none of the existing suite exercised the shipped default. `TestWatcher_ShippedDefaultStillWatches` builds its `Watcher` directly from `config.Default().Watch` instead of going through `setupWatcher`, specifically so it exercises `Enabled: false` for real rather than opting out of it like every other test in the package.

`go vet` and the full `internal/indexer` suite pass under `-race`, this test included.

zzet (approved):

PR #705, opened 2026-08-30
savings: book the retrieval half of the ledger

zzet:
## The report

A user on Discord: "I forced Claude to use gortex, but I didn't see my savings move." Their agent then blamed the facade — claiming `explore` / `search` / `read` are separate implementations that never call the recording tools, "intentional by design, not a bug."

That diagnosis is wrong in its mechanism and right in its conclusion. The facade forwards the same `ctx` to the same legacy handler under the legacy tool name (`facade_tools.go` `invokeFacadeSpec`), so `read(operation: "file")` does book — as `read_file`. Nothing about the facade breaks attribution.

The real problem is coverage plus a degenerate baseline, and it reproduces without any facade at all.

## Reproduced

A fresh daemon in an isolated `$HOME`, driven over real MCP frames:

| | before | after |
|---|---|---|
| recorded calls | 2 | 4 |
| tokens saved | 2,640 | 19,238 |
| ratio | 46.0% | 83.0% |
| cost avoided | $0.0132 | $0.0962 |

Same fixture repo, same six-call session (`search.symbols`, `search.text`, `relations.usages`, `trace.call_chain`, `read.file`, `read.source`).

On a small fixture, the "before" column is already flattering: with a two-file repo, the only calls that record are `read_file` (which books `saved = 0` by construction) and `get_symbol_source`, and the machine-wide ledger on my own dev box had **0 rows** across every table despite months of daily MCP use.

## Root cause

1. **Coverage.** `tokenStats.record` had five call sites yielding six tool names. 180 of the 187 facade-v1 operations reached none of them. That includes `explore` — the surface the instruction profile mandates as the first call of every task, which ships full fenced source bodies — and the entire `search` / `relations` / `trace` surface, whose stated purpose is replacing a grep-then-read sweep.
2. **Degenerate baseline.** The one tool an agent calls constantly, `read_file`, sets `fullFile := returned` unless a transform shrank the response, so an ordinary whole-file read books `saved = 0`. The tools that recorded were mostly recording zeros.

Net effect: the ledger measured the one operation that cannot save anything, and ignored the ones that do.

## The fix

Generalise the existing counterfactual from one file to the file **set** a retrieval page stands in for:

```
returned = tokens of the page actually shipped
fullFile = Σ whole-file tokens of the distinct files the page cites
```

`recordRetrievalSavings` runs at two dispatch points — `invokeFacadeSpec` for facade calls (booking under `spec.Legacy`, so both surfaces share a per-tool bucket) and the shared tool middleware for direct legacy calls. They cannot double-count: the facade holds the *unwrapped* handler, and facade names are not in the allow-list.

Citations are read back from all three wire formats Gortex emits — GCX1 (path columns located by name from the block header, with a column-count guard so prose riders like the momentum note are never mistaken for rows), JSON (via the existing `collectGraphFilePaths`), and TOON. GCX1 is the default for `claude-code` sessions and TOON is the default shape of `search_text` / `get_repo_outline` / `nav`, so skipping either would have left the same hole in a different place.

### What it deliberately does not claim

The ledger's failure mode must be "Gortex looks worse than it is", never "Gortex invented savings". Two limits, both under-reporting on purpose:

- **Once per session, per file.** A file's whole-file baseline is credited at most once, whichever call surfaces it first. A re-query over the same corner of the repo books nothing. This bounds everything a session can ever claim to the cost of reading its repo once. The read-family recorders now claim their file in the same set, so the two halves of the ledger never bill the same file twice.
- **Capped per call** at 8 distinct cited files. A 50-hit `find_usages` page does not mean the caller would have opened 50 files; it would have opened the few hits it cared about, and we cannot know which.

A call that credits no new file books **nothing at all** rather than a zero-baseline row — we did not measure that call's counterfactual, so we do not assert one.

Membership in the retrieval allow-list is a judgement about the counterfactual, so it is explicit rather than inferred. Absent by design: the six already-instrumented tools (they measure their own baseline exactly), `export_context` (it delegates to `handleSmartContext`, which books), the `response.*` re-cut tools (they re-cut a page already shipped), and every metadata / control / write tool.

### Also fixed: a mis-priced baseline

`recordFileBaselineSavings` priced a file by scaling its byte count through the chars-per-token ratio of the **response** — a marshalled node list or editing-context bundle, whose ratio is nothing like the source it stands in for. `EstimateFromSample`'s own contract says the sample must be "a smaller chunk of the same content". Every `get_file_summary` and `get_editing_context` baseline was wrong. It now calibrates on a bounded head of the file itself.

## Tests

`internal/mcp/savings_retrieval_test.go` — facade `search` / `relations` book under their legacy names; per-session per-file credit; cross-session independence; the read-family/retrieval double-bill guard; the per-call cap; the allow-list; all four wire-format shapes plus the absolute-path and prose-rider exclusions.

Every test was mutation-verified: neutering `recordRetrievalSavings` fails 4, neutering the `creditFile` dedupe fails 3, removing the cap fails 1.

`go test` and `go test -race` green on `internal/mcp`, `internal/savings`, `cmd/gortex`; `golangci-lint` clean.

## Follow-ups, not in this PR

- `get_symbol_source`, `batch_symbols` and `smart_context` record **before** their `if_none_match` check, so a polling client mints repeat observations. `read_file` / `get_file_summary` / `get_editing_context` already return before accounting; those three should match.
- `read_file`'s `returned` counts only the content string, excluding the JSON envelope and escaping it actually ships — so `tokens_returned` under-states wire cost.
- `export_context` books its observation under `smart_context`, because it delegates the whole retrieval to that handler.
- `smart_context` in `fidelity:"graded"` mode returns *more* source and books *less*: its record call is gated on `!graded`.
- Unrelated, found while validating: `relations.callers` on a Go repo with gopls attached killed the daemon with `fatal error: concurrent map writes` in `semantic/lsp.(*Provider).openDocument` (provider.go:2789), reached via `confirmSymbolRefsOnDemand`. Pre-existing on main; worth its own issue.

### DIFF
--- a/cmd/gortex/doctor_runtime.go
+++ b/cmd/gortex/doctor_runtime.go
@@ -549,9 +549,10 @@ func printSavings(w io.Writer, sv doctorSavings) {
 			}
 		}
 	}
-	fmt.Fprintln(w, "    note: only the read-family tools write here (saved = whole-file tokens −")
-	fmt.Fprintln(w, "          returned). explore / search / relations / trace record nothing, and")
-	fmt.Fprintln(w, "          reading a whole file records 0 — so this describes those calls only.")
+	fmt.Fprintln(w, "    note: saved = whole-file tokens − returned. The read-family tools book the")
+	fmt.Fprintln(w, "          one file they stand in for; explore / search / relations / trace book")
+	fmt.Fprintln(w, "          the files their page cites, each credited once per session. Reading a")
+	fmt.Fprintln(w, "          whole file still records 0 — it displaces nothing.")
 	fmt.Fprintln(w)
 }
 
--- a/cmd/gortex/savings.go
+++ b/cmd/gortex/savings.go
@@ -40,10 +40,13 @@ sessions: Today, Last 7 days, and All time. Each bucket shows a 16-cell
 saved/total bar, percentage saved, raw token counts, and the USD value of
 the tokens avoided (priced against popular models).
 
-Savings accumulate every time a source-reading MCP tool — read_file,
+Savings accumulate on two shapes of call. A read-family tool — read_file,
 get_file_summary, get_editing_context, get_symbol_source, batch_symbols,
-smart_context — returns a summary, symbol, or compressed view that stands
-in for a full-file read. The ledger lives in the machine-global sidecar
+smart_context — books the whole file its response stands in for. A
+retrieval tool — explore, the search / relations / trace families,
+get_repo_outline, get_artifact and friends — books the set of files its
+page cites, each file credited at most once per session so a re-query
+never mints savings twice. The ledger lives in the machine-global sidecar
 database (~/.gortex/sidecar.sqlite); flat-file ledgers from older
 releases (savings.json / savings.jsonl under the cache dir) are imported
 once and renamed *.bak.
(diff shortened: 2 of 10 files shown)


zzet:
Filed the follow-ups so they are not lost with this PR:

- #710 — `get_symbol_source` / `batch_symbols` / `smart_context` record before their `if_none_match` check
- #711 — `read_file`'s `tokens_returned` excludes the JSON envelope and escaping it ships
- #712 — `export_context` books its observation under `smart_context`
- #713 — graded `smart_context` returns more source and records none of it
- #714 — daemon dies with `fatal error: concurrent map writes` in `lsp.Provider.openDocument` (unrelated to this PR, found while validating it)
- #715 — proposal: make the reported number defensible (provenance, honest headline, calibrated counterfactual). Includes the per-symbol whole-file double billing this PR deliberately left alone.

PR #716, opened 2026-08-31
fix(mcp): let a scoped reindex resolve the failed receipt for that path

tiendungdev:
Implements **request 1 of #692**, and follows on from #693.

## The gap

A mutation whose disk write succeeded but whose graph ingest failed leaves a terminally failed freshness receipt, and that receipt fail-closes `change.detect`/`change.impact` for the **whole repository** until retention lapses.

`reindex_repository` is the recovery a caller reaches for first — and it does not work. It re-reads the file into the graph but never touches `s.mutationReceipts`, so the barrier keeps reporting a gap the graph no longer has. (Verified: `mutationReceipts` is referenced only in `edit_serialization.go` and the batch path; the reindex route never sees it.)

## The change

A successful **scoped** re-parse is exactly the evidence a failed receipt is waiting for — the graph has now read the current bytes of that path — so the handler resolves it, **in place**, the same way a superseding mutation does since #693.

**Kept deliberately narrow, and this is the part worth reviewing:**

- `IncrementalReindexRepo` scopes the pass to the requested paths, so with **exactly one** requested path a non-zero `StaleFileCount` *proves* that file was re-parsed.
- With several paths the count cannot be attributed per path → resolves nothing.
- A whole-repo pass resolves nothing either: it admits by mtime, so it makes no promise about any particular file.
- A relative path with no known repo root is not guessed at.

Clearing a receipt for a file that was **not** re-parsed would be a false green — the very thing the receipt exists to prevent — so every ambiguous case declines.

Pending receipts are untouched: they describe work still in flight, not a dead generation.

The decision lives in a pure helper (`reindexedReceiptPath`) rather than inline, so it can be tested without a live indexer.

## Tests

- a table over the decision: single relative path joined to the repo root, single absolute path used as-is, several paths, whole-repo, nothing re-parsed, the path itself in `FailedFiles`, a relative path with no known root, a missing result;
- a unit test that the resolve touches **only** the re-parsed path and leaves an in-flight receipt's `result` untouched;
- an **end-to-end** test through `handleReindexRepository` that proves the wiring, using the existing harness in `tools_core_reindex_test.go`.

**Seven mutants** each fail their test and pass restored: resolver body removed · path filter dropped · `completed` guard dropped · handler stops calling the resolver · single-path guard dropped · `FailedFiles` check dropped · empty-root guess allowed.

> Worth saying: the first pass had the decision inline in the handler, and the mutation run showed **three** mutants not binding — the handler logic had no test at all. Extracting the helper and adding the end-to-end test is what closed that; the design change came from the mutants, not from taste.

## Verification

`go build` and `go vet` clean. The full `./internal/mcp` suite shows 20 failures on this Windows workstation — **the same tests fail identically on a clean tree** (8 of them sampled A/B), so they are pre-existing environment failures, not regressions. None is in the mutation or reindex families.

## Not covered

The other asks in #692 — retry-with-backoff for a failed ingest, and degrade-instead-of-refuse in `detect` — are untouched here, and #692 should stay open for them. This PR only makes the recovery that callers already attempt actually work.

### DIFF
--- a/internal/mcp/edit_serialization.go
+++ b/internal/mcp/edit_serialization.go
@@ -220,6 +220,37 @@ func (s *Server) resolveSupersededFailedReceipts(succeededPath string, succeeded
 	})
 }
 
+// resolveReindexedPathReceipts resolves terminally failed freshness receipts
+// for a path that an explicit reindex has just re-parsed. A failed ingest
+// fail-closes change.detect for the whole repository until retention lapses,
+// and reindex_repository was the obvious recovery that did not work: it
+// re-reads the file into the graph but never touched this map, so the barrier
+// kept reporting a gap the graph no longer had.
+//
+// The evidence a failed receipt is waiting for is exactly "the graph has read
+// the current bytes of this path", which a successful scoped re-parse
+// provides. Pending receipts are left alone: they describe work still in
+// flight, not a dead generation, and clearing one would hide a real gap.
+func (s *Server) resolveReindexedPathReceipts(reindexedPath string) {
+	cleanPath := filepath.Clean(reindexedPath)
+	s.mutationReceipts.Range(func(_, value any) bool {
+		receipt, ok := value.(*mutationReceipt)
+		if !ok || filepath.Clean(receipt.path) != cleanPath {
+			return true
+		}
+		receipt.mu.Lock()
+		if receipt.completed && (receipt.result.Err != nil || !receipt.result.Reindexed) {
+			receipt.result = indexer.MutationResult{
+				RequestedGeneration: receipt.generation,
+				AppliedGeneration:   receipt.generation,
+				Reindexed:           true,
+			}
+		}
+		receipt.mu.Unlock()
+		return true
+	})
+}
+
 func (r *mutationReceipt) outcome(pending bool) mutationReindexOutcome {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
--- a/internal/mcp/edit_serialization.go
+++ b/internal/mcp/edit_serialization.go
@@ -231,11 +231,44 @@ func (s *Server) resolveSupersededFailedReceipts(succeededPath string, succeeded
 // the current bytes of this path", which a successful scoped re-parse
 // provides. Pending receipts are left alone: they describe work still in
 // flight, not a dead generation, and clearing one would hide a real gap.
-func (s *Server) resolveReindexedPathReceipts(reindexedPath string) {
+// failedReceiptsBefore snapshots the failed freshness receipts a reindex pass
+// is entitled to resolve. It MUST be called before the pass starts: a receipt
+// that fails its own ingest while the pass is running was never read by it, so
+// resolving it would certify bytes the graph never saw. Receipt IDs are the
+// bound rather than a timestamp because receipts carry no completion time.
+//
+// A nil result resolves nothing, which is the correct reading of "no snapshot
+// was taken".
+func (s *Server) failedReceiptsBefore(paths []string, root string) map[string]struct{} {
+	candidate, ok := reindexCandidatePath(paths, root)
+	if !ok {
+		return nil
+	}
+	eligible := make(map[string]struct{})
+	s.mutationReceipts.Range(func(_, value any) bool {
+		receipt, ok := value.(*mutationReceipt)
+		if !ok || filepath.Clean(receipt.path) != candidate {
+			return true
+		}
+		receipt.mu.RLock()
+		failed := receipt.completed && (receipt.result.Err != nil || !receipt.result.Reindexed)
+		receipt.mu.RUnlock()
+		if failed {
+			eligible[receipt.id] = struct{}{}
+		}
+		return true
+	})
+	return eligible
+}
+
+func (s *Server) resolveReindexedPathReceipts(reindexedPath string, eligible map[string]struct{}) {
 	cleanPath := filepath.Clean(reindexedPath)
 	s.mutationReceipts.Range(func(_, value any) bool {
 		receipt, ok := value.(*mutationReceipt)
 		if !ok || filepath.Clean(receipt.path) != cleanPath {
+			return true
+		}
+		if _, wasFailingBeforeThePass := eligible[receipt.id]; !wasFailingBeforeThePass {
 			return true
 		}
 		receipt.mu.Lock()
--- a/internal/mcp/mutation_freshness_test.go
+++ b/internal/mcp/mutation_freshness_test.go
@@ -233,9 +233,13 @@ func TestReindexedPathResolvesFailedReceiptsForThatPathOnly(t *testing.T) {
 		RequestedGeneration: 7,
 		Err:                 errors.New("unrelated failure"),
 	})
+	// Snapshot taken where production takes it: before the pass runs. The
+	// in-flight receipt below therefore appears mid-pass and is out of scope
+	// by construction, not by the completed/pending check alone.
+	eligible := s.failedReceiptsBefore([]string{"file.go"}, "/repo-a")
 	inflight := pendingFreshnessReceipt(s, "receipt-inflight", "repo-a", "/repo-a/file.go", 9)
 
-	s.resolveReindexedPathReceipts("/repo-a/file.go")
+	s.resolveReindexedPathReceipts("/repo-a/file.go", eligible)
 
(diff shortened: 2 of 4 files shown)


zzet (changes requested):
The design here is right and the narrowness is the good kind — I tried to break the core premise and couldn't. But the `FailedFiles` guard, which is the one thing standing between this and a false green, does not fire in the relative-path case the helper explicitly supports.

## P1 — a file that failed to index gets its receipt stamped fresh

`result.FailedFiles` carries **absolute** paths — pinned by `TestIncrementalReindex_FailedFileSurfacedAndRetried` (`internal/indexer/incremental_reindex_test.go:196`, `assert.Contains(t, res.FailedFiles, bad)` where `bad := filepath.Join(dir, "bad.go")`). But the guard compares against the caller's **raw** spelling, before the join to `reindexRoot`:

```go
if slices.Contains(result.FailedFiles, paths[0]) {   // paths[0] may be relative
    return "", false
}
resolved := paths[0]
if !filepath.IsAbs(resolved) { resolved = filepath.Join(root, resolved) }
```

Pass a relative path and the guard never matches. End-to-end through the real handler and the real indexer — an unreadable stale file (your own `chmod 0o000` recipe), requested as `"bad.go"`:

```
PROBE B: StaleFileCount=1
PROBE B: receipt after reindex -> Err=<nil> Reindexed=true
  ^ the file failed to index, yet the barrier now reports it fresh
```

This is not defence-in-depth: `StaleFileCount` counts discovered-stale files *including* the ones that failed, so `FailedFiles` is the only thing standing between a failed ingest and a green barrier.

The decision table's `"the path itself failed to index"` case uses `[]string{"a.go"}` with `FailedFiles: []string{"a.go"}` — relative on both sides, a combination the production path never produces. So the "FailedFiles check dropped" mutant does die, but against a spelling that cannot occur; the real flow is unguarded.

### Fix

Resolve first, then compare both spellings. Comparing both is deliberately fail-closed — refusing a resolution is always the safe direction:

```go
resolved := paths[0]
if !filepath.IsAbs(resolved) {
	if root == "" {
		return "", false
	}
	resolved = filepath.Join(root, resolved)
}
resolved = filepath.Clean(resolved)
// FailedFiles carries ABSOLUTE paths while paths[0] is whatever the caller
// passed. Compare both so a relative request cannot slip past the guard.
requested := filepath.Clean(paths[0])
for _, failed := range result.FailedFiles {
	if c := filepath.Clean(failed); c == resolved || c == requested {
		return "", false
	}
}
return resolved, true
```

I verified this: it closes the probe, and it keeps **all eight table cases plus every existing test green** — the full `./internal/mcp` suite passes (79s, macOS). Worth adding a table row with an absolute `FailedFiles` entry against a relative request, so the guard is pinned against the spelling production actually produces.

## Secondary — the resolve is not bounded to receipts that completed before the pass

`resolveReindexedPathReceipts` stamps any completed-failed receipt for the path at the moment it runs. The indexer read the file's bytes at T0; the resolve runs at T1. A mutation that writes the file at some T between T0 and T1, fails its ingest, and completes its receipt before T1 gets stamped fresh — certifying bytes the graph never read.

That needs a concurrent edit during the pass, so it is narrower than P1, but it is the same unsafe direction. `mutationReceipt` has no completion timestamp, so the cheapest fix is to snapshot the candidate receipts (or a cutoff) *before* calling the indexer and only resolve those. Using pass-start as the cutoff is conservative — it declines some safe resolutions — which is the right way to be wrong here.

## Two notes, neither blocking

- Path matching is case-sensitive. On macOS/Windows a spelling difference between the receipt path and the request silently skips a resolution. Fail-closed, so fine — but worth a comment saying so.
- `paths` accepts directories as well as files (per `handleReindexRepository`'s own doc). `["internal"]` with a non-zero count returns `("/repo/internal", true)`. That is harmless only because receipts hold file paths and matching is exact equality; if receipt matching ever became prefix-based it would become a hole. A line in `reindexedReceiptPath`'s doc noting the arg may be a directory would keep that from being rediscovered the hard way.

## What's right

I specifically tried to break the attribution premise and could not. Scoping the pass to an unchanged file while a *different* file was stale returns `StaleFileCount=0` — so "exactly one requested path plus a non-zero count proves that file was re-parsed" holds, and the refusals for multi-path and whole-repo passes are the correct call rather than excessive caution.

Pending receipts are genuinely untouched, and the test asserts the stronger property (`result` never stamped) rather than just the outcome. Resolve-in-place is consistent with what landed in #693. And letting the mutation run drive the helper extraction — rather than taste — is exactly the right instinct; the note about the first pass having three non-binding mutants is the most useful paragraph in the description.

Fix the `FailedFiles` spelling and this is good to go.

tiendungdev:
Confirmed both, fixed both in `ff374e90`.

## P1 — the guard never fired for a relative request

I reproduced it before changing anything, and the diagnosis is exactly right: `FailedFiles` is absolute, `paths[0]` is whatever the caller passed, and the comparison ran before the join. Applied your fix as written — resolve first, then compare **both** spellings — and pinned it with the two rows that were missing:

```go
{"an absolute failed file is refused for a relative request",
    []string{"internal/a.go"}, "/repo",
    &indexer.IndexResult{StaleFileCount: 1, FailedFiles: []string{filepath.Join("/repo", "internal", "a.go")}},
    "", false},
{"an absolute failed file is refused for an absolute request",
    []string{absFixture}, "/repo",
    &indexer.IndexResult{StaleFileCount: 1, FailedFiles: []string{absFixture}},
    "", false},
```

The observation that made the review worth more than the fix: my `"the path itself failed to index"` row was relative on both sides, so the mutant died against a spelling production never emits. A mutant dying is evidence the test binds *something* — it says nothing about whether that something is the real flow. Two mutants now separate those: "compare only the requested spelling" kills the new relative-request row, "compare only the resolved spelling" kills the old one. Neither alone covers both.

## Secondary — resolve bounded to pre-pass receipts

Took the snapshot approach. `failedReceiptsBefore(paths, root)` collects the candidate receipt IDs **before** the indexer call, and `resolveReindexedPathReceipts` touches nothing outside that set:

```go
if _, wasFailingBeforeThePass := eligible[receipt.id]; !wasFailingBeforeThePass {
    return true
}
```

Both call sites in `handleReindexRepository` now learn `reindexRoot` before the pass rather than after, so the snapshot can be taken with the same resolved path the resolution will use. `reindexCandidatePath` is split out for exactly that reason — if the two ever resolved the spelling differently, the snapshot would bound the wrong receipts, which is a failure mode that would be invisible.

A nil snapshot resolves nothing. `TestFailedReceiptsBeforeRefusesToSnapshotAnUnattributablePass` asserts that every shape `reindexedReceiptPath` refuses also yields a nil snapshot, so the two guards cannot drift into a state where only one of them is standing between a failed ingest and a green barrier.

`TestReindexResolveIgnoresReceiptsThatFailedDuringThePass` covers the window directly: a write that lands, fails its ingest and completes while the pass is running stays failed, and the barrier keeps reporting it.

## Both notes

Recorded on `reindexedReceiptPath`'s doc — case-sensitive matching is fail-closed, and `paths` may name a directory, which is harmless *only* because receipt matching is exact equality, with a line saying it would become a hole if matching ever went prefix-based.

## Verification

Six mutants, each failing its test and passing restored: compare only the requested spelling · compare only the resolved spelling · drop the `FailedFiles` guard · accept a multi-path pass in `reindexCandidatePath` · ignore the pre-pass snapshot · return an empty map instead of nil.

`go build ./...` and `go vet ./internal/mcp` clean. The freshness/reindex tests pass.

One thing I cannot match from here: you ran the full `./internal/mcp` suite green in 79s on macOS; on Windows it is red before my branch. I checked out the three files at `HEAD~` and re-ran the failures — `TestPathToFileURI_Absolute`, `TestDiagnosticsBroadcaster_PayloadShape`, `TestCitedFilesFromResult_WireFormats/absolute_paths_are_not_citations` and `TestSymlinkGuardRefusesEscape` all fail identically on the pristine tree (drive-letter path shapes: `file:///D:/work/main.go` vs the expected `file:///work/main.go`). The full-package run also times out in `TestGatherExploreContentCandidatesForBareTermsCapsSearchesAndHydratesOnce`, which does not hang when run outside the full parallel load. None of that touches this change, but it does mean my full-suite evidence is Windows-shaped and worth nothing against yours — happy to file the platform failures separately if they are news.

zzet (approved):

Issue #511, opened 2026-08-08
design: make an empty analyze result say whether the data was ever built (contract_tier_unbuilt as the template)


tiendungdev:
## Summary

`analyze` already has a pattern that makes an empty result safe to act on — the `contract_tier_unbuilt` diagnostic. It is not applied to the other kinds whose backing data can legitimately be absent, so those return a bare zero that is indistinguishable from "computed, genuinely empty".

This is a design request, not a bug report. Nothing below is broken in the sense of returning a wrong number; the numbers are right. The problem is that one class of zero cannot be told apart from another by the caller.

## The pattern you already invented

Asking for contracts on a repo whose contract tier was never built returns zero **and explains itself**:

```
total: 0
contract_tier_unbuilt: <repo> — no contract-tier completion marker is recorded for
these repositories, so this empty result is not evidence that they declare no
routes or contracts ... rebuild the tier with a whole-repo re-parse —
untrack_repository then track_repository, or restart the daemon with
GORTEX_WARMUP_FULL_RETRACK=1. reindex_repository will NOT rebuild it: it admits
files by mtime, and an unbuilt tier leaves every mtime unchanged.
```

Three things make this good, and I want to be specific because they are worth preserving verbatim as the template:

1. it says the zero **is not absence evidence**;
2. it names **which** repository is in that state, not just that some are;
3. it names the recovery **and** rules out the recovery that looks obvious but does not work.

I acted on that message and it was correct end to end: the retrack rebuilt the tier, and the same query then returned real data. That is a diagnostic doing its whole job.

## Where the same class of zero is silent

Measured on `v0.63.1+4f2d169`, one daemon, one session.

**`analyze co_change`.** For a file whose top co-change partner appears in **40 of 40 commits (100%)** in `git log`, the answer is:

```
co_changing[0]:
total: 0
```

Nothing about whether co-change edges were ever mined. I then ran `gortex enrich cochange`, which **succeeded** — `{"enriched": 20, "duration_ms": 726}` — and the same query still returns 0, as does the query from the other end of the pair, and one for a file I had edited that same day. I could not tell from any of it whether the enrichment covers this file, covers a window that excludes it, or wrote edges the read path does not find. (I did not report that separately: `analyze edge_audit` hit the 59 s deadline before I could enumerate the 20 edges, so I have a symptom without a cause and did not want to file one as the other.)

[...]

PR #717, opened 2026-08-31
feat(mcp): say which graph_status values can still change

tiendungdev:
One slice of **#511**, taken from a case that cost real wall-clock time this week.

## The problem

`change.receipt` renders four `graph_status` values as if they were stages in a progression. **Only one of them is.**

```go
func (r *mutationCommitRecord) pendingReindexReceipt() string {
	if r.graph != mutationGraphPending { return "" }
	return r.reindexReceipt
}
```

`mutationStatusPayload` refreshes a record through that accessor, so once a record leaves `pending`, **`fresh`, `stale` and `failed` are frozen for the life of the entry**. Nothing in the payload says so.

An agent on my project sat on a `stale` receipt for several rounds waiting for it to turn `fresh`. It was doing everything the docs ask — it stopped mutating the path, refused to retry, refused to reindex or restart. It was waiting on a value that **could not change**, and had no way to learn that.

The committed-write guidance made it worse by asserting the opposite outright:

> "the bytes are on disk — do not re-apply this edit; **if graph_status is not `"fresh"` the graph is still catching up**"

True for `pending`. False for the two terminal failure values — which are exactly the ones a caller is most likely to sit on.

## The change

- **`graph_status_terminal`** — `recorded && graph != pending`, where `recorded` says whether an outcome was ever written to the record.

  Both halves are load-bearing. `beginMutationCommit` seeds `graph` as `stale` and nothing reports back until the mutation completes, so **`stale` is overloaded**: it means both "the reindex reported without confirming an index write" (terminal) and "nothing has reported yet" (in flight). Deriving terminality from the status string alone told the caller this tool exists to serve — one whose edit was abandoned at its deadline and may still be completing — to stop waiting on a value about to go `fresh`. That is the same failure this PR fixes, pointing the other way. `graphRecorded` is set by `recordGraph`, and by `markNotApplied` / `markFailed`, which are genuinely terminal because nothing was written and no reindex is coming.

- **`graph_note`** — what the value means and what would change it. The `stale` and `failed` notes carry the two facts that cost the most time:
  1. the value will not resolve on its own;
  2. **`graph_status` does not gate `change.detect` at all** — that barrier reads the freshness receipts (`mutationReceipts`), not this ledger (`mutationCommits`), so a caller watching `graph_status` is watching the wrong structure entirely.

- the committed guidance now points at the flag instead of promising progress;
- the no-argument listing path renders through the same flag. It previously published `graph_status` with no terminality — the exact rendering this PR argues against, reachable by a different entry point;
- the tool description gains the same one-line rule, since that is what an agent reads *before* deciding whether to poll.

Nothing about the four values themselves changes; this is purely making the payload say what it already knows.

## Tests

A table over the four values plus a record that never had `recordGraph` called, asserting flag + note; a test for `markNotApplied` / `markFailed`, where the status string alone would give the wrong answer in the safe-looking direction; a listing test where two entries render the same status string and only the flag separates them; a guard that the committed guidance never again claims a non-fresh graph is still catching up.

`TestMutationCommitListingKeepsEverySnapshotField` derives its expectation from `mutationCommitSnapshot` itself rather than a hand-written key list: whatever the struct marshals, the listing must publish, plus `graph_status_terminal` and nothing else. Hand-rolling that rendering is what silently dropped `new_sha` and `bytes_written` in review round 2.

Mutants across the three rounds each fail their test and pass restored: flag always true · flag always false · terminality ignoring the recorded flag · `recordGraph` / `markNotApplied` / `markFailed` not setting it · the snapshot dropping it · `stale` and `failed` notes weakened · the unrecorded note dropped · guidance reverted to the old sentence · flag not published at all · the listing hardcoding terminality · `new_sha` and `bytes_written` dropped or carrying wrong content · `tool` and `path` losing `omitempty`.

## Verification

`go build ./...` and `go vet ./internal/mcp` clean. My full-suite numbers are Windows-shaped and worth nothing against a macOS run: several `./internal/mcp` tests fail on this platform on a pristine tree (drive-letter path shapes), so the affected tests are what I verify rather than the suite.


zzet (changes requested):
The insight behind this is right, and I verified the load-bearing claim rather than taking it on trust: `awaitMutationFreshnessForRepos` ranges `s.mutationReceipts`, and `mutationCommits` never appears anywhere in `edit_serialization.go`. So "a caller watching `graph_status` is watching the wrong structure entirely" is true, and it is the most valuable sentence in this PR. The freeze is real too — I confirmed a record observed as `failed` stays `failed` even after #693's supersede clears the underlying receipt.

But the flag is wrong in the one window this tool exists to serve.

## P1 — an edit still in flight is published as `graph_status_terminal: true`

`beginMutationCommit` seeds the record with `graph: mutationGraphStale`, and `recordGraph` runs only when the mutation reports back (six production call sites, all at completion). Between those two points a record renders:

```
graph_status          = stale
graph_status_terminal = true
graph_note            = "the reindex completed without confirming an index write, or no
                         outcome was recorded. This will NOT become "fresh" on its own...
                         The bytes are on disk: verify them and proceed"
```

and then the reindex reports and it becomes `fresh` — a value published as `terminal: true` changed.

That window is this tool's own documented primary use case. From the description you just edited: *"Use it when an edit_file / write_file / edit_symbol call was abandoned at its deadline (\"the work may still complete in the background\")"*. That is exactly a committed record whose reindex has not reported yet — and the new flag tells that caller the value is settled and to stop waiting.

It is the same failure mode you are fixing, in the opposite direction, aimed at the same caller. The old sentence made an agent wait forever on a terminal value; this makes it give up on a value that was about to go fresh.

The root cause is that `stale` is overloaded. It means both "the reindex completed without confirming an index write" (terminal) and "no outcome has been recorded yet" (in flight, not terminal). Your note names both readings — "or no outcome was recorded" — and then asserts terminality for both. Terminality cannot be derived from the status string alone, because the string does not carry that distinction.

The decision table cannot catch this: every case calls `record.recordGraph(tc.outcome)` before reading the payload, so none exercises a record in its initial state. That is also why the "flag always true" mutant dies — it dies against four recorded records, never against an unrecorded one.

### Fix

Carry whether an outcome was ever recorded, and key terminality off that:

- add a `graphRecorded bool` to `mutationCommitRecord`, set it in `recordGraph`;
- set it in `markNotApplied` and `markFailed` too — those are genuinely terminal, since nothing was written and no reindex is coming;
- surface it on the snapshot, and have `graphStatusTerminal` return `false` when unrecorded, with a note along the lines of "the reindex for this mutation has not reported back yet — poll this receipt; this value is not settled".

I verified this: the in-flight probe flips to `terminal: false`, all four of your table cases stay green unchanged, and the full `./internal/mcp` suite passes (82s, macOS). Worth adding a fifth table row for a record that has never had `recordGraph` called.

## Non-blocking

The no-argument listing path returns raw `mutationCommitSnapshot` values, so it still renders `graph_status` with no `graph_status_terminal` and no `graph_note` — the exact rendering this PR argues is misleading. An agent that lists recent mutations instead of selecting one gets the old, flag-free view.

Also worth a sentence somewhere: the frozen value is observer-dependent. If someone calls `change.receipt` between the ingest failure and a later supersede, the ledger says `failed` forever; if nobody looks in that window, the same history renders `fresh`. Not introduced here, and your `failed` note already handles the consequence well by pointing at the receipt rather than the ledger — but the asymmetry is surprising enough to name.

Fix the in-flight case and this is a clear improvement on what is there now.

tiendungdev:
Confirmed, fixed in `bbe09f67`.

I checked the P1 before touching anything rather than taking the trace on trust: `beginMutationCommit` seeds `graph: mutationGraphStale`, `recordGraph` is only reachable at completion, and `mutationStatusPayload` reads the field either way. So a committed record whose reindex has not reported renders `terminal: true` and then goes `fresh`. You are right that this is the worse direction — the old sentence made a caller wait forever, this told the caller most likely to be mid-flight to stop waiting on a value that was about to settle.

The framing that made it obvious: **`stale` is overloaded**, and my own note said both readings out loud — "the reindex completed without confirming an index write, **or no outcome was recorded**" — while asserting terminality for both. I wrote the disproof into the note and did not read it.

## The fix

Applied as suggested. `graphRecorded` on `mutationCommitRecord`, set in `recordGraph` and also in `markNotApplied` / `markFailed` — those are terminal because nothing was written and no reindex is coming, which is precisely where the status string alone gives the wrong answer in the safe-looking direction. Carried on the snapshot; `graphStatusTerminal(graph, recorded)` is `recorded && graph != pending`; unrecorded records get a note saying the value is a seed rather than an outcome and is not settled.

## On the table not catching it

This is the part worth keeping. Every case called `record.recordGraph(tc.outcome)` before reading the payload, so no case ever saw a record in the state `beginMutationCommit` leaves it in — and the "flag always true" mutant died against four recorded records while the unrecorded one, the only one where the flag was wrong, was unreachable from the table.

A mutant dying tells you a test binds *something*. It does not tell you the table's rows span the states production produces. Both PRs failed in exactly that shape, from opposite ends: here the table never built the initial state; in #716 it built a spelling the production path never emits. Neither gap is visible from the mutation results — only from asking which states a row can actually be in.

So the fifth row is `unrecorded: true`, skipping `recordGraph`, plus `TestMutationStatusTerminalForWritesThatNeverHappened` for the two never-written cases. Eight mutants now, each failing its test and passing restored: terminality ignoring the recorded flag · ignoring the status string · `recordGraph`, `markNotApplied`, `markFailed` each not setting it · the snapshot dropping it · the unrecorded note dropped · the listing hardcoding terminality.

## Both non-blocking notes

**The listing** now renders through the same flag (`mutationCommitListingPayload`). It was the one rendering this PR argues against, reachable by a different entry point — an agent that lists instead of selecting got the old view. The listing deliberately does *not* refresh pending records the way the single-record path does: a survey should not mutate what it surveys, and a caller acting on one entry selects it by receipt, which takes the refreshing path. `TestMutationCommitListingCarriesGraphTerminality` pins it with two entries rendering the same `graph_status` string where only the flag separates them.

**Observer-dependence** is recorded on that function's doc: `mutationStatusPayload` only refreshes while the record is pending, so whether the ledger reads `failed` forever or renders `fresh` for the same history depends on whether anyone called `change.receipt` in the window before a later supersede. As you say, the `failed` note already handles the consequence by pointing at the freshness receipt — the structure that actually gates `change.detect` — but the asymmetry is surprising enough that it should not have to be rediscovered.

## Verification

`go build ./...` and `go vet ./internal/mcp` clean; the mutation-status tests pass. Same caveat as on #716: my full-suite number is Windows-shaped and worth nothing against your 82s macOS run — several `./internal/mcp` tests fail on this platform on a pristine tree (drive-letter path shapes), so I verified the affected tests rather than claiming the suite.

zzet (changes requested):
The P1 is closed, and I checked it the way I'd want it checked rather than reading the diff and agreeing. One new regression, introduced by the fix for my own non-blocking note, and two wording items.

## P1 — confirmed fixed

The record in exactly the state `beginMutationCommit` leaves it in:

```
r1: graph_status=stale  terminal=true   "This will NOT become fresh on its own"
r2: graph_status=stale  terminal=false  "a seed rather than an outcome — poll this receipt; it is not settled"
```

I also checked the direction the fix could plausibly break instead: a terminal path that forgets to set `graphRecorded` would render `terminal:false` forever and reproduce the original wait-forever bug at a different address. There are exactly four writers of `r.graph` — `markNotApplied:183`, `markFailed:194`, `recordGraph:220`, and the seed at 457. All three non-seed writers set the flag, and `markCommitted` correctly does not, since the bytes are on disk but the reindex is still owed. No gap.

Five mutants, all bind: terminality ignoring `recorded` · `recordGraph` not setting it · `markNotApplied` not setting it · `markFailed` not setting it · the snapshot dropping it. The one that matters is the first — it now kills the `unrecorded` row, so the round-1 blindness is actually gone rather than relocated. M3 and M4 each kill only their own subtest.

Your point about mutation results is the right lesson and worth keeping: a dying mutant proves a test binds something, not that the rows span the states production can produce. Both PRs failed in that shape from opposite ends.

`./internal/mcp` is 4986 pass / 0 fail here on macOS, CI 11/11.

## Blocking — the listing drops two fields it used to publish

`mutationCommitListingPayload` hand-builds its maps instead of marshalling the snapshot, so `new_sha` and `bytes_written` are gone. The old `mutationCommitListing` returned `[]mutationCommitSnapshot`, and both fields marshal with `omitempty`:

```
OLD keys = [bytes_written committed_at disk_status graph_status mutation_id new_sha path receipt started_at tool]
NEW keys = [committed_at disk_status graph_status graph_status_terminal mutation_id path receipt started_at tool]
```

`new_sha` is how a caller tells which content actually landed. The severity is capped, because a caller acting on an entry selects it by receipt and takes the single-record path that still carries both — but it is a silent loss from a published payload, and nothing in `TestMutationCommitListingCarriesGraphTerminality` would catch it, since that test only asserts the flag and the status string.

Same shape as the two conditionals already there:

```go
if snap.NewSHA != "" {
	entry["new_sha"] = snap.NewSHA
}
if snap.BytesWritten != 0 {
	entry["bytes_written"] = snap.BytesWritten
}
```

Worth a row in the listing test pinning that a committed entry still carries `new_sha`, since this is the second time a hand-rolled rendering has quietly diverged from the structure it replaced.

Minor, same rewrite: `tool` and `path` are now set unconditionally, so an empty one renders as `""` where `omitempty` dropped it.

## Two wording items

**The `stale` note still carries its own disproof.** That branch is only reachable when `recorded == true`, so "or no outcome was recorded" is now provably impossible:

```go
case mutationGraphStale:
	return "the reindex completed without confirming an index write, or no outcome was recorded. ..."
```

That clause is the exact overloading this PR exists to remove, and your comment named this sentence as the thing you'd written and not read. It survived the fix. The note is the product here — it should read as the one thing it can now mean.

**The PR body still states the round-1 rule** — "`graph_status_terminal` — `true` for everything except `pending`" — rather than the shipped `recorded && graph != pending`. Worth correcting so the squash message doesn't ship the wrong rule.

Everything else — the flag, the notes, the guidance rewrite, the listing going through the same flag, observer-dependence recorded on the doc — is right, and the barrier insight is still the most valuable thing in this change.

tiendungdev:
All three fixed in `15de7c5f`, and the PR body now states the shipped rule.

## The listing regression — confirmed, and it was mine

I checked the key sets before changing anything rather than reading your diff and agreeing. `mutationCommitListingPayload` published nine keys where the marshalled snapshot published ten plus `bytes_written`; `new_sha` and `bytes_written` were both gone, and `tool` / `path` were unconditional where `omitempty` had dropped them.

Worth naming what happened: I introduced this **while fixing your non-blocking note**. The note was about the listing publishing too little information; my fix made it publish less. Replacing a marshalled struct with a hand-built map is a rewrite that looks like an addition, and I did not diff the output of the thing I replaced against the thing I replaced it with.

## The fix, and why the test is shaped this way

The two conditionals are back, plus `tool` and `path` restored to `omitempty` parity. But writing the expected key list by hand in a test would repeat the same mistake one layer up — I would be writing down the keys I *believe* the snapshot publishes, which is exactly the belief that was wrong. So `TestMutationCommitListingKeepsEverySnapshotField` derives it from the struct:

```go
raw, _ := json.Marshal(tc.record.snapshot())
json.Unmarshal(raw, &marshalled)
want := keysOf(marshalled) + "graph_status_terminal"
require.ElementsMatch(t, want, keysOf(entries[0]))
```

Whatever `mutationCommitSnapshot` marshals, the listing must publish; plus the one field the listing exists to add, and nothing else. Two records cover both directions — every field populated, and only the always-present ones, which is where a lost `omitempty` shows up. `new_sha` and `bytes_written` are additionally asserted by value, since a key carrying wrong content satisfies key parity.

Six mutants, each failing and passing restored: `new_sha` dropped · `bytes_written` dropped · `new_sha` carrying `"unknown"` · `tool` losing `omitempty` · `path` losing `omitempty` · the `stale` note regaining its clause.

That test would also have caught a third divergence I had not looked for, which is the point of deriving it rather than listing it.

## The `stale` note

You are right, and this is the second time that sentence survived a round it should not have. "or no outcome was recorded" is now provably unreachable — the branch only runs when `recorded == true` — so it was a clause asserting the exact overloading the change removes. It now reads:

> "the reindex reported back without confirming an index write. This will NOT become `"fresh"` on its own. It also does not gate `change.detect` — that barrier reads the freshness receipts, not this ledger. The bytes are on disk: verify them and proceed"

The table row for `stale` now asserts on `"reported back without confirming"` rather than the generic `"will NOT become"`, so the mutant that restores the old clause dies.

## PR body

Rewritten to `recorded && graph != pending`, with the reason both halves are load-bearing, so the squash message does not ship the round-1 rule.

## On your check of the opposite direction

Checking that a terminal writer forgetting `graphRecorded` would reproduce wait-forever at a different address, and then enumerating all four writers of `r.graph` to prove no gap, is a stronger closure than my own — I verified the fix worked, you verified the fix could not be half-applied. `markCommitted` correctly not setting it is the detail I would have been most likely to get wrong in the other direction.

zzet (approved):

PR #677, opened 2026-08-26
parser: C# scoping, attribution, and type-identity correctness (dispatch gate deferred)

pbednarcik:
## What this PR is now

Round 6 took the split recommended in rounds 3 through 5: the extraction-side correctness work lands here, and the type-argument dispatch gate (the closure proof and the fan-out filter) is deferred until type node identity is unique - a receiver that can carry a resolved field node ID, an implements edge that carries its own construction. `internal/resolver/csharp_iface_dispatch.go` is byte-identical to main again; the resolver side of this diff is exclusively recall pins.

What lands, each with its own commits and pins:

- **Block-scoped shadow index** - locals, patterns, and every other binding form carry byte extents; the `receiver_name` refusal asks a per-site question instead of a function-wide one. Includes the round-6 fixes: the switch BODY forms the scope, not the section (B1); per-record tier guards so sibling-scope redeclarations keep their evidence (B5); catch-filter and query-clause patterns scope to their clause, not the method.
- **Byte-interval call attribution** - two members sharing a physical line attribute by call offset, and a member kind without recorded extents falls back to line attribution instead of dropping the call (B3).
- **Identifier canonicalization, both sides** - base-list entries AND declarations reduce to one canonical spelling, so a verbatim-declared type (`@event`) keeps its hierarchy edges, its edge kind, and its member fan-out (B2). Node IDs are canonical now; the extractor version bump re-extracts stores carrying raw-spelled IDs.
- **Partial base-list merge with an identity proof** - same-file partial fragments merge their base lists only when namespace, enclosing-type chain, and generic arity all match; twins keep the old drop-the-bases behaviour, and fragments share one extends budget (B4).
- **`csharpBaseTypeName` final-segment fix**, the **variance stamp ordering**, the **shared `parser.CSharpBCLKeywordFolds` table**, and the **extractor version bump**.
- **Type-argument stamps** (`target_type_args`, `field_type_args`) stay as additive evidence for the deferred gate; the package doc in `csharp_base_type_args.go` records the deferral in one place.

The gate test families went with the gate, including the dispatch benchmark that exceeded the CI bench-job timeout.

PR #720, opened 2026-08-31
parser: C# accessor bodies, initializers, and the implicit value parameter own their calls

pbednarcik:
## What this fixes

A call inside a property accessor body was dropped outright at extraction. The call-owner lookup admits only methods and constructors, properties are field-kind nodes, and a call whose enclosing member the lookup cannot name is discarded, not misattributed. In a probe class with four authored call sites (get body, set body, expression-bodied property, expression-bodied method), exactly one survived: the expression-bodied method, the only shape that rides emitMethod. Accessors are everywhere in production C#, so this was the largest recall gap my counted fixture repo still showed. Field initializer calls and MediatR dispatch sites inside accessors died at the same gate, and instantiates edges from accessor bodies landed on the file node.

This is the fuller fix the round-6 byte-attribution work in #677 left a comment pointing at: record extents for the remaining member kinds instead of letting their calls fall through.

NOTE ON THE BASE: this branch is stacked on the #677 head (the extents machinery is its foundation). #677 has since merged - say the word if you want this rebased onto main.

## What lands

- **Properties own their calls.** emitProperty records the declaration byte span the way emitMethod and emitConstructor do, and csharpOwnerRanges widens the owner set with every extent-carrying member node. Block get/set/init bodies, expression-bodied properties AND accessors, static accessors, lambdas and local functions inside accessors, property initializers, explicit interface implementations, and the C# 13 field-keyword shape all attribute to the property node. A side effect worth naming: properties on a shared line now byte-attribute precisely instead of riding the round-6 line fallback.
- **C# 13 partial properties.** The declaration/implementation split meant the dedup kept whichever fragment extracted first as the extent owner - for the usual declaring-part-first layout that was the bodiless fragment, so every call in the implementation still died. Ownership now re-records from the body-bearing fragment, with a parallel line-span map because the minted node keeps the first fragment's lines while the code lives in the second. This is the shape source generators emit.
- **Field initializers own their calls, per DECLARATOR.** `int a = F(), b = G();` hands each call to the field it actually initializes. Initializer-less declarators stay out of the owner set (no code, and a 1-line range would tie lines needlessly - pinned via the ambiguity-stamp observable). const declarators are skipped entirely.
- **The implicit `value` parameter is typed by the property declaration** - as an offset-scoped record over each set/init accessor span, never owner-wide. Inside set/init, `value` is always the parameter (a same-named member needs this.value), so those sites get real receiver evidence; in a getter the bare name means a member, possibly inherited or declared in another partial fragment, so getter sites are left to ordinary member evidence rather than stamped with a confident wrong answer. The scope registration also keeps the field-identifier lane from reading the parameter as a field.
- **Lane consistency.** The reference-form pass (instantiates, casts, static access, attributes) and the MediatR dispatch pass consume the same widened owner set, so one line of accessor code no longer splits its evidence between two owners.
- **scope_ns on property and initialized-field nodes**, for the scoped-usings narrowing, matching every other owner kind.
- **Extractor version bump (csharp 17 to 18)** so existing stores re-extract and mint the recovered edges.

## Evidence

Every lane RED-first with committed pins (accessor matrix, per-declarator field inits, partial-property fragments, six value-parameter shapes including the inherited-member and typed-getter-local traps, owner-asserting same-line pins). On my counted fixture repo, cold store A/B against the base commit: the four-site probe class goes 1/4 to 4/4, and an 18-cell deep matrix (every accessor shape above, each with a uniquely named target asserted at exactly its authored count against the resolved callee) goes all-zeros to all-exact. Indexer and event add/remove accessor bodies stay at zero on both stores: those members are not emitted as nodes at all today, so their cells stay armed for a separate member-emission change rather than a bolt-on here. Full parser/resolver/indexer suites: fail-name sets byte-identical to clean-base controls on Windows; race pass green.

## Behavior notes, disclosed

- **More receiver_ambiguous stamps on one-line layouts.** An initialized field or a property sharing a physical line with another member now line-ties in ambiguousAt, so calls owned by extent-carrying members there carry the refusal stamp even though byte attribution names their owner precisely. Kept deliberately: the stamp exists for line-keyed consumers, and un-stamping would cut against the equal-span refusal from #677 round 4. If you would rather ambiguousAt consult offsets, that is a small follow-up.
- **Extent-less members sharing a line are refused, not misattributed** (review round). An indexer or event accessor records no bytes of its own, so its call reached the line fallback - and the widened owner set could make that fallback answer a property or initialized field that provably does NOT contain the call, minting a false edge with full receiver evidence and no ambiguity stamp where the base emitted nothing. The fallback now refuses when the line answer's recorded bytes exclude the offset (guard authored and corpus-verified in review). This also removes the pre-existing method-neighbour false edge and its read. The extent-less member's recall returns when indexers and event accessors are emitted with extents of their own - the member-emission follow-up already noted in Evidence.
- **The reference-form and MediatR lanes stay line-keyed internally** (pre-existing); they gained the widened owner set, not byte precision.
- **Downstream kind filters.** A few analysis rollups (architecture hierarchy leaf kinds, betweenness centrality's SQL pushdown, ref_facts in unscoped mode) filter callers to function/method kinds, so the recovered property-owned edges are correct in the graph but invisible to those surfaces. Left untouched here since widening them is a policy call - happy to file it separately.
- **Enrichment-tier interplay.** The line-keyed owner attribution in the enrichment tiers (the finding disclosed in the #677 round-6 thread) now has a new caller population to disagree with on shared lines. Nothing in this diff changes those tiers; it makes the case for that separate fix stronger.

### DIFF
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -40,7 +40,7 @@ var extractorVersions = map[string]int{
 	//   "go": 2,
 	"c":      generatedParserProjectionPolicyVersion, // generated parser projection covers all strictly detected table sizes
 	"php":    2,                                      // class/interface inheritance now emits typed structural edges
-	"csharp": 17,                                     // clause-specific binder extents, offset-aware typed locals + field uses, byte-interval caller attribution, canonical base-list spellings, partial-fragment base preservation (was: canonical-identifier type-arg domain)
+	"csharp": 18,                                     // accessor bodies, property/field initializers, and the typed value parameter own their calls (was: clause-specific binder extents, offset-aware typed locals + field uses, byte-interval caller attribution, canonical base-list spellings, partial-fragment base preservation)
 	"scala":  2,                                      // explicitly instantiated generic calls emit call edges
 	"go":     3,                                      // generic instantiations are marked so indexing a func value cannot bind (was: generic calls emit call edges)
 	"cpp":    2,                                      // templated and namespace-qualified calls emit call edges
--- a/internal/indexer/extractor_version_test.go
+++ b/internal/indexer/extractor_version_test.go
@@ -229,7 +229,7 @@ func TestStaleLangsDetection(t *testing.T) {
 		}
 
 		for _, path := range []string{"src/Handler.cs", "Views/Page.razor", "Views/Page.cshtml"} {
-			want := policySalt + "|csharp@17"
+			want := policySalt + "|csharp@18"
 			if got := merkleSaltFor(path); got != want {
 				t.Errorf("C# extractor salt for %s = %q, want %q", path, got, want)
 			}
--- a/internal/parser/languages/csharp.go
+++ b/internal/parser/languages/csharp.go
@@ -880,7 +880,7 @@ func (e *CSharpExtractor) extractCSharp(filePath string, src []byte) (*parser.Ex
 	// `x as Foo`), static / const access (`Foo.Empty`, `typeof(Foo)`,
 	// `nameof(Foo)`), and attribute type names (`[Foo]`). Inheritance is
 	// already covered by emitCSharpBaseList, so it is not re-emitted here.
-	emitCSharpReferenceForms(root, src, filePath, fileID, result)
+	emitCSharpReferenceForms(root, src, filePath, fileID, result, funcBytes, funcLines)
 
 	// Two same-named calls on ONE line whose receivers differ make the
 	// line unattributable: member companions dedupe to a single stored
@@ -1123,7 +1123,7 @@ func (e *CSharpExtractor) extractCSharp(filePath string, src []byte) (*parser.Ex
 	captureValueRefCandidates(result, root, filePath, src)
 	captureFnValueCandidates(result, root, filePath, src)
 
-	captureMediatRDispatch(result, root, filePath, src)
+	captureMediatRDispatch(result, root, filePath, src, funcBytes, funcLines)
 
 	return result, hadError, nil
 }
--- a/internal/parser/languages/csharp.go
+++ b/internal/parser/languages/csharp.go
@@ -520,7 +520,7 @@ func (e *CSharpExtractor) extractCSharp(filePath string, src []byte) (*parser.Ex
 			e.emitField(m, filePath, fileID, src, result, seen, fileAliases)
 
 		case m.Captures["prop.def"] != nil:
-			e.emitProperty(m, filePath, fileID, src, result, seen, fileAliases)
+			e.emitProperty(m, filePath, fileID, src, result, seen, fileAliases, funcBytes)
 
 		case m.Captures["using.def"] != nil:
 			e.emitUsing(m, filePath, fileID, result)
@@ -630,7 +630,7 @@ func (e *CSharpExtractor) extractCSharp(filePath string, src []byte) (*parser.Ex
 	// type environments. Owner attribution runs once per local, call and
 	// type use — the sorted lookup keeps that from multiplying into an
 	// O(locals×functions) linear-scan product on member-heavy files.
-	funcRanges := newCSharpFuncLookup(buildFuncRanges(result), funcBytes)
+	funcRanges := newCSharpFuncLookup(csharpOwnerRanges(result, funcBytes), funcBytes)
 
 	// Build type environments in legacy precedence, scoped per enclosing
 	// method — a same-named local of a different type in a sibling method
@@ -1901,7 +1901,7 @@ func (e *CSharpExtractor) emitField(m parser.QueryResult, filePath, fileID strin
 	emitCSharpTypeUseEdges(id, fieldTypeRaw, filePath, def.StartLine+1, result)
 }
 
-func (e *CSharpExtractor) emitProperty(m parser.QueryResult, filePath, fileID string, src []byte, result *parser.ExtractionResult, seen map[string]bool, fileAliases map[string]bool) {
+func (e *CSharpExtractor) emitProperty(m parser.QueryResult, filePath, fileID string, src []byte, result *parser.ExtractionResult, seen map[string]bool, fileAliases map[string]bool, funcBytes map[string][2]int) {
 	def := m.Captures["prop.def"]
 	owner := csharpDirectMemberOwner(def.Node, src, "class_declaration", "struct_declaration", "interface_declaration", "record_declaration")
 	if owner.kind == "" {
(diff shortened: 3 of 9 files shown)


zzet (changes requested):
This is strong work, and the measurement backing it holds up under independent checking. One finding needs resolving before it merges — the owner widening hands a *wrong* owner to the member kinds that still have no extents, in layouts where the base emitted nothing at all.

## What I verified independently

I harvested 432 C# fixtures from the **base's own** `*_test.go` inline strings, ran the extractor at `786a3929` and at `903b520b`, and diffed sorted node+edge dumps.

**19 lost / 24 gained, zero unexplained losses.** Every delta decomposes into:

- `instantiates` / `references` edges moving off the **file node** onto the owning member — the core recovery, exactly as described;
- three call/read re-attributions from a wrong same-line member to the right one — including `Q => _b.Take(); public void M(){}`, the false edge from the round-6 thread on #677. That one is genuinely fixed, and it was the thing I most wanted closed;
- two artifacts from non-C# fixtures my harvester wrongly admitted (a TypeScript and a Dart file). Mine, not yours.

Targeted probes confirm each claimed recovery, all previously dropped or file-attributed: a local function inside an accessor, a lambda in a field initializer, `value.Ship()` in a setter, property and method on separate lines. Extraction is deterministic across three runs. And the new tests are load-bearing — reverting `csharpOwnerRanges` back to `buildFuncRanges` turns **ten named tests** red, which is the standard I want on work this subtle.

## The finding

A member kind that still records no extents — indexer, event accessor — sharing a physical line with one of the **newly** extent-carrying owners now has its call attributed to that owner:

```csharp
public int this[int i] => _b.Idx(); public int Q => _b.Pro();
```

Head emits `Q01.Q -> *.Idx`. Base emitted **nothing**. Same shape with an event accessor:

```csharp
public event System.EventHandler E { add { _b.Ev(); } remove { } } public int Q => _b.Pro();
```

Head emits `Q03.Q -> *.Ev`.

Both carry full receiver evidence (`receiver_name:_b`) and no ambiguity stamp — `ambiguousAt` counts ranges, and an extent-less member contributes none, so there is no tie for it to detect. Downstream they are indistinguishable from the correct edge on the same line.

To be fair about what is new: when the same-line neighbour is a **method** (`this[int i] => _b.Take(); public void M(){}`) the false edge already exists on the base, and this PR does not introduce it. What is new is that **properties and initialized fields can now be the wrong owner**, in layouts where the base correctly emitted nothing. Worth noting that shape does not appear anywhere in the 432-fixture corpus — only the hand-written probes surfaced it, so the A/B you ran would not have shown it either.

One correction to the behavior notes: they say these calls "carry the refusal stamp even though byte attribution names the owner precisely." For this population byte attribution names the *wrong* member, and no stamp is applied. That is the sentence I would most want fixed regardless of which resolution you pick, because it is what a future reader will rely on.

This is also the half of the round-6 prescription that is still open: record extents for the remaining member kinds, **or keep the refusal**. Properties and field initializers got the first half. Indexers and event accessors kept the fallback.

## Options

**(a) Refuse — verified, six lines.** In `enclosingAt`, when the line fallback's answer has recorded bytes that exclude the offset, the offset is provably not inside that member:

```go
	if best != "" {
		return best
	}
	// The line fallback exists for member kinds that record NO extents
	// at all (indexer, event accessor). If the line answer does have
	// recorded bytes and the offset sits outside them, the offset is
	// provably not inside that member, so handing it the call invents a
	// false edge on a same-line neighbour - strictly worse than the drop
	// it replaced.
	fb := l.enclosing(line)
	if b, ok := l.bytes[fb]; ok && (offset < b[0] || offset >= b[1]) {
		return ""
	}
	return fb
```

I ran this: both false edges go away, every legitimate gain survives (corpus goes to 21 lost / 24 gained — the two extra losses are the pre-existing `BLIdx.M` false edge and its read). It needs the `indexer_shares_line` case of `TestResolveCSharp_ExtentlessMemberSharingALineKeepsItsCall` flipped, since that case asserts presence-only and knowingly tolerates the wrong owner.

**(b) Emit indexers and event accessors as nodes with extents** so they own their own calls. The real fix, and the one the round-6 note pointed at — but a bigger change, and I would not ask you to fold it in here.

**(c) Keep the tradeoff.** Defensible: a recall miss versus a precision miss is a real judgement call, and you made it deliberately and pinned it. If you go this way, then the behavior note and that test's doc comment both need to say plainly that the call lands on a *different, wrong* member — not merely that it "stays line-keyed."

My preference is (a) now with (b) tracked as a follow-up, because a false `calls` edge on an empty-bodied member is the kind of thing that quietly poisons `find_usages` and the impact rollups, and it is invisible to whoever reads the result. But I would take (c) with the wording fixed if you would rather not lose the indexer recall in this PR — say which and I will re-review promptly.

## Minor, non-blocking

- `emitField` adds `scope_ns` to **every** field, but the comment above it says "Initialized fields are call owners; they carry scope_ns" — the write is not gated on ownership. Harmless, arguably what you want, but the comment does not match the code and it grows every C# field node's meta (which is what the tstypes golden regen is showing).
- `csharpPropertyHasExecutableBody` treats any `block` or `arrow_expression_clause` as a body, so an auto-property with a block-lambda initializer (`public Func<int> F { get; } = () => { … };`) reads as body-bearing. Only reachable on the duplicate-declaration path, where valid C# makes it unreachable today — worth a comment noting why, rather than a change.

## Base

Still stacked on the #677 head. #677 is approved but unmerged, so this needs it to land first or a rebase — no action needed from you yet, just flagging that it gates the merge.

The accessor recall gap was worth chasing and the evidence you brought for it is the right kind. Close the extent-less owner case and this is ready.

pbednarcik:
Went with (a), plus the wording fixes - pushed as e38b54bb.

- Your guard, verbatim, in enclosingAt. Confirmed RED-first before applying: indexer-beside-property minted from Q01.Q, event-accessor-beside-property from Q03.Q, and the method-neighbour shape from BLIdx.M - all three now refuse, and each fixture also pins the property keeping its own call.
- The presence-only case is flipped, and the test is renamed TestResolveCSharp_SameLineMemberCallAttribution - "keeps its call" was the sentence the flip falsifies.
- Behavior notes corrected in the body: the stamp claim is now scoped to extent-carrying owners, and a new bullet documents the refusal, the removal of the pre-existing method-neighbour false edge, and that the recall returns with member emission - which is queued as my next arc (your option b), sequenced after this lands.
- emitField's comment now matches the ungated write; csharpPropertyHasExecutableBody documents why the block-lambda auto-property misread is unreachable rather than special-casing it.

On my counted C# fixture repo nothing moves: the armed indexer/event cells pin the loss at zero edges, and the guard can only turn a false edge into a refusal; no indexer-beside-property shape exists in the fixture tree, consistent with your corpus observation. Full parser/resolver/tstypes suites: fail-name sets identical to clean-base Windows controls.

Saw #677 merged - say the word if you want this rebased onto main rather than merged from the stack.

zzet:
> Saw https://github.com/zzet/gortex/pull/677 merged - say the word if you want this rebased onto main rather than merged from the stack.

Rebasing onto main sounds like a good idea (at least the change will be easy to see)

pbednarcik:
@zzet  - done

zzet (approved):

PR #734, opened 2026-09-01
csharp: scope label and when-guard pattern variables to their switch section

### LINKED ISSUE #724: csharp: case-label and when-guard pattern variables must scope to the switch section, not the switch body
Follow-up from the review of #677. `csharpScopeFormers` gained `switch_body` in place of `switch_section`. That is right for one kind of binder and wrong for the other, because **C# has two switch scoping rules that coexist**.

## Ground truth

Both of these compile clean under Roslyn (`dotnet` 9.0.305, `net9.0`):

```csharp
// (1) sibling sections may redeclare the same pattern variable
//     => a case-label / `when`-guard pattern variable is SECTION-scoped
public int SiblingSectionPatternVar(object o) {
    switch (o) {
        case int x: return x;
        case string x: return x.Length;
        default: return 0;
    }
}

// (2) a local declared in one section is alive in its siblings
//     => an ordinary local is BODY-scoped (redeclaring it is CS0128)
public int SiblingSectionLocal(int k) {
    switch (k) {
        case 1: int y = 1; return y;
        default: y = 2; return y;
    }
}
```

A third case pins the boundary: a pattern variable introduced by an `if` **inside** a section *does* escape to the switch body — Roslyn reports `CS0165: Use of unassigned local variable`, i.e. it resolves to the local and is merely not definitely assigned, rather than falling back to a field. So the discriminator is not the ancestor node type; it is **where in the section the binder sits**.

## Effect

Because everything now stops at `switch_body`, a case-label pattern variable shadows the entire switch body and eats a same-named field use in a sibling section.

```csharp
namespace App {
    public class Repo { public int Get(int id) { return id; } }
    public class Flow {
        private readonly Repo repo;
        public int Run(object o, int k) {
            switch (k) {
                case 1 when o is Repo repo: return repo.Get(1);
                default: return repo.Get(2);   // `repo` here is the FIELD
            }
        }
    }
}
```

`reads` edges to `unresolved::*.repo` from `Flow.Run`:

| revision | field reads | correct? |
|---|---|---|
| `f9e3f442` (merge base) | 2 | no — one is the label's own local, a false positive |
| `1a939e45` (previous head) | 1 | **yes** |
| `786a3929` (merged head) | **0** | no — the real field read is gone |

A regression against both the merge base and the previous head. An independent 20-fixture construct sweep across the same three revisions found this to be the *only* construct-level difference between the previous head and the merged head.

## Fix

Decide by position within the section: label territory is everything before the section's first statement.

```diff
--- a/internal/parser/languages/csharp.go
+++ b/internal/parser/languages/csharp.go
@@ func csharpLocalScopeOf(n *sitter.Node) csharpLocalScope {
 	for cur := n; cur != nil; cur = cur.Parent() {
+		// A switch SECTION is not a declaration space for the locals its
+		// statement list declares (those live in the switch BODY), but it
+		// IS one for a pattern variable bound in a case label or a `when`
+		// guard. Roslyn accepts `case int x:` beside `case string x:` in a
+		// sibling section AND accepts `case 1: int y = 1;` / `default: y =
+		// 2;`, so the two rules coexist and the answer depends on where in
+		// the section the binder sits: label territory is everything
+		// before the section's first statement.
+		if cur.Type() == "switch_section" && csharpInSwitchLabel(cur, n) {
+			return csharpLocalScope{start: int(cur.StartByte()), end: int(cur.EndByte())}
+		}
 		if csharpScopeFormers[cur.Type()] {
 			return csharpLocalScope{start: int(cur.StartByte()), end: int(cur.EndByte())}
 		}
 	}
 	return csharpLocalScope{start: 0, end: math.MaxInt}
 }
 
+// csharpInSwitchLabel reports whether binder sits in section's LABEL
+// region - before the first statement the section lists.
+func csharpInSwitchLabel(section, binder *sitter.Node) bool {
+	for i, _nc := 0, int(section.NamedChildCount()); i < _nc; i++ {
+		c := section.NamedChild(i)
+		if c == nil {
+			continue
+		}
+		t := c.Type()
+		if strings.HasSuffix(t, "_statement") || t == "block" {
+			return binder.StartByte() < c.StartByte()
+		}
+	}
+	return false
+}
```

Restores the correct answer (1 field read), keeps `TestResolveCSharpTypedLocal_AliveAcrossSwitchSections` and `TestResolveCSharpTypedLocal_SiblingScopeRedeclarationKeepsItsRecord` passing, and leaves 3,525 tests green in `internal/parser/languages` + `internal/resolver` and 1,612 in `internal/indexer` + `internal/indexer/merkle`.

Note `switch_body` is a real node in the vendored `tree-sitter-c-sharp` v0.23.5 grammar, so the current entry is live, not inert — the table just cannot express both rules on its own.

pbednarcik:
Fixes #724.

## What lands

Your patch and helper, verbatim - the positional discriminator in csharpLocalScopeOf (label territory is everything before the section's first statement) plus csharpInSwitchLabel. Additions:

- **Pins.** The eaten-field shape in BOTH label variants (`case 1 when o is Repo repo:` and `case Repo repo:`) - RED at 0 field reads pre-fix, exactly 1 after, matching your three-revision table. Plus the boundary case as a standing guard on the discriminator itself: a pattern variable introduced by an `if` INSIDE a section's statement list stays BODY-scoped (the fixture is deliberately CS0165 - resolution is what is pinned, not definite assignment, per your Roslyn ground truth), with a same-named decoy field of a different type proving the sibling section binds through the local. The two round-6 B1 tests you named keep passing, and the scope-former table's doc comment now states both rules instead of overclaiming `switch_body`.
- **Extractor salt 18 to 19** (own commit): existing stores carry body-wide label-pattern shadows that ate sibling field reads. Note #732 bumps 18 to 19 for the verbatim canonicalization on an independent branch - whichever lands second resolves the one-line conflict by taking 20.

## Evidence

- Full parser/resolver/indexer/tstypes suites: fail-name sets byte-identical to clean-main Windows controls; the fix's pin set individually revert-red.
- On my counted C# fixture repo, a new round pins both rules as battery cells. Cold-lab A/B, base (merged main) vs this branch, same recipe and fixture tree: the eaten-field cell goes 0 reads (the regression) to exactly 1 (the field read, the guard variable's own use riding the local); the boundary-control cell reads identically on both labs (both sibling sites bind through the body-scoped local past the decoy field, no companions). The rest of the battery is byte-identical between the two labs.

### DIFF
--- a/internal/indexer/extractor_version.go
+++ b/internal/indexer/extractor_version.go
@@ -40,7 +40,7 @@ var extractorVersions = map[string]int{
 	//   "go": 2,
 	"c":      generatedParserProjectionPolicyVersion, // generated parser projection covers all strictly detected table sizes
 	"php":    2,                                      // class/interface inheritance now emits typed structural edges
-	"csharp": 20,                                     // indexers and accessor-bearing events are emitted as members and own their body calls (was: verbatim-identifier canonicalization unified across type refs, the base-list prescan, and partial identity)
+	"csharp": 21,                                     // case-label and when-guard pattern variables scope to their switch section (was: indexers and accessor-bearing events are emitted as members and own their body calls)
 	"scala":  2,                                      // explicitly instantiated generic calls emit call edges
 	"go":     3,                                      // generic instantiations are marked so indexing a func value cannot bind (was: generic calls emit call edges)
 	"cpp":    2,                                      // templated and namespace-qualified calls emit call edges
--- a/internal/indexer/extractor_version_test.go
+++ b/internal/indexer/extractor_version_test.go
@@ -229,7 +229,7 @@ func TestStaleLangsDetection(t *testing.T) {
 		}
 
 		for _, path := range []string{"src/Handler.cs", "Views/Page.razor", "Views/Page.cshtml"} {
-			want := policySalt + "|csharp@20"
+			want := policySalt + "|csharp@21"
 			if got := merkleSaltFor(path); got != want {
 				t.Errorf("C# extractor salt for %s = %q, want %q", path, got, want)
 			}
--- a/internal/parser/languages/csharp.go
+++ b/internal/parser/languages/csharp.go
@@ -287,12 +287,16 @@ func (s csharpLocalScopes) shadowsAnywhere(owner, name string) bool {
 // an `if` condition escapes to the ENCLOSING block (definite-assignment
 // scoping), so stopping at the `if` would under-refuse. Loop headers do
 // not leak their pattern variables past the statement. The switch BODY
-// is one declaration space shared by every section — that is exactly
-// why redeclaring a name in a sibling section is CS0128 — so the extent
-// stops at `switch_body`, never at a single section (round-6 finding
-// B1: stopping at the section read a sibling-section local as expired,
-// dropping its receiver evidence and re-minting its assignments as
-// field writes).
+// is one declaration space shared by every section for the locals its
+// STATEMENT LISTS declare — that is exactly why redeclaring a name in a
+// sibling section is CS0128 — so those extents stop at `switch_body`
+// (round-6 finding B1: stopping at the section read a sibling-section
+// local as expired, dropping its receiver evidence and re-minting its
+// assignments as field writes). A pattern variable bound in a case
+// LABEL or `when` guard is the one exception: it is section-scoped
+// (sibling sections may redeclare it), handled positionally in
+// csharpLocalScopeOf rather than by this table — the ancestor node type
+// alone cannot express both rules.
 var csharpScopeFormers = map[string]bool{
 	"block":                       true,
 	"switch_body":                 true,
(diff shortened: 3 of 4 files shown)


zzet (approved):
Approving. I verified this independently rather than waving it through as my own prescription — the last fix I prescribed on this family (#677 round-5 B3) was implemented faithfully and reinstated a false-edge bug, so a patch being mine is a reason for more scrutiny, not less.

## Ground truth re-confirmed

All three Roslyn claims hold on `dotnet` 9.0.305 / `net9.0`:

- `case int x:` beside `case string x:` in sibling sections — compiles clean (section-scoped).
- `case 1: int y = 1;` / `default: y = 2;` — compiles clean (body-scoped).
- The boundary case — `error CS0165: Use of unassigned local variable 'esc'`. That is the load-bearing one: the sibling section binds the *escaped local* and is merely not definitely assigned, rather than falling back to the same-named field. So the discriminator has to be positional, and `csharpInSwitchLabel` is the right shape for it.

## The fix works and the pins bind

A/B against current `main` (`506252c4`, which carries the live `switch_body` regression):

| | `main` | this branch |
|---|---|---|
| `when guard pattern` | FAIL — 0 field reads | PASS — 1 |
| `case label pattern` | FAIL — 0 field reads | PASS — 1 |

Exactly the table in #724. `./internal/parser/languages`, `./internal/resolver` and `./internal/indexer` are all green here.

## Note A — the boundary test does not guard the discriminator

You describe it as "the boundary case as a standing guard on the discriminator itself". It isn't one. I removed the positional check entirely —

```go
if cur.Type() == "switch_section" { // every section binder becomes section-scoped
```

— and `TestResolveCSharp_SectionStatementPatternVarEscapesToBody` stays **green**. The mutant is killed only by the pre-existing `TestResolveCSharpTypedLocal_AliveAcrossSwitchSections`.

So the behaviour is right and the suite does catch the over-correction — but the new test is redundant with one that already existed rather than additive, and the discriminator's real guard is a test that is not part of this change. That matters the day someone loosens `AliveAcrossSwitchSections`. Worth either strengthening the fixture until it fails without the positional check, or just describing it accurately as a companion assertion.

## Note B — the `block` arm is load-bearing and completely unpinned

`csharpInSwitchLabel` treats `t == "block"` as the first statement. That arm has no coverage, and a braced section is idiomatic C# — it is the only shape that reaches it:

```csharp
switch (o) {
    case LbRepo repo: { return repo.Get(1); }
    default:          { return repo.Get(2); }
}
```

Measured on this branch: **1** field read — correct, the fix covers it. With `|| t == "block"` dropped: **0** field reads — #724 returns in full for every braced section — and both the `resolver` and `parser/languages` C# suites stay green.

This is the one thing I would like before merge, and it is a third subtest in the table you already have: the same two classes, the same assertion, braces around the section bodies. Cheap, and it pins the half of the discriminator that currently rides on nothing.

## Merge-order caveat

Confirmed: #732 is open and also moves `csharp` 18 → 19, as you noted. Whichever lands second needs to take 20 — a duplicate 19 means one of the two extractor changes never invalidates existing stores, and the failure is silent.

Nice work on the honesty in the description, incidentally: separating "RED at 0 pre-fix, exactly 1 after" from the standing-guard cell is the right way to present a pin set, and it is what let me go straight at the guard and find Note A.

pbednarcik:
Thanks for going at the guard directly. Both notes are addressed on the rebased branch (now on top of `ca495f5a`, after #736/#738/#739 landed).

**Note B (block arm):** added the third subtest you described, `case label pattern, braced section bodies`: the same two classes and the same assertion with braces around the section bodies. Measured on the branch: 1 field read. With `|| t == "block"` dropped it is the only red in the table (0 field reads), so the arm now rides on a pin.

**Note A (boundary test):** I took the cheaper option and described it accurately. Its doc comment now says it is a companion assertion, not the discriminator's guard: an `if` pattern binder sits inside a statement, so a discriminator that section-scoped every binder under `switch_section` would leave it green. The comment names `TestResolveCSharpTypedLocal_AliveAcrossSwitchSections` as the test that actually catches the over-correction, so nobody loosens it without seeing why it is load-bearing.

**Merge order:** main is at `csharp@20` since #738. This branch now takes **21**; #735 takes 22 and #741 takes 23 on their independent branches, each with the `TestStaleLangsDetection` pin updated in the bump commit. Whichever of the three lands later still resolves the one-line conflict by taking the next number.

Suites green here after the rebase: `internal/parser/languages`, `internal/resolver`, `internal/indexer`.

PR #744, opened 2026-09-02
hooks: silence enforcement only when the graph has no answer

vitaliyslion:
## Summary

**Scope note: the split announced earlier has been undone.** This PR carried an
MCP-server half and a PreToolUse-hook half; #654 landed mid-review and replaced
the transport the hook half was built on (`get_file_summary` over MCP → a
`file_coverage` verb resolved daemon-side), so the hook half was deferred. It is
back, rebuilt on that verb rather than patched onto the old one, and the sampling
scheme it used to rest on is gone rather than corrected.

Both halves answer one question: **when the graph has no symbols for a path, say
why.** "Not indexed" collapses situations that call for opposite next steps —
the walk will never hold this file, the file is held but defines no symbols, the
file is held but the active repo/project filter hides it, and nobody has indexed
it yet. A caller that cannot tell them apart cannot pick a step that works,
which is how a redirect to `read(target:{symbol})` ends up pointing at a vendored
path no symbol lookup can ever answer for, and how a PreToolUse advisory ends up
naming graph tools for a file no graph tool has a row for.

## Server side

- `get_file_summary` reports *why* a file has no indexed symbols.
  `unindexable` / `excluded` / `indexed` / `out_of_scope` ride the
  `file_not_indexed` guidance's `data` block, and `SuggestedTools` narrows to the
  tools that can actually answer: a path the graph will never hold is offered
  `read_file` alone rather than two locators that return zero rows, and a file
  hidden by the active filter is offered `set_active_project`, since every graph
  tool applies that same filter.

- `Indexer.PathIndexability` answers the walk's admission question for a single
  path through the same `admitWalkEntry` verdict both walks use, so the probe
  cannot drift from what the walk does. It deliberately does *not* apply
  `SkipUntrackedAssets`, which by its own config contract governs the cold
  full-index walk alone — applying it reported "never indexable" for files the
  incremental watcher indexes on the next save.

- `Server.pathIndexability` fans it across repos. A path that names its own
  corpus routes to that repo, longest prefix wins; only a bare path is put to a
  vote, and `unanimousPathSkip` fails closed on both disagreement and silence —
  a repo that cannot answer gets no vote, because its zero `PathSkip` is
  bit-identical to "indexable".

- `admitWalkEntry` runs the exclude rules **before** language detection. This is
  hardening, not tidying: `effectiveLanguage` falls through to `readSniffPrefix`,
  which `os.Open`s the file, so language-first opened files inside vendored trees
  — and opened them before `shouldExclude`'s `SymlinkEscapes` guard had refused
  links pointing out of the repo. `walkAdmission` grows an `excluded` flag,
  because exclusion now short-circuits before `lang` is known and both
  `PathSkip.ByRule` and `admitScopedWalkFile` relied on `lang` to tell the
  rejection stages apart.

## Hook side

- **Per-file.** `fileIndexStatus` replaces the `(indexed, count)` pair every read
  door used to branch on. That pair collapsed "excluded by design", "held but
  defines nothing" and "no verdict at all" into a single not-indexed branch, so
  the advisory fired for paths where every tool it names returns nothing.
  `noGraphAnswer` now silences Read and Bash `cat`/`head`/`tail` only on a
  positive verdict, and `Answered` qualifies the negative fields alone — it never
  gates `Covered`, because a daemon predating the field still reports coverage
  truthfully and daemons outlive the binary upgrade that starts them.

- **Per-directory.** The Grep/Glob scope question moved into the daemon as a
  `dir_coverage` control verb, answered in two exact halves: the graph half reads
  the serving view for a file node under the scope, and the walk half runs only
  when the graph holds nothing, putting the directory through `admitWalkEntry`
  under a 500 ms budget. No MCP round trip, no sample, no extension list. That
  leaves the hook a decision small enough to read at once:

  ```go
  func scopeTrackedFromCoverage(result daemon.DirCoverageResult) (hasSource, probeOK bool) {
  	if !result.Answered { return false, false }
  	if result.HasSource { return true, true }
  	return false, result.Walked && !result.Indexable
  }
  ```

  Only a completed walk that claimed nothing proves a scope holds no source. An
  unfinished walk, an unreadable subtree, an exhausted budget and a daemon that
  predates the verb all land on unproven, which keeps enforcement on.

- `pathTracked` gained the single-indexer fallback its sibling `indexerForPath`
  already had. Without it every path on such a daemon reported untracked, and
  `Answered = !Tracked` then claimed the file was outside every corpus —
  silencing the read doors repo-wide.

## Trade-off, measured rather than asserted

The exclude-before-language reorder is not free, and gortex at 94% Go (4157 of
4446 tracked files) is the one place the cost cannot show. Measured against a
synthetic 4000-file tree with a configurable source/vendor mix, A/B against a
worktree with the ordering flipped back, `benchstat` n=10 —
`internal/indexer/admission_bench_test.go` and `bench_nonsource_test.go` are in
the branch:

```
Index_NonSourceTree-16   31.05m ± 1%  →  92.22m ± 3%   +197.06% (p=0.000 n=10)
```

The same tree at gortex's own mix is `+1.62%`, and the real repo is `~` at n=5.
Per entry, `unclaimed_included` is +426% and `unclaimed_vendored` +635%, while
both `claimed` cases are `~`. The whole cost lands on files no extractor claims,
which is the population a vendored tree is made of.

Breaking the reordered prefix down puts the justification and the price in
different places: `pathguard.SymlinkEscapes` is 1.28 µs, `effectiveLanguage`
(sniff open) 5.73 µs, `shouldExclude` 24.1 µs — of which `MatchAbsDir` is ~22 µs.
The reason the order is load-bearing is the sniff open, and satisfying that needs
the 1.28 µs guard hoisted, not the 22 µs exclude matcher. Hoisting only
`SymlinkEscapes` measures at +11% cold index instead of +197%, same security
property.

That alternative is not in this PR. Two things about it are unresolved:
guard-first adds a duplicate lstat on claimed files (+4–5%, fixable by threading
the stat through), and it changes `ByRule` for unclaimed-extension files inside
vendored trees, which changes what `PathIndexability` reports to the hook. The
full ordering discussion is in the thread.

## Known gaps

- `FileCoverage` consults the walk only when the graph holds nothing for a path,
  so a tree excluded *after* it was indexed keeps its nodes and reports covered
  until a re-index drops them.
- `DirCoverageResult.Tracked` is reported by the daemon and not read by the hook.
  `Answered && !Tracked` is a positive "outside every indexed corpus" verdict and
  the file door acts on it, so the two doors currently differ on that case.
- `graphHoldsUnder` finds the first file node under a scope via
  `NodesByKind(KindFile)`. Both reader implementations materialise their result
  before yielding, so the early exit does not shorten the scan.
- The extension-list divergence you named is closed for directory scopes and
  still open for single-file ones. `dir_coverage` removed `looksLikeSourceFile`
  from the scope path entirely, but `hookSearchScope`'s single-file branch still
  returns `NonSource` from it before probing, so a Grep scoped to one indexed
  `main.tf`, `Chart.yaml`, `App.csproj`, `report.qmd` or `init.luau` is silently
  allowed. Reordering it so a positive graph verdict wins is a small change and
  it passes, but it adds a deny rather than removing one and costs a
  `file_coverage` round trip on every single-file scope, docs included — both
  outside this PR's thesis, so it is not in the diff.

## Testing

- [x] `go build ./...` clean
- [x] `go test` green on `internal/hooks`, `internal/indexer`, `internal/mcp`,
      `cmd/gortex`, `internal/daemon`
- [x] `go test -race ./...` — two environment-only exceptions, verified against `main`
- [x] New tests: `internal/hooks/file_index_scope_test.go` (per-door verdict
      matrix), `internal/indexer/admission_test.go`,
      `scope_admission_test.go` (the `main.tf` / `Chart.yaml` / `playbook.yml` /
      `App.csproj` list against the real extractor registry),
      `internal/mcp/path_indexability_test.go`,
      `cmd/gortex/daemon_controller_{coverage,dir_coverage}_test.go`
- [x] Benchmarks committed alongside the numbers above


zzet (changes requested):
This is careful work, and the framing is the right one: the interesting bug was never "stop redirecting reads of vendored files", it was "do not let "we could not ask" masquerade as "we asked, and the answer is no"". `noGraphAnswer` gets that right, and I went through every `!ProbeOK` branch trying to find a fail-open and could not — a failed probe with the daemon up and the file tracked still enforces, and an unreadable registry (`OwnershipUnknown`) still enforces. That is the load-bearing claim of the PR and it holds.

One finding, where the new corroboration path re-creates the failure it was built to prevent.

## The witness walk decides "no source here" from a hardcoded list

`scopeWitnessViaWalk` picks its representative file with `looksLikeSourceFile` — a hardcoded 26-extension list — and `scopeTrackedFromProbes` maps `witnessNone` to a **proven** verdict:

```go
case witnessNone:
	return false, true   // hasSource=false, probeOK=true
```

`probeOK=true` is the claim "we asked, and the answer is no". But the walk did not ask the graph anything; it asked a string list. Every one of these yields `witnessNone` → proven source-free → enforcement silenced for the scope, and every one is a language with a first-class extractor:

```
main.tf      -> witnessNone -> hasSource=false probeOK=true
Chart.yaml   -> witnessNone -> hasSource=false probeOK=true
playbook.yml -> witnessNone -> hasSource=false probeOK=true
App.csproj   -> witnessNone -> hasSource=false probeOK=true
report.qmd   -> witnessNone -> hasSource=false probeOK=true
init.luau    -> witnessNone -> hasSource=false probeOK=true
README.md    -> witnessNone -> hasSource=false probeOK=true
```

So a Terraform module, a Helm chart, an Ansible role, a .NET project directory or a docs tree reads as settled-non-source, and the door goes quiet on exactly the repos where it should not. Your per-file door is right — it is the *selector* that picks the witness that uses a narrower definition of source than the indexer does, so the corroboration is only ever as good as the list.

The one-line fix that matches this PR's own philosophy is to stop calling it proof:

```go
case witnessNone:
	return false, false   // walked it, recognised nothing — not the same as asked
```

The thorough fix is to drop the pre-filter and put the first N regular files through `queryFileIndexScope` until one returns a verdict, so the witness is chosen by the same authority that judges it.

Worth saying plainly: `looksLikeSourceFile` is pre-existing with ten callers, so this is a divergence you inherited rather than introduced. This PR is what makes it decide a whole scope's enforcement, which is why it surfaces here.

## Two smaller ones

**The witness walk is the only probe on this path with no deadline.** Its neighbours all have one — `grepProbeTimeout` 200ms, `readOutlineTimeout` 250ms, `fileIndexedTimeout` 2s. `witnessWalkBudget` bounds entry *count* at 4096 but not duration, and a found witness then costs a further 2s `queryFileIndexScope`. On a cold or network filesystem that is an unbounded stall on the PreToolUse critical path, which is the one path that must never hang.

**The reorder costs more than the trade-off note says.** It is described as one stat per skipped file; it is two syscalls plus matcher work — the hoisted `d.Info()`, plus `shouldExclude` → `pathguard.SymlinkEscapes` → `isSymlink` lstat, plus the exclude and hierarchical-gitignore matchers, all now running on files `effectiveLanguage` used to short-circuit. "Benchmarks run if performance-relevant" is unchecked and this qualifies. One caveat if you do measure it: this repo is 94% Go (4157 of 4446 tracked files), so it is the one place the regression cannot show. It needs a repo with a large non-source population.

## Mechanical

The branch is `DIRTY` — conflicts in `internal/hooks/pretooluse.go` and `internal/indexer/indexer.go`, which are the two files this review is about, so I will want to re-read the resolution. `govulncheck` is the only red check and is very likely the advisory main already fixed in 56836c4f; worth confirming after the rebase rather than assuming.

One thing that cost me a few minutes and may cost you some: GitHub's `baseRefOid` on this PR is stale, so diffing against it shows phantom C# changes including a deleted `resolver/csharp_switch_label_scope_test.go`. Those are not in your branch — `git merge-base HEAD origin/main` is `ca495f5a` and the diff against that is clean. The rebase will clear it.

## Verified, so please do not spend another round on these

- `noGraphAnswer` does not fail open on any `!ProbeOK` path.
- Both admission gates are immutable after construction — `contentAdmissionGate` and `untrackedAssetGate` write their maps once and only read in `skip()` — so sharing them through `probeGateCache` across concurrent probes is race-free and accumulates no budget. That was my main worry about the 30s cache and it is unfounded.
- The git shell-out under the mutex is genuinely bounded: `gitcmd.Output` honours the context, 2s timeout, and the mutex means at most one stall per TTL window. Your trade-off note is accurate.
- Exclude-before-language is real hardening rather than tidying — `shouldExclude` runs `SymlinkEscapes` first, so `readSniffPrefix` no longer opens files inside vendored trees or symlinks pointing out of the repo. Good catch, and worth keeping the comment that explains why the order is load-bearing.
- `unanimousPathSkip` fails closed on both conflict and no-answers, and dropping abstentions is the right call for the reason the doc gives.
- `absWithinRoot` uses the correct Join → Rel → ".." containment pattern.
- 7769 tests pass under `-race` across `internal/hooks`, `internal/indexer` and `internal/mcp` here.

vitaliyslion:
Worth saying up front that this was still a draft and the review is earlier than I would have asked for — but I am glad you looked, because the witness finding is correct and it is the one I would have shipped without seeing.

**Please hold off on re-reviewing until this leaves draft.** What is on the branch right now is an intermediate state, not the proposal. #654 landed mid-review and replaced the transport the hook half was built on — `get_file_summary` over MCP became the `file_coverage` control verb, resolved daemon-side — so the hook-side enforcement is currently trimmed out while I rebuild it on the new verb. I briefly considered shipping the server half alone and following up, but that leaves the PR without the behaviour it exists to change, so the rest is coming back into this PR. I will mark it ready when it does.

**The witness walk.** Conceded without reservation. `witnessNone` claiming `probeOK=true` off a 26-extension string list is the same "we could not ask" wearing "we asked" that this PR exists to kill, one level down, and your `main.tf` / `Chart.yaml` / `playbook.yml` / `App.csproj` list makes it concrete. It is fixed as part of the rebuild rather than patched in place — both of your fixes are captured, and I would rather choose between them against the rebuilt probe than commit to one now. The missing deadline is captured with it, as is your point that the whole witness path needs to fit one budget rather than the walk getting its own.

**Exclude-before-language: kept, and moved.** This part is settled and will not change in the rebuild, so if you want to spend any time on the branch before it is ready, this is the piece worth it. Since you asked for the order explicitly and upstream `admitWalkEntry` is language-first, keeping it meant putting the order into the shared gate rather than a parallel copy — which means I have changed your new code:

- `admitWalkEntry` now runs `shouldExclude` first, with the comment explaining why the order is load-bearing.
- `walkAdmission` grows an `excluded` flag. Exclusion now short-circuits before `lang` is known, so `lang != ""` can no longer stand in for "the rejection was not about being unclaimed" — `PathSkip.ByRule` and `admitScopedWalkFile` both relied on that.
- `admitScopedWalkFile` loses its trailing `!shouldExclude` re-check, which is now unreachable: reaching that branch already means the ignore list let the path through.
- `TestAdmitWalkEntry_SymlinkEscapeRefusedBeforeSniff` pins it.

**The reorder's cost.** Your correction is right and the trade-off note now states it in your terms — two syscalls plus exclude and hierarchical-gitignore matcher work, not one stat. The benchmark box stays unchecked on purpose: your 94%-Go point means a number measured on this repo would be reassuring and meaningless. It needs a repo with a real non-source population, and that is queued before this leaves draft.

**Mechanical.** Conflicts are resolved and the branch is mergeable. `govulncheck` was the x/crypto advisory `main` already fixed in 56836c4f, as you guessed — merging cleared it. The phantom C# diff from the stale `baseRefOid` is gone too; thanks for chasing that down, it would have cost me the same few minutes.

vitaliyslion:
**Rebuilt on the `file_coverage` transport, and the witness is gone.**

The scope question moved to the daemon as a `dir_coverage` control verb, alongside the `file_coverage` verb #654 introduced. Your finding is closed by deletion rather than by a patch, because both fixes you offered were still corrections to a sampling scheme, and the sampling was the problem.

`dir_coverage` answers in two exact halves. The graph half reads the serving view directly for a file node under the scope — the same question the `find_files` probe asked, without the MCP round trip or the sort. The walk half runs only when the graph holds nothing, and puts the directory through `admitWalkEntry`, the same gate the index walk uses. No extension list, no vote, no sample. `internal/indexer/scope_admission_test.go` pins your `main.tf` / `Chart.yaml` / `playbook.yml` / `App.csproj` list against the real extractor registry.

That leaves the hook with a decision small enough to read at once:

```go
func scopeTrackedFromCoverage(result daemon.DirCoverageResult) (hasSource, probeOK bool) {
	if !result.Answered  { return false, false }
	if result.HasSource  { return true, true }
	return false, result.Walked && !result.Indexable
}
```

Only a completed walk that claimed nothing proves a scope holds no source. An unfinished walk, an unreadable subtree, an exhausted budget and a daemon that predates the verb all land on unproven, which keeps enforcement on.

**Also in this push, all from your review or its neighbourhood:**

- `pathTracked` had no single-indexer fallback while its sibling `indexerForPath` did, so on such a daemon every path reported untracked and `Answered = !Tracked` then claimed the file was outside every corpus — silencing the read doors repo-wide.
- `PathIndexability` applied the untracked-asset gate, which by its own config contract governs the cold full-index walk alone. It reported "never indexable" for files the incremental watcher indexes on the next save. Removing it retires `probeAdmissionGates`, the 30s memo and the `git ls-files` shell-out you verified as safe — it was safe, it just should not have been consulted there.
- The multi-repo fallback in `Server.pathIndexability` handed every repo a repo-prefixed path, so it either reached nothing or reached an unrelated file with a colliding spelling. It now routes a prefixed path to its own repo, longest match, and votes only on bare paths.
- `admitScopedWalkFile` grew a second unreachable clause when the first was removed: with `size = -1` the oversize branch cannot fire, so `adm.lang != ""` was dead.
- Comment volume: the branch now removes 62 more comment lines than it adds.

`noGraphAnswer`'s `Tracked` branch I looked at and left. With `pathTracked` fixed the divergence I was worried about is gone, and the daemon is the more precise authority — the hook's own tracked-repo check is a coarse gate a path outside every corpus can still pass.

vitaliyslion:
**The reorder benchmark, and what it says.**

Measured against a synthetic 4000-file tree with a configurable source/vendor
mix rather than this repo, for the reason you gave — gortex at 94% Go is the one
place the regression cannot show. A/B is a worktree at HEAD with the ordering
flipped back to language-first, compared with `benchstat`, n=10.

Cold index, 5% source / 35% vendored:

```
Index_NonSourceTree-16   31.05m ± 1%  →  92.22m ± 3%   +197.06% (p=0.000 n=10)
```

The same tree at gortex's own mix (94% source, 3% vendored) is `+1.62%`, and the
real repo is `~` at n=5. So the ordering is free exactly where it cannot matter
and expensive exactly where it does.

The per-entry picture:

```
AdmitWalkEntry/unclaimed_included-16   5.804µ ± 0%   30.578µ ± 2%  +426.79% (p=0.000 n=10)
AdmitWalkEntry/unclaimed_vendored-16   5.859µ ± 0%   43.090µ ± 1%  +635.51% (p=0.000 n=10)
AdmitWalkEntry/claimed_included-16     28.80µ ± 1%   28.96µ ± 2%   ~ (p=0.956 n=10)
AdmitWalkEntry/claimed_vendored-16     37.27µ ± 2%   37.36µ ± 1%   ~ (p=0.436 n=10)
geomean                                13.82µ        34.55µ       +149.97%
```

Claimed files are unaffected; the whole cost lands on files no extractor claims,
which is the population a vendored tree is made of.

**Where the cost actually is.** Breaking the reordered prefix down:

| step | cost |
|---|---|
| `pathguard.SymlinkEscapes` | 1.28µs |
| `effectiveLanguage` (sniff open) | 5.73µs |
| `shouldExclude` (full) | 24.1µs — of which `excludeMatcher.MatchAbsDir` ~22µs |

That splits the justification from the price. The reason the order is
load-bearing is that `effectiveLanguage` falls through to `readSniffPrefix`,
which opens the file — so it must not run on a path inside a vendored tree or
behind a symlink pointing out of the repo. Satisfying that needs the 1.28µs
guard hoisted, not the 22µs exclude matcher.

Hoisting only `SymlinkEscapes` and leaving the exclude matcher behind the
language gate measures at:

```
AdmitWalkEntry/unclaimed_included-16   7.189µ ± 0%  +23.84% (p=0.000 n=10)
AdmitWalkEntry/unclaimed_vendored-16   7.295µ ± 1%  +24.52% (p=0.000 n=10)
Index_NonSourceTree-16 (cold index)    34.46m ± 1%  +10.99% (p=0.000 n=10)
```

Same security property, +11% instead of +197%.

**Two caveats I have not resolved.** Guard-first adds a duplicate lstat on
claimed files (+4–5%, fixable by threading the stat through). And there is a
`ByRule` behavioural delta for unclaimed-extension files inside vendored trees
that I have not written a test for yet — under guard-first they would be
reported as unclaimed rather than excluded, which changes what
`PathIndexability` tells the hook.

I would rather take guard-first than keep the current ordering, but you called
this part settled, so I am putting the numbers up rather than changing it
unilaterally.

Issue #704, opened 2026-08-30
analyze routes silently drops entire files: 153 routes on v0.63.7, 130 on v0.63.9 (KYC + checkout handlers vanish)


tiendungdev:
## Summary

Between `v0.63.7` and `v0.63.9`, `analyze --kind routes` stopped emitting routes for **entire files**, silently. On the same repository and the same route inventory it returned **153 routes on v0.63.7** and returns **130 on v0.63.9** — the 23 missing ones are not scattered, they are three whole files that never appear in the output at all.

There is no truncation flag, no caveat and no note: the payload simply reports 130 rows as if that were the whole surface.

## The gap

| File | Routes in the repo | In `analyze routes` output |
|---|---|---|
| `internal/http/kyc/handler.go` | 14 | **0** |
| `internal/http/checkout/handler.go` | 8 | **0** |
| `cmd/web/metrics.go` | 1 | **0** |
| everything else | 130 | 130 |
| **total** | **153** | **130** |

153 − 130 = 23, matching the three files exactly.

For context on why this one stings: those are the KYC and checkout handlers — the identity-verification and payment endpoints. An agent asking "what routes does this service expose" is told 130 and has no way to know the authentication/payment/PII surface is missing.

## Environment

- `gortex v0.63.9+16340d2`, Windows 11, native MCP.
- Repository: a Go service using `net/http` `ServeMux` with method-prefixed patterns (`mux.HandleFunc("POST /path/{id}", ...)`).
- Index healthy at measurement time: `workspace{operation:"index"}` → `health_score=100`, `edges_ok=true`, clean path liveness, `orphan_files=0`, `resolution_regressions=0`, `repo_ownership.consistent=true`.
- Ground truth: a generated route inventory in the repo, 153 registrations, verified against source.

## What I ruled out

1. **Not a scoping artifact.** `analyze{kind:"routes"}` and `analyze{kind:"routes", options:{repo:"…"}}` both return exactly 130.
2. **Not a multi-repo workspace artifact.** First measured on a four-member workspace (main checkout + three linked worktrees). I then shrank the workspace to a **single repo**, restarted the daemon, let it fully re-index to `ready`, and re-ran both queries: **still 130, still the same three files missing.**
3. **Not a missing file.** `search{operation:"files", query:"kyc/handler.go"}` returns the file, one node, `language=go`, correct repo.
4. **Not an enrichment-timing artifact.** Measured after `state ready`, and re-measured after enrichment finished; identical.
[...]

PR #746, opened 2026-09-03
feat(mcp): say whether ownership was ever mined instead of answering zero

tiendungdev:
First slice of **#511**, and the one I picked deliberately: it is the smallest surface that needs every part of the contract, so it fixes the vocabulary before the pattern is applied anywhere else.

## The zero being fixed

`analyze kind=ownership` answers `no owners matched` for two states that a caller must act on differently:

- nobody in scope carries an owner — **an answer**;
- git blame was never mined — **the absence of one**.

They render identically. The second one is the sentence an agent turns into "this code is unowned".

## The contract

`contract_tier_unbuilt` already solves this for the contract tier, and its three properties are why it is safe to act on. All three are kept:

1. the zero says it is **not absence evidence**;
2. it names **which** repositories are in that state;
3. it names the recovery **and rules out the one that looks obvious**.

Property 3 carries the most weight here. Indexing never reads `git blame`, so `reindex_repository` — and a full `untrack` / `track`, the heavier thing a caller escalates to — leave the answer exactly as empty as they found it. The recovery string says so by name.

## Five states, from the scan the answer already runs

| state | meaning |
|---|---|
| `never_built` | no symbol in scope carries a stamp |
| `partial` | some repositories stamped, others not |
| `built` (filters) | no symbol was in scope at all — the zero is about `kind` / `path_prefix` |
| `built` (threshold) | owners were found and `min_symbols` removed every one |
| `built` | the data is there and nothing matched |

The `built` half is the valuable one, exactly as the issue argues: it lets a caller say "actually nothing" with the tool's own confidence instead of hedging every zero forever. The threshold case is a third silent zero this handler already had, rendered like the other two.

## Two choices worth arguing

**`partial` and `never_built` are attached even when the answer has rows.** A repository in scope with no blame stamps makes the answer an undercount either way, and a non-empty undercount is the more dangerous shape — an empty answer at least looks suspicious, while rows look like the answer. This is the same argument the route-inventory case in #511 makes: 130 of 153 rows is the reading that gets acted on. Only a `built` state is dropped when rows exist, since rows are their own proof that blame was mined.

**The classification is not built on `graph.EnrichmentState`,** even though that marker exists and would have been the obvious reuse. It is written only by the semantic providers, and `recordEnrichMarker` refuses to record on a dirty tree or without a sha — so a repo enriched under either condition would be reported as never enriched. Deriving the state from the same scan the answer came from is stronger evidence and, more importantly, cannot contradict the number it accompanies. The tallies ride along in the existing loop; no extra graph pass.

## Tests

Five end-to-end cases through `handleAnalyze`, a table over the classifier, and a compact-encoding test — that path is prose only, so a caller there sees the line or nothing.

Thirteen mutants, each failing its test and passing restored: `never_built` and `partial` each collapsed into `built` · an empty scope and a threshold-emptied answer each blamed on enrichment · the caveat naming stamped repositories too · the note dropping its refusal of the absence reading · the recovery no longer ruling out reindex · a `built` state offering a recovery · the candidate and stamped tallies each not counted · the caveat dropped whenever rows exist · a complete answer annotated · the payload never carrying it.

One test failure during development is worth reporting because it changed the test rather than the code: the multi-repo case asserted `repos == ["repo-b"]` and got `["(default) repo-b"]` — the test server's own fixture nodes are unstamped too, so naming them is correct. The assertion now pins the property that matters (the unstamped repo is named, the stamped one is not) instead of a count that depended on the fixture.

## What I would do next, if this shape is right

`co_change` is the same slice with a different source — `tools_cochange.go` already has the presence check the classifier needs. Then the shared-node contract attribution from the second comment on #511. The `routes` partial case I would leave until #704 is understood, because "which files were parsed but produced no rows" is extractor bookkeeping rather than a rendering change, and guessing at it would publish a `complete: false` that is itself unverified.

## Verification

`go build ./...` and `go vet ./internal/mcp` clean; the analyze / ownership / contract-tier test subset passes. Same caveat as my last two PRs: several `./internal/mcp` tests fail on Windows on a pristine tree (drive-letter path shapes), so I verify the affected tests rather than claiming the suite.

### DIFF
--- a/internal/mcp/data_state.go
+++ b/internal/mcp/data_state.go
@@ -66,6 +66,11 @@ type dataStateCaveat struct {
 	// Recovery is the command that would produce the data. Empty when the
 	// state is built, where there is nothing to recover.
 	Recovery string
+	// Eligible and Stamped size the shortfall: how many symbols the pass
+	// admits in scope, and how many carry data. A caller deciding whether to
+	// act on a partial answer needs the magnitude, not only the verdict.
+	Eligible int
+	Stamped  int
 	// Note carries the reading a caller must not make, in prose.
 	Note string
 }
@@ -81,6 +86,10 @@ func (c dataStateCaveat) payload() map[string]any {
 	}
 	if c.Recovery != "" {
 		out["recovery"] = c.Recovery
+	}
+	if c.Eligible > 0 {
+		out["symbols_eligible"] = c.Eligible
+		out["symbols_stamped"] = c.Stamped
 	}
 	return out
 }
@@ -108,32 +117,53 @@ const blameDataStateSource = "blame_enrichment"
 // so neither a re-index nor a re-track produces a single stamp.
 const blameRecovery = "run `gortex enrich blame` (or `gortex enrich all`). reindex_repository and untrack/track will NOT stamp authorship: indexing never reads git blame, so a re-parse of every file still leaves it empty."
 
-// ownershipDataState classifies an empty ownership answer from the counts its
-// own scan produced.
+// ownershipDataState classifies an ownership answer from the counts its own
+// scan produced.
 //
 // candidates and stamped are per-repo tallies over exactly the nodes the
 // handler considered: candidates counts every symbol that passed the kind and
-// path filters, stamped counts those carrying blame authorship. owners is the
-// number of distinct owners found before the min_symbols filter — non-zero
-// there with no rows means the data was fine and the threshold removed the
-// answer, which is a third silent zero this handler used to render
-// identically to the other two.
+// path filters AND that the blame pass would admit (blame.Eligible), stamped
+// counts those carrying authorship. owners is the number of distinct owners
+// found before the min_symbols filter — non-zero there with no rows means the
+// data was fine and the threshold removed the answer, which is a third silent
+// zero this handler used to render identically to the other two.
+//
+// Coverage is compared, not presence. One stamped symbol proves the pass ran
+// over that repository; it does not prove the repository is covered.
+// blame.EnrichGraph is best-effort per file — a file git cannot blame is
+// skipped and the pass reports success — so a single repository routinely
+// holds both stamped and unstamped eligible symbols. Reading a single stamp as
+// full coverage published exactly the misleading non-empty undercount this
+// caveat exists to prevent.
+//
+// The counting population is blame's own admission set for the same reason in
+// reverse: a symbol the pass never looks at is not a coverage hole, and
+// counting it would report a shortfall no enrichment could close.
 func ownershipDataState(candidates, stamped map[string]int, owners int) dataStateCaveat {
-	var unstamped []string
+	var unmined, incomplete []string
 	covered := 0
 	total := 0
+	eligibleTotal, stampedTotal := 0, 0
 	for repo, n := range candidates {
 		if n == 0 {
 			continue
 		}
 		total++
-		if stamped[repo] > 0 {
+		eligibleTotal += n
+		got := stamped[repo]
+		stampedTotal += got
+		switch {
+		case got == 0:
+			unmined = append(unmined, repo)
+		case got < n:
+			incomplete = append(incomplete, repo)
+		default:
 			covered++
-			continue
 		}
-		unstamped = append(unstamped, repo)
 	}
+	unstamped := append(append([]string{}, unmined...), incomplete...)
 	sort.Strings(unstamped)
+	sort.Strings(unmined)
 
 	switch {
 	case total == 0:
@@ -145,13 +175,15 @@ func ownershipDataState(candidates, stamped map[string]int, owners int) dataStat
 			Source: blameDataStateSource,
 			Note:   "no symbol matched the kind / path_prefix filters, so this empty result is about the filters rather than about authorship data.",
 		}
-	case covered == 0:
+	case stampedTotal == 0:
 		return dataStateCaveat{
 			State:    dataStateNeverBuilt,
 			Source:   blameDataStateSource,
-			Repos:    unstamped,
+			Repos:    unmined,
 			Recovery: blameRecovery,
-			Note:     "not one symbol in scope carries a blame stamp, so this empty result is NOT evidence that nobody owns this path — authorship is stamped by a separate enrichment pass and is absent until it runs.",
+			Eligible: eligibleTotal,
+			Stamped:  stampedTotal,
+			Note:     "not one symbol in scope carries a blame stamp, so this result is NOT evidence that nobody owns this path — authorship is stamped by a separate enrichment pass and is absent until it runs.",
 		}
 	case len(unstamped) > 0:
 		return dataStateCaveat{
@@ -159,7 +191,9 @@ func ownershipDataState(candidates, stamped map[string]int, owners int) dataStat
 			Source:   blameDataStateSource,
 			Repos:    unstamped,
 			Recovery: blameRecovery,
-			Note:     "these repositories hold candidate symbols but no blame stamps, while others in scope are stamped. Any ownership answer over this scope is an undercount, and an empty one says nothing about the unstamped repositories.",
+			Eligible: eligibleTotal,
+			Stamped:  stampedTotal,
+			Note:     "these repositories hold symbols the blame pass admits but did not stamp — either it never ran over them, or it ran and skipped files it could not blame, which it does silently. Any ownership answer over this scope is an undercount: the rows below are real, and the symbols behind the shortfall are invisible to them.",
 		}
 	case owners > 0:
 		return dataStateCaveat{
--- a/internal/mcp/data_state.go
+++ b/internal/mcp/data_state.go
@@ -1,0 +1,177 @@
+package mcp
+
+import (
+	"sort"
+	"strings"
+)
+
+// An empty analyze result has more than one cause, and the caller cannot tell
+// them apart from the number alone: the backing data may never have been
+// built, it may cover only part of the scope, or the query may have been
+// answered in full and found nothing. Only the last one is safe to act on, and
+// it is the one an agent assumes.
+//
+// contract_tier.go answers that question for the contract tier. This file
+// generalises the same three properties to any surface whose data is produced
+// by a pass separate from indexing:
+//
+//  1. say the zero is not absence evidence;
+//  2. name WHICH scope is in that state, not just that some are;
+//  3. name the recovery, and rule out the recovery that looks obvious but
+//     does nothing.
+//
+// The third property is the one that saves the most time. Every one of these
+// surfaces is populated by an enrichment pass, and reindex_repository — the
+// call a caller reaches for first — never runs one.
+//
+// Deliberately NOT built on graph.EnrichmentState. That marker is written only
+// by the semantic (LSP) providers, and it refuses to record on a dirty tree or
+// without a sha, so a repo enriched under either condition would be reported
+// as never enriched. Each caveat here is instead derived from the same data
+// the answer itself was computed from, which is stronger evidence and cannot
+// contradict the number it accompanies.
+
+// dataStateCaveatKey is the stable token a client matches on. It names the
+// condition rather than the presentation, so callers can branch without
+// parsing prose.
+const dataStateCaveatKey = "data_state"
+
+const (
+	// dataStateNeverBuilt: the enrichment that populates this surface has
+	// never run over the scope, so the empty result carries no information
+	// about the question that was asked.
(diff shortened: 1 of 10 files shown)


zzet (changes requested):
Requesting changes for one blocking correctness issue:

**[P1] Treat same-repository partial blame coverage as partial**

`ownershipDataState` currently treats `stamped[repo] > 0` as complete coverage. That only proves one candidate symbol was mined. `blame.EnrichGraph` is explicitly best-effort per file and skips failed or unblameable files, so a single repository can contain both stamped and unstamped candidates.

In that state the classifier returns `built`; when any owner row survives, `handleAnalyzeOwnership` removes the caveat and publishes a silently incomplete ownership answer—the exact misleading non-empty undercount this PR is intended to prevent.

Please align candidate counting with blame eligibility and classify incomplete eligible coverage as `partial`, or use a durable per-repository completion marker.

Required regression coverage:

- classifier: candidates=3 and stamped=1 in the same repository => `partial`
- handler: one stamped and one unstamped same-repository symbol retains `data_state` even with owner rows
- incremental reindex after blame enrichment reports partial coverage

tiendungdev:
Confirmed and fixed in `e5d89183`.

I read `EnrichGraph` before changing anything, and the mechanism is exactly as you describe:

```go
for path, nodes := range byPath {
    lines, err := Run(repoRoot, path)
    if err != nil || len(lines) == 0 {
        continue          // ← file skipped, pass still returns success
    }
```

plus `pickLatest` returning nil for a symbol whose lines carry no blame data. Both leave an eligible symbol unstamped inside a repository that has other stamps. `stamped[repo] > 0` proved the pass had run over the repository and nothing else — and since the handler drops the caveat once rows exist, the published answer was a silently incomplete rollup. Same failure as the one this PR is about, one level down.

## The fix

Coverage compared per repository: `stamped == 0` is unmined, `0 < stamped < eligible` is incomplete, and both report `partial` with the repository named. `never_built` is now reserved for a scope where nothing at all is stamped.

## Aligning the population, in both directions

I took the eligibility route rather than a durable marker. Two reasons, and the second is the one that decided it.

`blame.Eligible(n)` is exported and **`EnrichGraph` itself now calls it**, replacing the two inline guards, so the reporter and the pass cannot drift into disagreeing — they are one decision with one definition.

The direction I want to flag, because it is the trap in "just count everything unstamped": counting symbols the pass never admits would report a shortfall **no enrichment could ever close**. A repository containing one `KindFile` node would sit at `partial` forever, and a caveat that never clears is one a caller learns to skip — which costs more than the bug being fixed. So the population is blame's admission set, and the `stamped` tally is filtered the same way, so a stale stamp on a symbol that is no longer eligible cannot fill a gap it is not part of.

The caveat also now carries `symbols_eligible` / `symbols_stamped`. Deciding whether to act on a partial answer needs the size of the shortfall, not only the verdict.

## Requested coverage, all three

- **classifier** — `candidates=3, stamped=1` in one repository → `partial`, plus the one-short case and a partly-covered repo alongside a wholly unmined one.
- **handler** — `TestAnalyzeOwnership_PartialCoverageInsideOneRepoKeepsTheCaveat`: one stamped and one unstamped same-repository symbol, an owner row survives, `data_state` stays.
- **incremental reindex** — `TestAnalyzeOwnership_ReindexAfterEnrichmentReportsPartialCoverage` runs a real `git init`/commit, a real `blame.EnrichGraph`, asserts the fully-stamped scope carries **no** caveat, then adds a file and reindexes. The answer keeps its rows and reports `partial`. The clean-state assertion is deliberate: without it the test would pass on a classifier that always says partial.

## Two mutants survived, and both were real test gaps

Eight mutants; six died immediately. Two did not, and neither was a spurious mutation:

- **`stamped` ignoring eligibility** survived because no test had a stamp on an ineligible symbol. That mutation lets one stale stamp make a short repository read as covered — the P1 again, through a different door. Now covered by `TestAnalyzeOwnership_AStampOnAnIneligibleSymbolCannotFillTheGap`.
- **`Eligible` returning true for every kind** survived because the ownership handler's own kind filter kept non-blame kinds out of reach. Now covered by a table in `internal/blame` asserting `Eligible` against `shouldEnrichBlame` for fifteen kinds.

Both are the lesson from the last two rounds arriving on schedule: a dying mutant proves a test binds something, not that the rows span what production can produce.

## One test change worth reporting

`TestAnalyzeOwnership_BuiltZeroIsReportedAsReal` and `..._CompleteAnswerCarriesNoCaveat` started failing, and the classifier was right: `setupTestServer` seeds unstamped eligible symbols, so that fixture genuinely has partial coverage and can no longer be used to assert the *absence* of a caveat. Both moved to an isolated empty graph rather than being relaxed.

## Verification

`go build ./...`, `go vet ./internal/mcp ./internal/blame` clean; the ownership and blame suites pass.

One environment caveat, since a `git`-driven test is new here: `TestEnrichGraph_PersonNodeRepoScoped` fails intermittently on my machine with `TempDir RemoveAll cleanup: unlinkat …\.git: The directory is not empty`. That is a cleanup error, not an assertion, it reproduces on a pristine `blame.go` (1/4 runs, versus 2/4 with my change — same noise), and I traced it to my own toolchain: the `git` on my PATH is a fork that writes `.git/ai/logs` and `.git/ai/working_logs` on every `git init`, so a temp `.git` is not reliably empty when the test ends. Nothing to fix upstream, but worth naming in case my new `git init` test ever looks flaky in CI for a different reason.

internal/mcp/tools_enhancements.go
```diff
@@ -1406,14 +1406,32 @@ func (s *Server) handleAnalyzeOwnership(ctx context.Context, req mcp.CallToolReq
...
+		// enrichment could ever close.
+		if blame.Eligible(n) {
+			candidatesByRepo[n.RepoPrefix]++
+		}
 		la, ok := lastAuthoredFrom(ownBlame, n)
 		if !ok {
 			continue
 		}
+		if blame.Eligible(n) {
+			stampedByRepo[n.RepoPrefix]++
```
zzet, line 1433: **[P1] Cross-repository blame can be certified as complete**

The new per-repository completeness check treats any blame row on a node as proof that its `RepoPrefix` was covered, but multi-repo enrichment calls `EnrichGraph` on the combined graph for each repository root. `stripRepoPrefix` maps a foreign node onto the current root whenever both repositories share the same relative path, so repo A blame can stamp a repo B node. The counts then match and this handler can publish `built` and drop `data_state` for repo B even though its ownership data came from repo A.

Scope `EnrichGraph` to the target repository prefix and add a two-repository same-relative-path regression, including the empty-prefix single-repo case.

internal/mcp/data_state.go
```diff
@@ -0,0 +1,211 @@
...
+	case total == 0:
+		// Nothing was in scope to be owned. Blame state is irrelevant: no
+		// enrichment would put a row here, so reporting one as missing would
+		// send the caller after the wrong thing.
+		return dataStateCaveat{
+			State:  dataStateBuilt,
+			Source: blameDataStateSource,
+			Note:   "no symbol matched the kind / path_prefix filters, so this empty result is about the filters rather than about authorship data.",
+		}
+	case stampedTotal == 0:
```
zzet, line 195: **[P2] `never_built` claims run history that is not observable**

`stampedTotal == 0` is reported as `never_built` and says enrichment is absent until it runs, but `EnrichGraph` is best-effort: every git-blame failure or empty result is skipped while the pass can still return success. A completed pass can therefore produce zero stamps, leaving clients with a false state and a recovery command that cannot clear it.

Use an observational state/name or persist per-repository pass outcomes before claiming never-run.

internal/mcp/data_state.go
```diff
@@ -0,0 +1,211 @@
...
+		return dataStateCaveat{
+			State:    dataStateNeverBuilt,
+			Source:   blameDataStateSource,
+			Repos:    unmined,
+			Recovery: blameRecovery,
+			Eligible: eligibleTotal,
+			Stamped:  stampedTotal,
+			Note:     "not one symbol in scope carries a blame stamp, so this result is NOT evidence that nobody owns this path — authorship is stamped by a separate enrichment pass and is absent until it runs.",
+		}
+	case len(unstamped) > 0:
```
zzet, line 209: **[P2] Partial coverage hides a simultaneous threshold zero**

`partial` is returned before the `owners > 0` threshold case is considered. When coverage is partial and `min_symbols` filters out every known owner, the response only reports the enrichment shortfall and even says "the rows below are real" although there are no rows.

Preserve both causes, or pass the post-threshold row state into this rendering, and add a partial-plus-threshold regression.

zzet (changes requested):
Requesting changes. The earlier same-repository partial-coverage issue is fixed, but the new completeness contract is still unsound in multi-repository graphs: existing cross-repository blame projection can be counted as coverage and suppress `data_state`. Two additional state/rendering cases also need correction before this can safely claim that an ownership result is complete.

tiendungdev:
All three fixed in `54ec3dc8`. Every one confirmed against the source before I touched it.

## P1 — cross-repository blame counted as coverage

Confirmed, and it is worse than a classifier bug: `EnrichGraph` was **writing** the wrong stamps, and my completeness check merely believed them.

`stripRepoPrefix` tries the node path under the current root, then retries without its leading segment. So for a pass over repo-a's root, a repo-b node at `repo-b/internal/foo.go` misses, gets trimmed to `internal/foo.go`, finds **repo-a's** file, and takes repo-a's authors. The result is a stamp that is well-formed, plausible, and about a different file.

`EnrichGraph(g, root, repoPrefix)` now admits only nodes carrying that prefix, mirroring `cochange.EnrichGraph`. Both call sites already had the prefix in hand and were dropping it — `handleAnalyzeBlame` was iterating `for prefix, root := range roots` and passing only `root`.

Regression as requested, plus the direction that keeps the daemon honest:

- two repositories sharing `internal/foo.go`: after a repo-a pass, repo-a's node is stamped and repo-b's is not;
- the empty-prefix single-repo case still stamps — `""` had to keep meaning "this graph's only repository", not "no repository";
- a prefixed pass does not reach the unprefixed shared-externals bucket.

## P2 — `never_built` named a history I cannot see

You are right, and the fix is a rename rather than a reword, because the same objection applies to the state at the other end. Both named a pass; neither named an observation:

| was | is | what it means |
|---|---|---|
| `never_built` | **`absent`** | nothing in scope carries this data |
| `partial` | `partial` | some does, some does not |
| `built` | **`complete`** | every symbol the pass admits carries it |

I renamed `built` too on my own initiative — say the word and I will put it back. My reasoning: leaving it would keep one history claim in a trio the other two no longer make, and `complete` is what the code can actually check.

The `absent` note no longer says enrichment "is absent until it runs", and the recovery now says what a state that *survives* the pass means: the remaining files are ones git cannot blame — an unborn commit, an untracked or generated file — not ones nobody has enriched. Without that sentence the recovery is a loop.

`TestOwnershipDataStateNamesNoRunHistory` asserts the note contains none of "never ran", "never been", "until it runs", "has not run", so the claim cannot come back through a later reword.

## P2 — the threshold zero was hidden behind the shortfall

Confirmed: `partial` returned before the `owners > 0` case, so an answer that was both short on coverage and emptied by `min_symbols` reported one cause and, worse, said "the rows below are real" with no rows below.

The classifier now takes the post-threshold row count and **composes** both causes rather than choosing one. The partial wording no longer promises rows either — "whatever rows it returns are real" holds whether there are none or many.

## The mutant that survived

Seven mutants, six died at once. The seventh — `thresholdEmptied := rows == 0`, dropping the `owners > 0` half — survived, and it was a real gap: with no owners at all, that mutation appends "owners were found and min_symbols removed every one of them" to the `absent` note. A false sentence, in the one state where a caller has the least information to check it against. Now asserted in both the absent and the genuinely-empty-complete cases.

## One test-harness note

The new blame tests read authorship from the sidecar as well as node `Meta`. My first version checked `Meta` only, and every assertion passed for the wrong reason: a store implementing `BlameEnrichmentWriter` writes the rows and never touches `Meta`, so both the "stamped" and "not stamped" checks read empty. The positive assertions caught it — which is the argument for always pairing a negative assertion with a positive control in the same test.

## Verification

`go build ./...` and `go vet ./internal/mcp ./internal/blame ./cmd/gortex` clean; the blame and ownership suites pass.

The mutant run for `internal/blame` is scoped to the new scope tests plus the eligibility tables. The package's older git-heavy cases carry a `TempDir` cleanup flake on this machine — a `git` fork on my PATH writes `.git/ai/logs` after the test body ends — and including them would let a mutant read as BINDS for a reason unrelated to the mutation.

zzet (approved):

Issue #685, opened 2026-08-28
tools/list byte ceiling: the core preset has 229 bytes of headroom, and nothing surfaces that until CI is red


tiendungdev:
`TestToolsListByteCeilings` is doing its job, but the `core` preset is close enough to its ceiling that ordinary documentation work now fails it, and the margin is invisible until it is already gone. Reporting this with numbers rather than proposing a fix, since which way to go is your call.

## Where it stands on `main` today

Measured at `06939fce`, `go test -run TestToolsListByteCeilings ./internal/mcp`:

| preset | bytes | ceiling | headroom |
|---|---:|---:|---:|
| agent | 29,332 | 29,700 | 368 (1.24%) |
| loc | 21,039 | 21,350 | 311 (1.46%) |
| core | **96,271** | **96,500** | **229 (0.24%)** |
| full | 254,460 | 289,808 | 35,348 (12.20%) |

`core` is five to ten times tighter proportionally than its siblings, and it is the one that grows whenever any tool description gains a sentence.

## How it behaves in practice

While working on #647 I hit it three times in one PR, twice turning CI red:

1. Documenting the recursive `dir/*` exception in the `fidelity_globs` schema — the first wording measured 96,690 against the 96,500 ceiling and failed. A shared description costs roughly twice its character count, because two tool registrations carry it.
2. Documenting the `dir/*` rule in the `find_files` schema — 96,505, failed by 5 bytes.
3. Disclosing the new admission limits, which you asked for in review — 96,745, failed by 245. Headroom at that point was 35 bytes, so there was no wording that both said the required thing and fit. I made room by tightening existing prose in the same constant rather than moving the ceiling, and flagged it in the PR as a judgement call.

That last one is the shape worth naming: **a review asked for more documentation and the budget made it impossible**, and the only way through was editing text nobody had asked me to touch.

## Two things that make it sharper than the number suggests

**The margin is invisible in a normal run.** The per-preset figures come from `t.Logf`, so they appear only under `-v` or on failure. A green CI run prints:

```
ok  github.com/zzet/gortex/internal/mcp  6.040s
```

Nobody learns there are 229 bytes left until someone spends a review round discovering it.

**The assertion reads as a ratchet, not a budget.** The failure message is *"core preset must shrink below its pre-diet baseline (96500)"*. A ratchet is the right tool for finishing a diet; once the diet is done, a fixed number that only ever gets closer is a trap for the next person who improves a tool description.

## Options, in your hands

[...]

Issue #674, opened 2026-08-26
search_text does not disclose limit-bound truncation: count reports the ceiling as the total, so a capped result is indistinguishable from a complete one


chenroth-island:
**Describe the bug**

When a `search_text` response is bound by `limit` rather than by the corpus, nothing in the response says so. `count` is set to `len(enriched)`, so the array and the count corroborate each other at the ceiling: a truncated result is byte-indistinguishable from a complete one.

The byte budget already handles this correctly — it sets `_truncated_by_budget` and leaves `count` exact. The limit path has no equivalent.

This is the disclosure half of #672. That issue is about the ceiling being low (1000); this one is about truncation being undetectable at *any* ceiling. Raising the limit without adding disclosure just moves the silent cliff, so this is arguably the more important of the two — as noted there, the flag matters more than the number.

**Why this bar specifically**

`search_text` is described as "the alt grep backbone", and agents reach for it in place of `grep`/`ripgrep`. ripgrep's contract is that a result set is either complete or the process tells you why not; it never returns a plausible prefix silently. An agent cannot verify a search result against anything except a second tool, so a partial answer that looks whole doesn't degrade quality gracefully — it produces confident wrong conclusions ("this symbol has 1000 references" when it has 8963, or a refactor that misses two thirds of the call sites).

**To Reproduce**

Any query whose real match count exceeds the effective limit:

```
gortex call search --arg operation=text --arg query=<common term> \
  --arg 'limit:=100000' --arg 'max_bytes:=200000000' --index <repo> --format json
```

Response on a 63,820-file TypeScript repo, `git grep` ground truth **8963**:

```json
{ "query": "Typography", "count": 1000 }
```

No flag, no cursor, and `count` reports the ceiling as the total.

**Expected behavior**

The response states that the limit bound the result, and does not present the ceiling as an exact total.

**Environment:**
- OS: macOS 26.5.1 (25F80)
- Go version: 1.26.5
- Gortex version: v0.63.7+0988412 / built from `main@27b83069`

**Additional context**

I have this working locally and it is small — `internal/mcp/tools_search_text.go`, additive, no behavior change for complete results. Keep the requested limit before the clamp:

```go
requestedLimit := limit
if max := searchTextMaxLimit(); limit > max {
    limit = max
}
```

then disclose at response construction:

```go
if limit > 0 && len(enriched) >= limit {
    resp["_truncated_by_limit"] = true
    resp["_limit_applied"] = limit
    resp["count_is_exact"] = false
[...]

Issue #672, opened 2026-08-26
search_text: limit is clamped to 1000 and the clamp also truncates count, so a capped result is indistinguishable from a complete one


chenroth-island:
**Describe the bug**

`search_text` clamps the requested `limit` to 1000 (`internal/mcp/tools_search_text.go:52`). The clamp bounds the `matches` array *and* the reported `count`, and sets no truncation flag — so a caller cannot tell a capped result from a complete one. There is no cursor in the response to page past it, and `path` filtering is applied *after* truncation, so it cannot recover the remainder either.

This matters more for `search_text` than for the two siblings sharing the same clamp (`tools_find_declaration.go:71`, `tools_graph_query.go:56`): it is the literal-search backbone agents use in place of `grep`, where an unbounded sweep is the normal request. The byte budget (`max_bytes`) already bounds response size, and does so *detectably* via `_truncated_by_budget`.

*(Repo and workspace names below are anonymized.)*

**To Reproduce**

1. Track a large TypeScript monorepo (63,820 files here, called `repo-ts` below).
2. Run:
   ```
   gortex call search --arg operation=text --arg query=Typography \
     --arg 'limit:=100000' --arg 'max_bytes:=20000000' \
     --index ~/work/repo-ts --format json
   ```
3. Response: `count: 1000`, 1000 matches, no `_truncated_by_budget`, no cursor.
4. Ground truth: `git grep -nP Typography | grep -cE '^[^:]*:[0-9]+:'` → **8963**.

**Expected behavior**

Either the requested `limit` is honored, or the response says it was bound — a flag, or a `count` that stays exact the way the byte budget keeps it exact. Silent loss of 7,963 of 8,963 matches, with a `count` that corroborates the wrong number, is the one failure mode a grep replacement cannot have.

**Environment:**
- OS: macOS 26.5.1 (25F80)
- Go version: 1.26.5
- Gortex version: v0.63.7+0988412

**Additional context**

Measured losses, all with the flag unset:

| query | returned | real | lost |
|---|---|---|---|
| `repo-ts` / `Typography` | 1000 | 8963 | 7963 |
| `repo-ts` / `useState` | 1000 | 5071 | 4071 |
| `repo-go` / *(4-char term)* | 952 | 3415 | 2463 |
| `repo-ts` / `useEffect` | 1000 | 3156 | 2156 |
| `repo-agent` / `Installer` | 1000 | 2271 | 1271 |
[...]

PR #751, opened 2026-09-04
feat(mcp): disclose when search_text was bound by limit, and make the ceiling configurable

tiendungdev:
Closes **#674**, and the ceiling half of **#672**.

## The zero-cost failure

`search_text` is described as the alt grep backbone, and a result bound by `limit` was byte-indistinguishable from a complete one: `count` is `len(enriched)`, so the array and the count corroborated each other at the ceiling. On the repo in #672 that is `count: 1000` against a `git grep` truth of 8963, with no flag, no cursor, and nothing else in the payload to doubt.

The byte budget has always disclosed its own truncation (`_truncated_by_budget`). The limit path had no equivalent.

## What a truncated response now carries

```jsonc
{
  "count": 1000,
  "_truncated_by_limit": true,
  "_limit_applied": 1000,
  "count_is_exact": false,
  "_limit_requested": 100000,     // only when the ceiling, not the caller, chose it
  "truncation_note": "…"
}
```

The note follows the `contract_tier_unbuilt` shape that already works elsewhere in this codebase: say what the number means, say how to widen, **and rule out the widening that looks obvious but does not work**. Here that last part is `path`: the filter runs over what survived truncation rather than over the corpus, so slicing a subtree returns whatever was left of the global cut. #672 lists that as one of three workarounds it had to rule out by hand.

## The ordering is the fix

`rawMatches` is counted **before** the path and scope filters, and that is the substance rather than an implementation detail:

```go
matches = <searcher>(query, …, limit)   // stops at limit
rawMatches := len(matches)              // ← here
matches = filterTextMatchesByPath(…)    // runs over what survived
matches = s.filterTextMatchesByResolvedScope(…)
```

A response holding 952 matches can be a truncated 1000. Measuring after the filters would miss exactly the case a caller cannot detect from the outside — and the 952 / 998 rows in #672's table are that shape, which is why the issue notes that even `len(matches) == 1000` is not a reliable client-side detector.

`TestSearchText_TruncationIsMeasuredBeforeThePathFilter` pins it with a fixture where the count necessarily lands below the limit: four matching files, `limit: 3`, filtered to a directory holding two. Whichever three the searcher returns, at most two survive, and the disclosure must still fire.

Landing on the effective limit is the signal, so this can fire on a corpus holding exactly `limit` matches. A spurious "verify this" is the safe direction to be wrong in.

## The ceiling

`GORTEX_SEARCH_TEXT_MAX_LIMIT`, defaulting to 1000 so no existing caller's response changes shape. An unset, unparseable or non-positive value keeps the default rather than lifting the bound — a typo must not become an unbounded scan.

Raising the ceiling without the disclosure only moves the silent cliff, which is why they land together. This is why I marked #674 as the one this closes and #672 as refs: the flag is the fix, the override is the affordance.

## Tests

Five end-to-end cases through the handler and tables over both pure helpers. Ten mutants, each failing its test and passing restored: the flag never published · `count_is_exact` always true · landing exactly on the limit not counted as truncation · truncation measured after the filters · the clamped request not disclosed · the disclosure riding on every response · the ceiling override ignored · an unusable ceiling value honored · the note dropping the recovery that does not work · the applied limit not reported.

Three of those initially came back as "nothing ran" rather than as failures — each mutation orphaned a variable and the package stopped compiling, which is a void result, not a dead mutant. They were re-run in compile-clean form (`if _ = rawMatches; …`) so all ten are real.

## Verification

`go build ./...` clean; the `SearchText` suite passes.

Two measurements worth reporting rather than asserting:

**The schema addition spends 41 of the 62 bytes the core preset has left.** My first version spent 233 and turned CI red on `TestToolsListByteCeilings` (`core` at 97671 against a 97500 baseline; `main` sits at 97438). The description now carries only the part an agent needs before calling — that a bound result sets `_truncated_by_limit` — and the recovery detail rides on the response in `truncation_note` instead. Both ceiling tests pass at 97479.

I had originally checked `TestCompactToolsListIsStaticAndBudgeted` instead, measured 14984 bytes on both trees, and reported that as evidence the addition was free. That number is about the facade surface, which `search_text` is not on; it was the wrong ceiling. Headroom as measured on `main`: **62 bytes** on the core preset, **16 bytes** on the facade list — both recorded on #685.

**One unrelated test is order-dependent.** `TestSearchTextRefusalIsTheCapabilityEvaluationsRefusal` failed for me, and I did not call it flaky until I had the mechanism: on a cold ref-view build it gets `view_building: … retry after 2s` instead of the `capability_unavailable` refusal it asserts, and on the next run — the build now cached under a content-hashed id that survives the per-test `t.TempDir()` — it passes. I reproduced both states on my branch and confirmed the pristine tree passes only in the warm state. Nothing in this change touches view building. Filing it separately.

### DIFF
--- a/internal/mcp/tools_analysis.go
+++ b/internal/mcp/tools_analysis.go
@@ -67,7 +67,7 @@ func (s *Server) registerAnalysisTools() {
 			mcp.WithDescription("Trigram-accelerated literal (or regexp) code search across the indexed repository — the alt grep backbone. Each hit carries the enclosing graph symbol (symbol_id / symbol_name) so you see which function or method a match landed in without a follow-up call. A trigram index narrows the candidate files, so a repo-wide search costs roughly the size of the matching files, not the whole tree. Use for literal-string / regexp lookups; use search_symbols for symbol-name / concept queries."),
 			mcp.WithString("query", mcp.Description("Literal substring (case-sensitive) to search for — or a regular expression when regexp=true.")),
 			mcp.WithBoolean("regexp", mcp.Description("Treat query as a regular expression instead of a literal substring. An invalid pattern is returned as a tool error. Default false.")),
-			mcp.WithNumber("limit", mcp.Description("Max matching lines to return (default 100, capped at 1000; raise the cap with GORTEX_SEARCH_TEXT_MAX_LIMIT). A result bound by the limit carries _truncated_by_limit and count_is_exact=false — read those before treating count as a total, because a `path` filter cannot recover the remainder.")),
+			mcp.WithNumber("limit", mcp.Description("Max matching lines to return (default 100, capped at 1000). A bound result sets _truncated_by_limit.")),
 			mcp.WithString("path", mcp.Description("Restrict matches to one or more sub-paths (comma-separated) -- a monorepo-service slice. Anchored, slash-segment-boundary prefixes relative to the repo root.")),
 			mcp.WithString("repo", mcp.Description("Restrict matches to a single repository prefix.")),
 			mcp.WithString("project", mcp.Description("Restrict matches to repositories in a specific project.")),
--- a/internal/mcp/tools_analysis.go
+++ b/internal/mcp/tools_analysis.go
@@ -67,7 +67,7 @@ func (s *Server) registerAnalysisTools() {
 			mcp.WithDescription("Trigram-accelerated literal (or regexp) code search across the indexed repository — the alt grep backbone. Each hit carries the enclosing graph symbol (symbol_id / symbol_name) so you see which function or method a match landed in without a follow-up call. A trigram index narrows the candidate files, so a repo-wide search costs roughly the size of the matching files, not the whole tree. Use for literal-string / regexp lookups; use search_symbols for symbol-name / concept queries."),
 			mcp.WithString("query", mcp.Description("Literal substring (case-sensitive) to search for — or a regular expression when regexp=true.")),
 			mcp.WithBoolean("regexp", mcp.Description("Treat query as a regular expression instead of a literal substring. An invalid pattern is returned as a tool error. Default false.")),
-			mcp.WithNumber("limit", mcp.Description("Max matching lines to return (default 100, capped at 1000).")),
+			mcp.WithNumber("limit", mcp.Description("Max matching lines to return (default 100, capped at 1000; raise the cap with GORTEX_SEARCH_TEXT_MAX_LIMIT). A result bound by the limit carries _truncated_by_limit and count_is_exact=false — read those before treating count as a total, because a `path` filter cannot recover the remainder.")),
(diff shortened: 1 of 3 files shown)


tiendungdev:
CI was red on me, fixed in `f2ff364c`. Two things I got wrong are worth stating plainly, because the second one invalidates a claim in my PR description.

## What failed

`TestToolsListByteCeilings` — the **core preset**, not the facade surface:

```
core   mode=defer   bytes=97671  (baseline 97500)
core preset must shrink below its pre-diet baseline (97500), got 97671
```

macOS failed on it and ubuntu was cancelled by fail-fast, which is why both test jobs read as failures.

## Mistake 1 — I checked the wrong ceiling

Before pushing I *did* go looking for the #685 hazard, found `TestCompactToolsListIsStaticAndBudgeted`, measured 14984 bytes on both trees, and concluded the cost was zero. That number is real and it is irrelevant: it covers the facade surface, where `search_text` is not present at all. The budget this change actually spends is the core preset's, and the test that guards it is the one I did not run.

The PR description says "the compact tools/list is byte-identical at 14984 on both trees" as evidence the addition was free. It was evidence about a surface my change cannot reach. I have corrected the body.

## Mistake 2 — my isolation measured my own code

When I tried to confirm the cost after the CI failure, I reverted the schema file and re-ran:

```
git checkout -- internal/mcp/tools_analysis.go
```

and got **97671 again** — identical to my branch. I nearly reported that as "the failure is pre-existing on main". It is not: `git checkout --` restores from the index, and the change was already committed, so I had re-measured my own edit and read the matching numbers as proof of innocence.

Checking out `upstream/main` itself settled it:

```
main        core  97438  PASS
my branch   core  97671  FAIL   (+233)
```

Worth recording because the earlier isolation runs in this PR were sound only by accident — those changes were still uncommitted, where `git checkout --` does revert. The technique quietly stops working the moment the work is committed, and it fails in the direction that exonerates the diff.

## The fix

Available headroom on main is **62 bytes**. The description now carries only what an agent cannot learn anywhere else *before* calling:

```
Max matching lines to return (default 100, capped at 1000). A bound result sets _truncated_by_limit.
```

41 bytes, landing at **97479**. The rest of what I had put there — how to widen, and that a `path` filter cannot recover the remainder — already rides on the response in `truncation_note`, which is where a caller reads it after seeing the flag. Both ceiling tests and the `SearchText` suite pass.

If you would rather not spend 41 of the 62 remaining bytes on this at all, say so and I will drop the schema line entirely; the flag is discoverable from the response either way. That is a budget call, not a correctness one.

## For #685

Fresh numbers, since this is the second time in one day I have had to reverse-engineer a budget from a log line:

- core preset: **97438 / 97500 → 62 bytes** of headroom on `main`
- facade `tools/list`: **14984 / 15000 → 16 bytes**

Nothing surfaces either number until CI is red, and neither failure message names the headroom — only a total that has to be subtracted by hand against a baseline stated in the message. I have added this measurement to #685.

zzet (approved):

PR #756, opened 2026-09-05
fix(lsp): guard sourceCache with docMu and close openDocument check-then-act

### LINKED ISSUE #714: daemon dies with fatal concurrent map writes in lsp.Provider.openDocument via relations.callers
## Summary

A `relations.callers` call on a Go repo with gopls attached killed the entire daemon with `fatal error: concurrent map writes`. This is a Go **fatal**, not a panic, so the per-handler recover firewall in `wrapToolHandlerMode` cannot catch it: the process dies and every session's MCP transport drops, not just the offending call.

## Stack (observed, daemon built from `origin/main` @ 217391f8)

```
fatal error: concurrent map writes

goroutine 279 [running]:
internal/runtime/maps.fatal(...)
	/usr/local/go/src/runtime/panic.go:1195 +0x20
github.com/zzet/gortex/internal/semantic/lsp.(*Provider).openDocument(...)
	internal/semantic/lsp/provider.go:2789 +0xf8
github.com/zzet/gortex/internal/semantic/lsp.(*Provider).EnsureFileOpen(...)
	internal/semantic/lsp/provider.go:2961 +0x64
github.com/zzet/gortex/internal/semantic/lsp.(*Provider).ConfirmSymbolRefs(...)
	internal/semantic/lsp/provider.go:3007 +0x170
github.com/zzet/gortex/internal/mcp.(*Server).confirmSymbolRefsOnDemand(...)
	internal/mcp/tools_lsp.go:88 +0x144
github.com/zzet/gortex/internal/mcp.(*Server).handleGetCallers(...)
	internal/mcp/tools_core.go:2681 +0x478
github.com/zzet/gortex/internal/mcp.(*Server).invokeFacadeSpec(...)
	internal/mcp/facade_tools.go:1112 +0x11f4
github.com/zzet/gortex/internal/mcp.(*Server).handleFacade(...)
	internal/mcp/facade_tools.go:618 +0x1100
...
	internal/mcp/tool_deadline.go:154 +0x2e4
```

## Root cause

`openDocument` releases `docMu` after its check-then-act on `openDocs`, then writes `sourceCache` with no lock at all (`provider.go:2773-2789`):

```go
p.docMu.Lock()
if p.openDocs[absPath] {
    p.docMu.Unlock()
    return nil
}
p.docMu.Unlock()          // <- lock released here

content, err := os.ReadFile(absPath)
if err != nil {
    return err
}
if p.sourceCache == nil {
    p.sourceCache = map[string][]byte{}   // <- unguarded map write
}
p.sourceCache[absPath] = content          // <- unguarded map write
```

Both facts are documented in the file, which is why this reads as an invariant that quietly stopped holding rather than an oversight:

- `provider.go:91` — "docMu guards docVersions / openDocs / lastDiag" — `sourceCache` is deliberately **not** in that list.
- `provider.go:2030` — "sourceCache is the unsynchronised interactive-navigation cache".

`sourceCache` was safe while it was only touched from a single interactive path. `confirmSymbolRefsOnDemand` now reaches `openDocument` from the tool-request path, where two concurrent `relations.callers` / `relations.usages` calls on different files race directly on the map. The check-then-act on `openDocs` makes it worse: two goroutines can both pass the `openDocs` gate for the *same* file and race on the same key.

`applyDocumentEdit` at `:2837-2840` and `delete(p.sourceCache, ...)` at `:2864` are the other unguarded writers; the reader at `:3134-3137` is unguarded too.

## Reproduction

Tracked Go repo, gopls attached, then a `relations.callers` facade call concurrent with other tool traffic. Observed once during PR #705 validation on a 6-file fixture — small repos are enough, no special load needed.

## Diagnosing it from the outside

The fatal lands in the daemon log (`$GORTEX_HOME/cache/daemon.log`, or the `daemon start` stdout). The *failing* call itself surfaces as a timeout or a cancelled request; the first clear symptom is that the **next** call returns `daemon unavailable: dial unix ... connection refused`. So the stack trace in the log is the only signal that this happened rather than an ordinary hang.

## Suggested fix

Bring `sourceCache` under `docMu` (or give it its own mutex) at all four sites, and close the `openDocs` check-then-act so one file is opened once. Update the two comments that currently promise the opposite.

Unrelated to #705 — found while validating it.

atirna:
## Summary

`relations.callers` (and any other tool traffic that reaches `EnsureFileOpen`) could kill the whole daemon with `fatal error: concurrent map writes`, because `openDocument` released `docMu` after its `openDocs` check-then-act and then wrote `sourceCache` with no lock at all. That was fine while the cache was only touched from the serialized interactive-navigation path, but `confirmSymbolRefsOnDemand` now funnels concurrent tool-request goroutines into `openDocument`, so two calls on different files race the map writes directly, and two calls on the same file both pass the `openDocs` gate.

## Changes

- `sourceCache` is now guarded by the provider's existing `docMu` at every site: the writes in `openDocument` / `changeDocument`, the deletes in `closeDocument` / `releaseOpenClaim`, the read in `getSource` (RLock), and the reset in `resetForReconnect`. No new mutex — the open/close bookkeeping already held this lock around the same state.
- `openDocument` now claims the path in `openDocs` before its disk read, so concurrent opens of the same file funnel into one `didOpen`. On a read or notify failure the claim (and any bytes cached meanwhile) is released, so a retry after the failure starts clean instead of silently skipping its `didOpen`.
- The two comments that promised the opposite (`provider.go` docMu guard list, `resetForReconnect`) are updated to match.

## Testing

- [x] New regression test `TestOpenDocument_ConcurrentIsRaceFree` (8 goroutines sweeping 12 overlapping files through `EnsureFileOpen`): on unfixed main it fails with 3 `WARNING: DATA RACE` reports inside `openDocument` at the exact #714 stack, and the didOpen count never settles because same-file opens double-fire; with the fix it is race-free, exactly one `didOpen` per file, one cache entry per file, and `getSource` serves the cached bytes.
- [x] New test `TestOpenDocument_FailedOpenLeavesNoClaim`: a missing file errors without leaving an `openDocs` claim or cached bytes, and the retry after the file appears does send its `didOpen`.
- [x] `go test -race -count=1 ./internal/semantic/lsp/` — ok (full LSP package).
- [x] `go test -race -count=1 -timeout 45m ./...` — all packages ok (the four slow ones were re-run with CI's timeout: `store_sqlite` ~16m, `indexer` ~14m, `mcp` ~8m, `agents/opencode`).
- [x] `gofmt -s` clean on the changed files; `go vet ./internal/semantic/lsp/` clean.

Fixes #714

### DIFF
--- a/internal/semantic/lsp/provider.go
+++ b/internal/semantic/lsp/provider.go
@@ -82,14 +82,18 @@ type Provider struct {
 	// interactive session — where hover / definition traffic keeps the
 	// provider's lastUsed fresh and the router never reaps it — does
 	// not retain every navigated file's bytes for the daemon's
-	// lifetime. Accessed lock-free (see getSource, which tolerates a
-	// miss by falling back to col=0): the interactive
-	// open→lookup→close sequence is serialized per file, so it needs
-	// no docMu.
+	// lifetime. Written under docMu: the interactive
+	// open→lookup→close sequence was serialized per file when this
+	// cache was introduced, but on-demand confirmation
+	// (ConfirmSymbolRefs) and the MCP tool path now reach openDocument
+	// from concurrent request goroutines, and an unsynchronised map
+	// write there is a fatal (uncatchable) concurrent map writes
+	// crash, so every access shares the docMu the open/close
+	// bookkeeping already holds.
 	sourceCache map[string][]byte
 
-	// docMu guards docVersions / openDocs / lastDiag so concurrent
-	// callers (LSP push notifications + MCP request goroutines) can
+	// docMu guards docVersions / openDocs / lastDiag / sourceCache so
+	// concurrent callers (LSP push notifications + MCP request goroutines) can
 	// share one client safely.
 	docMu       sync.RWMutex
 	docVersions map[string]int          // absPath → most-recent didOpen / didChange version
@@ -2026,13 +2030,13 @@ func (p *Provider) resetForReconnect() {
 	p.docVersions = map[string]int{}
 	p.openDocs = map[string]bool{}
 	p.lastDiag = map[string][]Diagnostic{}
-	p.docMu.Unlock()
-	// sourceCache is the unsynchronised interactive-navigation cache
-	// (written lock-free by openDocument); drop it on reconnect so a
-	// long-lived provider doesn't carry the dead session's file bytes.
-	// nil is the freed state — openDocument lazily re-creates it and
-	// getSource nil-checks before reading.
+	// sourceCache rides the same reset under docMu (every access is
+	// docMu-guarded); drop it on reconnect so a long-lived provider
+	// doesn't carry the dead session's file bytes. nil is the freed
+	// state — openDocument lazily re-creates it and getSource
+	// nil-checks under the same lock before reading.
 	p.sourceCache = nil
+	p.docMu.Unlock()
 }
 
 // dialOrSpawn builds the LSP client according to the provider's spec.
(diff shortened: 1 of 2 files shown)


zzet:
@atirna thank you for the fix!

PR #757, opened 2026-09-05
fix(mcp): enable guarded source edits in exact worktrees

zzet:
## Problem

The routed-view mutation guard rejected all source edits, even dry runs from a session already inside the selected worktree. Its advice to edit from that checkout's own working copy could not resolve the refusal.

## Fix

Enable the audited single-file paths: `edit.file`, `edit.write`, and `edit.symbol`, including legacy equivalents and inferred facade operations. Own-CWD and explicit path/checkout-ID selectors use the same admission path.

- Acquire a checkout-specific mutation lease using the existing interactive build admission and reconciliation lock. Revalidate route epoch, incarnation, filesystem root identity, HEAD/base identity, and dirty fingerprint.
- Dry runs perform no disk commit, route invalidation, or generation build.
- Withdraw the old dirty route before a real write, preserve already-pinned readers, then refresh through the selected checkout's sparse coordinator. Never enqueue the primary watcher or index the worktree into the canonical corpus.
- Preserve the existing disk-commit receipt if graph refresh fails. Shutdown joins admitted mutations; cancellation and panic paths release admission safely.
- Confine both the physical read target and the atomic-write directory entry to the selected checkout. Refuse primary/sibling paths, symlink escapes in either direction, nested repositories, and Git metadata.
- Keep all inexact fallbacks and immutable ref/commit views read-only. Unsaved editor-buffer cohorts require saving/discarding before an on-disk edit.

Batch editing/recovery, lifecycle operations, and LSP refactors have separate write paths and remain explicitly refused through routed views. This PR does not enable unaudited writes by removing the guard. Ordinary canonical-checkout editing is unchanged.

Implementation notes and reproducible benchmarks are in `docs/worktree-source-mutations.md`. No schema migration or extra store is required.

## Validation

- Full `internal/indexer`, `internal/mcp`, `internal/graphview`, and `cmd/gortex` package suites.
- Full graphview race suite; focused MCP mutation/routing and indexer coordinator/mutation race suites.
- Real temporary Git-worktree E2E through the registered MCP facade: file/write/symbol, explicit/inferred operations, previews and writes, own-CWD/path/ID selection, unchanged primary files, and subsequent exact graph/source reads.
- Failure regressions for stale views, cancellation, shutdown, root replacement, panic cleanup, publication failure, and symlink directory-entry confinement.
- Scoped golangci-lint and CLI build.
- Apple M1 Pro tiny-fixture benchmarks: checkout admission 10.06–10.21 ms/op; complete MCP dry run 12.35–19.55 ms/op. These are admission-path measurements, not large-repository indexing claims.

The CLI suite initially reported its documented false positive when our MCP calls changed the running daemon's query log during its real-home audit. It passed when rerun without concurrent Gortex queries. The live daemon was never restarted or stopped. Validation used temporary fixtures; no explicit tracking, store deletion, or administrative reindex was performed.

### DIFF
--- a/docs/worktree-source-mutations.md
+++ b/docs/worktree-source-mutations.md
@@ -1,0 +1,71 @@
+# Source edits in selected worktrees
+
+An exact, live automatic worktree supports `edit.file`, `edit.write`, and
+`edit.symbol` (and their legacy tool equivalents). The session may already be
+inside the worktree, or select it explicitly with `view.kind: worktree` and a
+registered checkout ID or absolute root path. No explicit tracking or dedicated
+graph is required.
+
+The old mutation guard refused every routed source tool, including dry runs,
+and suggested editing from the checkout's own working copy. That suggestion
+could not work: an automatic checkout's own session also receives a routed view.
+The replacement admits only the audited checkout-aware write paths.
+
+## Safety and freshness
+
+- Admission obtains the existing checkout coordinator's interactive build
+  admission and reconciliation lock. It verifies that the checkout and the
+  route epoch still match the view used to interpret the edit.
+- File targets must belong to that checkout, including after resolving symlinks
+  and existing parent directories. Absolute primary/sibling paths, nested Git
+  repositories, and Git metadata are refused rather than redirected.
+- Dry runs validate against the selected working copy without committing bytes,
+  withdrawing its route, or building a generation.
+- Immediately before a real write, the old dirty route becomes pending. Existing
+  leased readers remain valid; new requests cannot mistake the old generation
+  for the updated working copy.
+- The existing single-file commit ledger still distinguishes disk commit from
+  graph refresh. Refresh uses the selected checkout coordinator, never the
+  primary repository's watcher or incremental indexer. A bounded refresh failure
+  leaves a stale/failed graph outcome and schedules reconciliation; it must not
+  report that a successful disk write did not happen.
+- Coordinator shutdown joins admitted writes before checkout state can retire.
+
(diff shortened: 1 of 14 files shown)


zzet:
Fixed the Windows failures from CI run 33970485298 in three atomic commits:

- `74245b67`: replace the estimate scanner test's 1 ns timer race with already-canceled and already-expired contexts; require the matching error and no partial results.
- `5502ef91`: compare discovered checkout roots using `pathkey.EqualPaths`, since Git and Go can spell Windows paths with different separators.
- `042f400c`: use consistent path identity in the checkout mutation guard. Differently spelled root boundaries must also identify the same physical directory, preserving confinement on case-sensitive volumes. Added root-alias, metadata, nested-checkout, sibling, and distinct-directory regressions.

Validation in isolated test storage, without restarting or modifying the live daemon:
- Store cancellation regression x100; related store race tests x10: PASS.
- Worktree mutation E2E suite x3: PASS.
- Full MCP suite: PASS (188.931 s).
- Targeted worktree/checkout mutation race suite x3: PASS (93.030 s).
- Full MCP race suite, including streamable: PASS (494.201 s / 1.776 s).
- Root-identity regression x100: PASS.
- MCP and store package lint: 0 issues.
- Path-confinement benchmark, five samples: median 66.17 -> 75.21 microseconds/op; 19,504 -> 21,040 B/op. The added identity checks cost about 9 microseconds on this machine.

The new GitHub CI run is [33974842442](https://github.com/zzet/gortex/actions/runs/33974842442), testing head `042f400c`. Platform jobs are still running; Windows has not yet been confirmed green.

PR #759, opened 2026-09-06
fix: unblock automatic worktree discovery and mutation recovery

zzet:
## Summary

Repair the complete agent lifecycle for automatic worktrees:

- Discover a newly added checkout on its first CWD request within an already known Git family, without explicit tracking or a synchronous index.
- Separate successful source writes from asynchronous checkout publication. Return pollable pending receipts; bind publication to checkout incarnation, branch/HEAD, snapshot, and committed bytes.
- Keep receipt inspection and checkout-scoped recovery available while the graph builds. Detect changes in the selected working copy, with incomplete/unknown graph analysis when necessary.
- Preserve read-only fallbacks and reject `require_exact` requests that would otherwise silently receive a fallback.
- Preserve bounded discovery contention through daemon admission as retryable `view_building`, not an erroneous untracked/track response.

The coordinator owns publication independently of request lifetime. Recovery cannot certify another checkout's content, and a successful scoped recovery only retires eligible historical failures from the current-view barrier; historical errors remain inspectable.

## Validation

- Full indexer suite: passed, 401.650 seconds.
- Full MCP suite: passed, 273.348 seconds.
- Full reconcile and graphview suites: passed.
- Repository-wide package/test compilation (`go test -p 2 ./... -run '^$'`): passed.
- Lint across the changed packages: zero issues.
- Real-parser blocked lifecycle, selector variants, and facade variants: passed repeatedly; race run passed in 68.535 seconds.
- Broad worktree/checkout/mutation race suite: passed, 199.629 seconds. Repeated real admission-contention/recovery race test: passed, 12.985 seconds.
- Final admission/prefix-catalog failure race tests ×3: passed, 17.963 seconds. Daemon admission/handshake regression race suite: passed, 8.778 seconds; new wire tests ×3 passed in 13.867 seconds.
- Focused queue, publication identity, cancellation, supersession, SQLite contention, scope, and receipt regressions included.
- Benchmarks and incident analysis recorded in `docs/worktree-source-mutations.md`. Median dry-run latency was 12.38 ms versus 12.07 ms at baseline; blocked-indexer edits returned in 33.55–37.57 ms independently of publication.

New-worktree discovery/search/edit/exact-read succeeds both through the daemon CWD admission hook and direct tool routing. Unknown families remain unregistered.
The real cross-workspace regression confirms rejection before checkout registration/coordinator activation. First-CWD contention is retryable and the same session recovers without changing tracking intent.

## Boundaries

No schema migration, live daemon restart, graph-store rebuild, or tracking-config changes were performed. Validation uses isolated real-Git/SQLite fixtures and in-process MCP handlers/coordinators. This repairs interactive lifecycle behavior; it does not claim to make large dependency-closure indexing inexpensive.

Native graph post-checks against the isolated implementation checkout were attempted but rejected with `view_building` by the currently running daemon. They are not counted as successful validation; the executable tests above exercise the new implementation without replacing that daemon.

### DIFF
--- a/cmd/gortex/daemon_mcp.go
+++ b/cmd/gortex/daemon_mcp.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -26,9 +27,12 @@ import (
 // ID into ctx via gortexmcp.WithSessionID before HandleMessage runs.
 // Tool handlers resolve per-client state through Server.sessionFor(ctx).
 type mcpDispatcher struct {
-	srv          *gortexmcp.Server
-	multiIndexer *indexer.MultiIndexer
-	logger       *zap.Logger
+	// checkoutCWDProbe preserves discovery errors across the daemon admission
+	// boundary. Nil delegates to the server's checked checkout resolver.
+	checkoutCWDProbe func(context.Context, string) (bool, error)
+	srv              *gortexmcp.Server
+	multiIndexer     *indexer.MultiIndexer
+	logger           *zap.Logger
 	// router is held behind an atomic.Pointer so ControlProxy can
 	// publish a freshly-built (or torn-down) router live without
 	// racing a concurrent dispatch read.
@@ -92,7 +96,15 @@ func (d *mcpDispatcher) Dispatch(ctx context.Context, sess *daemon.Session, fram
 	// and tools/list answers an empty/track-only list, instead of a
 	// connection-poisoning errored initialize. Only tools/call is refused (with
 	// the structured not-tracked error the agent can act on).
-	untracked := sess.CWD != "" && !d.cwdReachable(ctx, sess.CWD)
+	// Discovery needs the calling CWD for its scope authorization, but binding
+	// a client session must wait until admission has succeeded.
+	ctx = gortexmcp.WithSessionID(ctx, sess.ID)
+	ctx = gortexmcp.WithSessionCWD(ctx, sess.CWD)
+	reachable, admissionErr := d.cwdReachableChecked(ctx, sess.CWD)
+	untracked := sess.CWD != "" && !reachable && admissionErr == nil
+	if admissionErr != nil && !checkoutAdmissionMetadataMethod(peekFrameMethod(frame)) {
+		return checkoutAdmissionError(sess.CWD, frame, admissionErr), nil
+	}
 	if untracked {
 		d.logUncoveredCWDOnce(sess)
 	}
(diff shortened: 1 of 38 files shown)


zzet:
CI failures were traced and repaired in three atomic commits:

- `2d0cd114`: pending `change.detect` uses explicit file-only Git mapping instead of an empty in-memory graph. The architectural constructor fence remains unchanged.
- `f3adb675`: Windows checkout identity is captured before waits, preventing lazy `os.SameFile` lookup from accepting a replacement directory. Handles are closed immediately; non-Windows capture keeps its cheaper path.
- `a6e8b6f3`: slow first-worktree discovery shares bounded lifecycle-owned work across requests. The caller budget remains 250 ms, jobs retain their slots until exit, authorization is per caller, and cached Git-family proof is revalidated. Pending/stale discovery no longer becomes successful empty search or an unproven base fallback. Windows fixture/path assertions were corrected without skipping coverage.

Validation includes the original failing CI benchmark command, full affected packages, lint, CLI compilation, and repeated race regressions for real Git worktree discovery, search, edit/pending refresh, receipt/recovery, and a second edit. New negative tests cover physical-root and Git-metadata replacement, canceled/expired job capacity, per-caller authorization, and canonical-only fallback authority.

Benchmark results and limitations are recorded in `docs/worktree-source-mutations.md`. Windows identity tests cross-compile locally; Windows runtime validation is left to the new CI run, not claimed from Darwin tests.

No live daemon restart, store reset, or tracking/configuration changes were made. The old running daemon still cannot certify these new checkout post-checks, so validation uses isolated stores and real Git/MCP test fixtures.

Full package results: indexer 424.412s; MCP 216.553s; Git state 15.404s; analysis 7.417s; graph 8.661s; graph views 2.048s. Final lint reports zero issues. The two older automatic-sibling fallback tests were split into positive canonical-fallback coverage and negative unbound-worktree coverage; both now pass, including their metrics assertions.

Pushed head: `a6e8b6f3af4050fe55709acce2f61027e7a457d3`. Fresh CI is running: https://github.com/zzet/gortex/actions/runs/34035213929 . This comment does not claim that the new platform jobs have finished.

zzet:
Follow-up for the remaining macOS failure, pushed as atomic commit `d0863f6d`.

Graph-guided debugging found a real shared-discovery lifetime bug: a foreground validation timeout/cancellation or catalog error canceled an already-successful lifecycle-owned observation job. A subsequent reader could repeat Inventory. The fix cancels only on a positively classified `ErrCheckoutMutationStale`; other errors still refuse the caller. Reauthorization, physical/catalog revalidation, the five-second lifetime, and the active-job bound remain unchanged.

Validation:

- Deterministic regression fails under the old unconditional cancellation for all five non-stale error cases, then passes with this fix. Bare, wrapped, and joined stale errors still cancel and drain.
- Full `internal/indexer` race + coverage: PASS, 928.991s, 85.3% statement coverage.
- Routed/worktree/checkout MCP tests repeated three times under race: PASS, 747.665s, 744 passing test/subtest executions. This includes new-worktree search and end-to-end source mutations.
- New policy race regressions ×3: PASS. Indexer build: PASS. Indexer lint: zero issues.
- Cached-proof benchmark ×3: 137–143 µs/op, about 21.6 KB/op, 288 allocs/op; exactly one Inventory across 2,388 / 2,496 / 2,684 requests. This is reuse validation, not a claimed speedup; the success path is unchanged.

The original timing failure did not reproduce in 20 local race/coverage repetitions. The confirmed bug is consistent with that failure, but the CI log alone does not establish an exclusive cause. The original single-inventory assertion and deadlines are unchanged; synchronized diagnostics now distinguish foreground invalidation from background failure if it recurs.

My initial three-repeat MCP command exhausted its five-minute *total suite* budget while progressing, with no assertion/race failure before timeout. The identical selection and repeat count passed with a fifteen-minute suite budget; no individual test deadline was relaxed.

Findings and measurements are recorded in `docs/worktree-source-mutations.md`. Native graph post-checks were not counted as certification because selected-view change detection reported the canonical repository root. Validation used the isolated worktree and test storage; the live daemon, store, tracking configuration, and unrelated files were untouched.

New CI run is in progress: https://github.com/zzet/gortex/actions/runs/34039347916

Issue #767, opened 2026-09-07
v0.64.0: sustained write amplification (~549 GB/day) — non-converging dirty/sha reconciliation with worktree-tracking state (store 2.16× legitimate size)


is911:
## Summary

A gortex daemon (v0.64.0, macOS arm64, Go sqlite backend, WAL mode) serving one ~1.7k-file Go repo wrote **549.76 GB of file-backed data in 13.55 h** (11.27 MB/s sustained; macOS resource-diag "disk writes" event, limit 6,362.91 KB/s/86,400 s — 86x over; 656/661 sampled stacks in `pwrite`). The store reached 1.33 GB for a graph whose legitimate content is ~617 MB; 295 MB of the file was freelist.

## Evidence

1. macOS resource diag (`disk writes` event, PID 32029, 2026-09-06 09:11:25 → 22:44:13 +08): 549.76 GB / 48,768 s, action none.
2. Store inspected read-only after the daemon died:
   - `edges`: 441,755 rows ≈ 640 MB across 7 indexes over fully-qualified string keys (the `UNIQUE(from_id,to_id,kind,file_path,line,view_gen)` autoindex alone is 142.6 MB);
   - freelist: 75,504 pages = **295 MB (23% of the file)** — heavy delete/reinsert churn;
   - `analysis_generations`: 3 rows, all `state=2` (stale); the newest was built at `build_revision` **1389** by a daemon life that ran **1,389 mutation batches in ~77 minutes — one batch every 2.7 s on a tree nobody was editing**;
   - `repo_index_state.indexed_sha` = `3395325efb59…` — a **REMOVED git worktree's commit** — while the checkout is @`56333de…`, `dirty=1`. The store's `checkouts` family had tracked 3 scratchpad worktrees (all later `removal_grace`);
   - fresh re-index of the same repo: 616,554,496 bytes, freelist 7 pages, 386,957 edges, 88,058 nodes — the bloated store was **2.16x the legitimate size**.
3. Arithmetic: 549.76 GB ÷ 1.33 GB ≈ 413 full-store rewrites; ÷ 1,389 batches ≈ **395 MB per mutation batch** — each batch dirtied ~1/3 of the store's pages.
4. SQLite settings are the documented tuned set (`wal_autocheckpoint` 8,000 pages, `journal_size_limit` 64 MB, `synchronous` NORMAL, TRUNCATE checkpoints at batch end) — not a checkpoint-storm misconfiguration.

## Suspected mechanism

[...]

PR #774, opened 2026-09-08
fix(worktrees): prioritize selected views and bound sparse scans

zzet:
## Summary

- Prioritize demand-selected worktree builds, including coordinators already queued at background priority. Promotion is coalesced, bounded and non-preemptive; it does not reset debounce or cancel active builds.
- Use the existing generation index for positive-generation capability scans and high-water lookup. Preserve generation-zero SQL, isolation, pagination and enrichment.
- Add deterministic scheduling/race regressions, generation-isolation/query-plan tests, same-store benchmarks, and an opt-in isolated daemon test for new worktree → exact search → dry-run/edit → refreshed search → warm restart → removal.
- Document measured results and remaining limitations in `docs/worktree-readiness-performance.md`.

Refs #767.

## Validation

- Full SQLite, indexer and CLI package suites: PASS (59.46s, 462.75s, 67.62s).
- Focused scanner tests ×3 and race ×3; gate/coordinator tests and race ×10.
- Scoped golangci-lint: 0 issues.
- Fixed-binary isolated worktree lifecycle: PASS (43.30s).
- Existing cold/warm isolated idle-I/O regression: PASS; stable generation sequence and graph counts. Process disk-write counter deltas were zero, but logical writes and WAL growth were nonzero.
- Same-store scan benchmark, 128 target edges + 50,000 foreign edges: legacy 4.84–5.02 ms/op vs generation-indexed 0.502–0.537 ms/op (~9.2× at medians). No universal improvement claimed for the small zero-foreign case.
- Independent peer reviews of scanner, gate and lifecycle changes found no blocking issues.

The user's live daemon was not restarted and its store/configuration were not changed. Gortex post-edit certification remained unavailable while the checkout graph rebuilt; reported test results come from local tests and isolated private daemons.

## Scope / remaining latency

This does not bypass an already-running physical build or promise immediate first use under active load. A safe same-tree metadata-only route needs an empty-only builder contract: an empty preliminary plan can still adopt physical work or encounter a changed dirty snapshot. Retirement and other full-finalization costs remain separate profiling work. No schema migration, new index, tracking requirement or relaxed edit/fallback guard is introduced.

### DIFF
--- a/cmd/gortex/issue767_worktree_readiness_integration_test.go
+++ b/cmd/gortex/issue767_worktree_readiness_integration_test.go
@@ -1,0 +1,154 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestIssue767WorktreeReadinessIntegration exercises the agent workflow against
+// an explicitly supplied binary in a private daemon, store, and Git fixture.
+// It never connects to or restarts the user's daemon. Admission under an occupied
+// build gate is covered by deterministic coordinator tests, not wall-clock races
+// between child processes here.
+func TestIssue767WorktreeReadinessIntegration(t *testing.T) {
+	binary := os.Getenv("GORTEX_ISSUE767_READINESS_BINARY")
+	if binary == "" {
+		t.Skip("set GORTEX_ISSUE767_READINESS_BINARY for isolated worktree validation")
+	}
+	if deadline, ok := t.Deadline(); ok && time.Until(deadline) < 6*time.Minute {
+		t.Fatal("use go test -timeout 10m or longer for isolated worktree validation")
+	}
+	binary, err := filepath.Abs(binary)
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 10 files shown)


zzet:
Fixed the Ubuntu CI failure in 33350cae.

The failing LSP enrichment-budget test used a real 400 ms deadline for its fake RPC fixture. Scheduling delays under the full race/coverage CI run could exhaust the budget before hover and hierarchy work ran. The fixture now runs entirely inside testing/synctest, retaining the same deadline, 60 ms reference delays, and both assertions. No production code or live daemon settings changed.

Validation:
- Race + coverage: 600 repetitions across CPU settings 1, 2, and 4 passed.
- Full internal/semantic/lsp package with race + coverage, repeated 3 times: passed.
- Package lint: 0 issues.
- Negative control: a temporary Go build overlay setting the reserve to zero failed both assertions 3/3 times; the compiled constant was verified. This control ran without coverage because the local coverage+overlay combination retained the original non-test constant.
- Same 50-repeat command: 14.868 s before, 2.212 s after.
- Independent review approved; working tree is clean.

Native graph post-checks remained unavailable while the worktree publication was pending; the checks above are local, not native graph certification. New CI run: https://github.com/zzet/gortex/actions/runs/34194526763

Issue #777, opened 2026-09-08
Windows v0.64.1: exact worktree reads succeed, but edit.write including dry-run fails mutation admission after restart


tiendungdev:
## Summary

On Windows/amd64, explicit worktree reads succeed with `freshness.exact=true`, while `edit.write` repeatedly fails mutation admission, including with `dry_run:true`. Restarting the daemon and reorienting the MCP session restored reads but did not restore writes.

## Environment

- CLI and daemon: `v0.64.1+173cad8`
- Build: `2026-09-07T22:34:17Z`
- Go: `go1.27.0`
- OS/architecture: `windows/amd64`
- Client: Codex MCP client `0.153.4`
- Multi-repository daemon with multiple linked Git worktrees and concurrent MCP sessions.
- Primary checkout and target worktree are on different Windows drives.
- Target worktree was clean; it is an automatically discovered worktree, not separately tracked.
- Daemon reports `ready`. During diagnosis its uptime was approximately 26 minutes (reported warmup approximately 13 minutes).

Personal paths and checkout identifiers are omitted.

## Observed sequence

This reproduces in the affected workspace; it is not yet a minimal repository reproduction.

1. Select an existing linked worktree explicitly with `view:{kind:"worktree",path:"<absolute-worktree-path>"}`.
2. Call `read(operation:"file", ...)`: it succeeds and reports `freshness.exact=true`.
3. Call `change(operation:"impact", ...)`: it also succeeds for that exact view.
4. Attempt `edit(operation:"write", ...)` to create a small Markdown plan file in the same worktree.
5. The write is refused before changing any file.
6. Repeat with `dry_run:true`: the same refusal occurs, measured at approximately 0.32 seconds in the latest call.
7. After a daemon restart, MCP explicitly reports session reset. Reorient and retry: exact reads work, mutation admission still fails.

Both an explicit worktree path and a checkout-ID selector were tried. An absolute target file path was also tried; the result was unchanged.

Sanitized dry-run call:

```json
{
  "operation": "write",
  "dry_run": true,
  "target": {"file": "docs/example-plan.md"},
  "view": {"kind": "worktree", "path": "<absolute-worktree-path>"},
  "content": "# Example plan\n\nDraft only.\n"
}
```

Actual response:

```text
view_read_only: checkout mutation was not admitted; no files were changed:
indexer: checkout mutation lane is busy; retry: context deadline exceeded
```

Git status confirmed that the failed writes did not change the worktree.

## Goroutine evidence

[...]

PR #782, opened 2026-09-09
fix(indexer): queue checkout mutation admission

tiendungdev:
## Summary

Addresses the fixed admission timeout in #777. Checkout source mutation admission queues behind the shared view-build gate and checkout cycle lock until the caller context or coordinator lifetime ends, instead of refusing ordinary contention after a separate 250 ms budget.

This PR is limited to three `internal/indexer` files. The automatic-worktree `strict` change and all associated MCP test edits have been removed; the documented labeled base fallback and its original responsiveness assertions are preserved.

## Behavior and limits

- Contention that clears before the caller deadline allows admission to proceed.
- A deadline reached while waiting for either lane returns an error matching both `ErrCheckoutMutationBusy` and `context.DeadlineExceeded`, including `lane is busy; retry` and the stage (`shared view-build gate` or `checkout cycle lock`).
- Queue capacity exhaustion retains both the busy and queue-full identities. Explicit cancellation, including coordinator shutdown, remains cancellation.
- The default MCP tool budget is still 60 seconds (`DefaultToolCallTimeout`), shared with the rest of the request. A large serialized coordinator build can exceed that budget: this change can turn a 250 ms refusal into a longer wait that still times out. It does not guarantee completion for the large-worktree workload in #777 or extend the tool deadline. The outer MCP timeout wrapper may return its own timeout diagnosis if it wins the deadline race.
- `dry_run:true` still takes the mutation lease and validates the same exact route and disk snapshot as a write, then closes without preparing or committing. It can therefore wait behind a build and time out too. A read-only preview that bypasses this lane needs a separate consistency contract and is deferred; this PR does not claim to resolve that part of #777.
- A failed admission changes neither source nor route. Cancellation releases the queued waiter, any acquired gate, and source-mutation tracking so a later retry can succeed.

## Validation

- Fail-first regression: `go test ./internal/indexer -run '^TestCheckoutMutationContentionDeadline' -count=1 -timeout=5m` failed at both the shared gate and cycle lock because the busy identity was missing.
- After the fix: `go test -race ./internal/indexer -run '^(TestCheckoutMutation|TestViewBuildGate)' -count=1 -timeout=10m` passed on Windows/amd64 with Go 1.27.0 (210.301 s).
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.13.1 run --timeout=5m --new-from-rev=831169a4 ./internal/indexer/...` passed: 0 issues.
- `git diff --check` passed. `git diff 831169a4 --quiet -- internal/mcp` confirmed that no MCP changes remain against the original base.
- All 10 GitHub checks passed at `f669ea7c9762fe8e2be2c057ac37bbe298a6f3b2`: Linux/macOS/Windows tests, lint, static Linux build, ONNX build, benchmark, govulncheck, trivy-fs, and Trivy. [CI run](https://github.com/zzet/gortex/actions/runs/34443274876), [security run](https://github.com/zzet/gortex/actions/runs/34443274958).

## Follow-up

The removed strict-CWD proposal is preserved locally on `split/issue-777-strict-worktree-original` for separate discussion. It is not part of this admission fix. A future proposal should document the fallback contract, `view_building` retry behavior, and responsiveness tradeoff in `docs/mcp.md`.

### DIFF
--- a/internal/indexer/checkout_mutation.go
+++ b/internal/indexer/checkout_mutation.go
@@ -21,8 +21,9 @@ var ErrCheckoutMutationStale = errors.New("indexer: checkout mutation view is st
 // routed generation. It does not mean the disk mutation was rolled back.
 var ErrCheckoutMutationPending = errors.New("indexer: checkout mutation refresh is pending")
 
-// ErrCheckoutMutationBusy refuses admission before source is written. Retry
-// once the active checkout build releases its lane; no primary fallback writes.
+// ErrCheckoutMutationBusy refuses admission before source is written when the
+// queue is full or the caller deadline expires waiting for a build lane. Retry
+// once active work releases its lane; no primary fallback writes.
 var ErrCheckoutMutationBusy = errors.New("indexer: checkout mutation lane is busy; retry")
 
 // CheckoutMutation owns one checkout's physical build lane and route lock for
@@ -50,6 +51,9 @@ type CheckoutMutation struct {
 // the caller materialized. It changes neither disk nor catalog: a dry run may
 // simply close the lease. Gate-before-cycleMu matches background reconciliation
 // so an edit can never hold the route lock while waiting on a build that needs it.
+// Admission waits until the caller deadline or coordinator shutdown, without a
+// separate admission timeout. Dry runs take the same lease to validate the exact
+// route and disk snapshot; they may therefore also wait behind active builds.
 func (l *CheckoutLifecycle) BeginCheckoutMutation(ctx context.Context, checkoutID, expectedRoot string, expectedRouteEpoch int64) (*CheckoutMutation, error) {
 	if l == nil || l.catalog == nil || checkoutID == "" || expectedRoot == "" || expectedRouteEpoch <= 0 {
 		return nil, fmt.Errorf("%w: exact checkout identity and route epoch are required", ErrCheckoutMutationStale)
@@ -159,7 +163,10 @@ func (m *CheckoutMutation) Prepare(ctx context.Context) error {
 }
 
 func checkoutMutationAdmissionError(ctx context.Context, stage string, err error) error {
-	if errors.Is(err, ErrViewBuildQueueFull) {
+	// Preserve the retry diagnosis and the deadline identity for a wait that
+	// ran out. Explicit cancellation (including coordinator shutdown) remains
+	// cancellation, rather than a retryable lane-busy error.
+	if errors.Is(err, ErrViewBuildQueueFull) || errors.Is(err, context.DeadlineExceeded) {
 		return fmt.Errorf("%w: waiting for %s: %w", ErrCheckoutMutationBusy, stage, err)
 	}
 	return err
--- a/internal/indexer/checkout_mutation.go
+++ b/internal/indexer/checkout_mutation.go
@@ -93,7 +93,7 @@ func (l *CheckoutLifecycle) BeginCheckoutMutation(ctx context.Context, checkoutI
 	defer cancelAdmission()
 	releaseGate, err := c.gate.Acquire(admissionCtx, ViewBuildInteractive)
 	if err != nil {
-		return nil, checkoutMutationAdmissionError(waitCtx, err)
+		return nil, checkoutMutationAdmissionError(waitCtx, "shared view-build gate", err)
 	}
 	gateOwned := true
 	defer func() {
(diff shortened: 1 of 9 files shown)


zzet:
Thanks for the investigation write-up — the goroutine samples are exactly the right way to
characterise this, and the diagnosis is correct: a fixed 250 ms admission budget turning ordinary
coordinator contention into a user-visible failed edit is a real defect.

I'd like this split into two PRs before it lands, because it currently carries a second change that
isn't in the title or the Changes list.

## 1. The admission fix (`internal/indexer/checkout_mutation.go`) — good, and separable

I confirmed it stands alone: reverting the `view_request.go` line plus commit fe54632 leaves a clean
three-file change, and `go test ./internal/indexer -run '^TestCheckoutMutation' -count=1` is green.

Two things I'd want addressed in that PR:

**The user-visible error loses its actionable half.** `view_mutation.go` formats whatever
`BeginCheckoutMutation` returns straight into the tool result, so with the gate held by a background
builder and the caller's deadline expiring, the message changes:

```
base   view_read_only: checkout mutation was not admitted; no files were changed:
       indexer: checkout mutation lane is busy; retry: context deadline exceeded
head   view_read_only: checkout mutation was not admitted; no files were changed:
       context deadline exceeded
```

The first line is verbatim what #777 reports. After the change, `ErrCheckoutMutationBusy` is
reachable from `BeginCheckoutMutation` only via `ErrViewBuildQueueFull`, so gate and cycle contention
no longer says "lane is busy; retry" — the one part of the message that told the reporter what to do.
Keeping a typed busy error for "the wait ran out while something else held the lane" would preserve
that, without reintroducing the fixed budget.

**Worth stating the intended behaviour past the caller deadline.** The wait is bounded by
`DefaultToolCallTimeout` (60 s, `internal/mcp/tool_deadline.go:33`), while a coordinator commit build
of a large diverged worktree runs into the minutes and those builds are serialized. For the workload
in #777 that mostly converts a 0.25 s refusal into a ~60 s wait ending in a less specific one. That
may still be the right trade, but the PR should say so — and say whether `dry_run:true`, which never
writes, should be taking the mutation lane at all. That is the reporter's literal complaint.

## 2. `strict: false -> true` in `viewForSessionCWD` — please pull this out

This one line changes a documented contract. A session whose CWD auto-binds to a worktree with a
half-built route used to get a base-corpus answer carrying `exact:false`,
`fallback_reason:view_building`, `actual_view:base`; it now gets a hard tool error. The fallback is
deliberate and never silent — it exists so an agent can keep working while the route builds — and the
MCP server instructions shipped to every client still describe it: "A building overlay may return a
labeled base fallback for eligible graph reads." The checklist item "No schema, public API,
dependency, or primary-checkout write behavior change" doesn't hold for this hunk, and no docs are
updated.

The test edits are what convinced me it needs its own discussion rather than riding along:

```diff
 internal/mcp/worktree_agent_lifecycle_test.go
-require.Less(t, time.Since(started), 2*time.Second, "cold search must not wait for indexing")
+require.Less(t, time.Since(started), 20*time.Second, "cold search must resolve through the exact checkout")
-require.Less(t, time.Since(callStarted), 2*time.Second, "each cold search must remain responsive")
+require.Less(t, time.Since(callStarted), 6*time.Second, "each cold search must remain responsive")

 internal/mcp/checkout_control_test.go
-require.Less(t, time.Since(started), time.Second)
+require.Less(t, time.Since(started), 6*time.Second)
```

Those budgets were the regression guards for the responsiveness the fallback exists to provide;
relaxing them tenfold is the cost of the flip, made concrete. There may well be a good case for
failing closed — I'd rather weigh it on its own, with `docs/mcp.md` updated and the `view_building`
retry story spelled out, than have it land under an admission-timeout title.

Happy to review the split admission PR quickly.

tiendungdev:
Addressed in f669ea7c:

- Removed the strict-CWD change and all associated MCP test edits. The diff is now limited to the three indexer files; labeled fallback and the original responsiveness assertions are restored.
- Gate/cycle deadline errors retain both ErrCheckoutMutationBusy and context.DeadlineExceeded, including the retry hint and contention stage. Regression tests cover cleanup and successful retry.
- Documented the unchanged 60-second tool budget and that dry runs still acquire the lease; bypassing it is deferred to a separate consistency discussion. The strict proposal is also deferred and preserved separately.

All 10 checks are green at this head, including Linux/macOS/Windows tests and lint. Ready for another review of the admission fix. Thanks!

zzet (approved):

PR #787, opened 2026-09-09
fix(pi): hold the first turn until gortex tools are registered

vitaliyslion:
## Summary

The Pi extension could start its first LLM call before its Gortex tools were registered, and the note explaining Pi's tool renaming could be lost for the whole session.

`before_agent_start` was synchronous. It ran whenever Pi's agent loop reached it — which, at startup, can be while `session_start`'s `await c.start()` / `await registerGortexTools(pi)` is still in flight, because Pi arms the editor's submit handler before it awaits `rebindCurrentSession()`. It then latched `orientationInjected = true` unconditionally (`decision.orientation` from the Go hook is effectively always non-empty), so `piAliasNote()` — which reads the not-yet-populated `gortexToolNames` — returned `""` and was never retried.

Two consequences, both observed in a deterministic harness against the live `index.ts`:

- The first turn ran with **zero** Gortex tools registered.
- The model then followed Gortex's bare-name guidance into Pi's own `read` / `edit` and got a denial naming the tool it had just called.

The same shape reaches further than startup: `/reload` and `/new` reach `before_agent_start` before their `session_start` fires at all, and `/reload` re-imports the extension module outright.

## Changes

- **A readiness barrier.** `session_start`'s body moves into `startSession()` wrapped in `try/finally`, so a readiness promise settles on every exit path — including the early `return` when a `/reload` supersedes the bridge mid-handshake. `before_agent_start` is now `async` and awaits it; Pi's `emit()` awaits each listener, so this genuinely holds the turn. Capped at `READY_WAIT_MS = 15_000` — `start()`'s worst case is 60s + 1s backoff + 60s, which must never become first-turn latency.
- **The promise is armed at factory time.** `/reload` re-imports the module (`resource-loader.js` clears the extension cache; the loader runs jiti with `moduleCache: false`) and `/new` builds a fresh runtime, so in both cases an instance exists — and can be asked for a turn — before its own `session_start` runs. Arming at construction closes that window; `armSessionReady()` is idempotent so a turn parked in the gap holds the same promise the incoming `session_start` settles.
- **The rename now rides the tool.** `piAliasNote()` is replaced by `piAliasedDescription()`, which front-loads the explanation into the aliased tool's own description. It cannot be lost to ordering: if the model can see `gortex_read`, it can see what `gortex_read` is. Non-colliding tools are untouched, and tools promoted later by `tools_search` get it too, since they take the same registration path.
- **Orientation is cleared once the `context` push lands**, so it survives a context shape the hook cannot append to.

## Testing

- [ ] New tests added for new functionality — **none in this PR**, see below
- [ ] Benchmarks run if performance-relevant — not performance-relevant

### Test suite

`cmd/gortex/testdata/agent-render/pi.txt` is regenerated in this PR.

### No tests are added by this PR

The behaviour changed here lives entirely in `internal/agents/pi/extension/index.ts`, and there is no JS/TS test harness in the tree today; `internal/agents/pi/adapter_test.go` covers the Go side and stops at the file boundary, so none of the behaviour above is reachable from `go test`. The change is verified by a standalone harness that derives its subject from the live `index.ts` — resolving the four `{{ SENTINEL }}` placeholders exactly as the installer does, and redirecting `node:child_process` to a mock that controls when the simulated `gortex mcp` child answers `tools/list`. It therefore cannot drift from the source: rename a sentinel and it fails loudly.

Four scenarios, 19 assertions, exit 0/1, no dependencies:

| | asserts |
|---|---|
| A — `before_agent_start` fires mid-handshake, nothing awaits `session_start` | barrier held 626ms for a 300ms×2 handshake; 3 tools live at the first LLM call; orientation present; exactly one injected message |
| B — descriptions | `read` → `gortex_read` explains its own rename and keeps its original description; `search` untouched |
| C — registration outlives the cap (forced to 20ms against a 400ms registration) | gave up at 46ms, so no unbounded wait; turn 1 keeps the orientation; nothing injected on later turns; late-registered tools still explain the rename |
| D — prompt submitted in the `/reload` gap, module re-imported | turn held through the gap; resumed only after `session_start` landed; session 2 has its tools and re-injects the orientation |

Pointed at a pre-fix tree the suite fails, on the right assertions — four of D's six, including that the pre-fix code loses the orientation entirely on `/reload` (the turn parks it, then `session_start` clears `pendingOrientation`).

**That harness is not part of this PR, so the checklist item above is unchecked.** Landing it means adding a second toolchain to a Go repo, and that is a decision worth taking on its own: where it runs, whether a missing `node` skips or fails, and whether the tree gets its first `package.json`. Filed separately. Happy to fold it into this PR instead if the maintainers would rather not merge the fix without it — in which case this PR should wait on that decision.

## Checklist

- [ ] Language extractor includes `Meta["methods"]` for interfaces — n/a
- [ ] Methods have `EdgeMemberOf` edges to their containing type — n/a

## Scope and limits

- **Startup ordering is closed by construction; the rest rests on a read of Pi 0.85.1.** At startup there is no `await` between `setupEditorSubmitHandler()` and `emit(session_start)` (`interactive-mode.js:708-712` → `rebindCurrentSession` → `bindCurrentSessionExtensions` → `agent-session.js:1906`), so the handler's synchronous prologue always runs first. If a future Pi inserts an await there, the factory-time arming still covers it.
- **The 15s cap is a judgement call.** A cold daemon plus warmup could exceed it, in which case the turn proceeds without tools. Warm-daemon handshake measured at 41ms, so the cap is only reachable when something is wrong.
- **Denial text is unchanged.** `pretooluse.go:568` still renders `` `read(target:{symbol:"<id>"})` `` — bare names, in Claude Code's vocabulary — and those snippets are what a model copies. With the tool descriptions in place the model can translate those names itself. The real fix is a host-vocabulary mapping at the `internal/hooks/pi.go` boundary. Filed separately; not in this PR.

### DIFF
--- a/cmd/gortex/testdata/agent-render/pi.txt
+++ b/cmd/gortex/testdata/agent-render/pi.txt
@@ -446,18 +446,14 @@ function piAliasName(name: string): string {
   return PI_RESERVED_TOOL_NAMES.has(name) ? PI_ALIAS_PREFIX + name : name;
 }
 
-// piAliasNote tells the model which tools got aliased this session, so it
-// can translate bare names in Gortex's own guidance. Pi-local by design —
-// not a fact any other agent needs.
-function piAliasNote(): string {
-  const aliased = Array.from(gortexToolNames).filter((n) => n.startsWith(PI_ALIAS_PREFIX));
-  if (aliased.length === 0) return "";
-  const pairs = aliased
-    .map((n) => `\`${n.slice(PI_ALIAS_PREFIX.length)}\` -> \`${n}\``)
-    .join(", ");
+// piAliasedDescription front-loads the rename into the tool's own
+// description.
+function piAliasedDescription(bare: string, aliased: string, description: string): string {
+  if (bare === aliased) return description;
   return (
-    `[Gortex] This session renamed these Gortex tools to avoid colliding with Pi's own built-ins ` +
-    `of the same name: ${pairs}. Wherever Gortex's guidance or tool descriptions mention the bare name, call the renamed one instead.`
+    `Gortex's \`${bare}\` tool, registered as \`${aliased}\` because Pi has a built-in ` +
+    `\`${bare}\`. Gortex guidance and denial messages that name \`${bare}\` mean this tool.` +
+    `\n\n${description}`
   );
 }
 
@@ -513,7 +509,7 @@ function registerOneTool(pi: any, desc: ToolDescriptor): void {
   const def: any = {
     name,
     label: name,
-    description: (desc.description || name).trim(),
+    description: piAliasedDescription(name, piAliasName(name), (desc.description || name).trim()),
     parameters,
     async execute(_id: string, params: Record<string, unknown>) {
       if (!client) throw new Error(`gortex ${name}: MCP bridge is not connected`);
@@ -587,6 +583,16 @@ function scheduleSyncTools(pi: any): void {
 // Extension entry point
 // ---------------------------------------------------------------------------
 
+// How long before_agent_start waits for session_start to finish. Pi arms
+// the editor's submit handler well before session_start runs — at startup
+// (interactive-mode.js init()), across /reload (agent-session.js reload()
+// awaits a settings + resource reload first) and across /new
+// (agent-session-runtime.js newSession() awaits teardown + runtime build
+// first) — so a fast prompt can reach the agent loop while registration is
+// still pending or hasn't even begun. Capped so a wedged daemon costs the
+// first turn seconds; start() alone can take ~2 min.
+const READY_WAIT_MS = 15_000;
+
 export default function (pi: any) {
   let orientationInjected = false;
   // Orientation awaiting injection into the next LLM call. The `context`
@@ -594,6 +600,46 @@ export default function (pi: any) {
   // systemPrompt: a systemPrompt change sits at messages[0] and invalidates
   // prefix prompt caching. Computed once per session.
   let pendingOrientation = "";
+
+  // Settles when the current session_start handler is done — bridge up and
+  // tools registered, or the handshake failed. Never rejects.
+  //
+  // Armed at factory time: /reload re-imports this module (resource-loader.js
+  // clears the extension cache, and the loader runs jiti with
+  // moduleCache:false) and /new builds a fresh runtime, so an instance can be
+  // asked for a turn before its own session_start fires. Every path that
+  // constructs an extension goes on to emit session_start, so the wait ends.
+  let sessionReady!: Promise<void>;
+  let settleSessionReady: () => void = () => {};
+  let sessionReadyPending = false;
+
+  function armSessionReady(): void {
+    if (sessionReadyPending) return; // keep the promise parked waiters hold
+    sessionReadyPending = true;
+    sessionReady = new Promise<void>((resolve) => {
+      settleSessionReady = () => {
+        sessionReadyPending = false;
+        resolve();
+      };
+    });
+  }
+  armSessionReady();
+
+  function waitForSession(ms: number): Promise<void> {
+    const ready = sessionReady;
+    return new Promise<void>((resolve) => {
+      let settled = false;
+      const done = () => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        resolve();
+      };
+      const timer = setTimeout(done, ms);
+      (timer as any)?.unref?.();
+      ready.then(done, done);
+    });
+  }
 
   // Pi resets the session's tool registry on every session_start, so
   // (re)register here. Clear the name guard first — it persists across
@@ -603,6 +649,17 @@ export default function (pi: any) {
     orientationInjected = false;
     pendingOrientation = "";
     bridgeError = "";
+    armSessionReady(); // no-op when a turn is already parked on this session
+    try {
+      await startSession();
+    } finally {
+      settleSessionReady();
+    }
+  });
+
+  // startSession holds the body of session_start so the readiness promise
+  // above settles on every exit path, early returns included.
+  async function startSession(): Promise<void> {
     ensureDaemon();
     gortexToolNames.clear();
     if (syncTimer) {
@@ -635,12 +692,15 @@ export default function (pi: any) {
         bridgeError = err?.message || String(err);
       }
     }
-  });
+  }
 
   // Fires before the agent loop's first LLM call. It can't mutate messages
   // itself, so it just parks the orientation for the `context` hook.
-  pi.on("before_agent_start", () => {
+  pi.on("before_agent_start", async () => {
     if (orientationInjected) return;
+    // Pi awaits each listener, so this holds the turn until the tools are
+    // registered and bridgeError reflects the handshake (or the cap expires).
+    await waitForSession(READY_WAIT_MS);
     const decision = callHook({ event: "session_start", cwd: pi?.cwd ?? process.cwd() });
     const parts: string[] = [];
     if (bridgeError) {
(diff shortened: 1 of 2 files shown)


zzet:
Nice work — the diagnosis is right and the redesign is better than what it replaces. Moving the alias
explanation onto the tool's own description is the correct call; I checked the plumbing and it holds:
`registerOneTool` passes the **bare** `desc.name`, `safeRegister` is what applies `piAliasName` to
`def.name`, so `piAliasedDescription(name, piAliasName(name), …)` genuinely names the tool it is
registered under. Non-colliding tools fall through unchanged, and only `read`/`edit` collide today, so
the extra prose lands on two tools.

Two things I'd like changed before this lands, then I'm happy with it.

## 1. The barrier's resolver slot is shared, so a stale `session_start` can settle a newer session's promise

`settleSessionReady` is a single mutable variable that `armSessionReady()` reassigns. If two
`session_start` invocations overlap and a third arms after the first settles, the older handler's
`finally` resolves the **new** promise, and a turn parked on it is released while registration is still
running — the exact failure this PR exists to prevent.

I transcribed the state machine verbatim out of `index.ts` into a standalone node script (no Pi needed)
and drove three invocations: A slow, B fast and overlapping, C starting after B settles.

```
B(fast): registration finished
C: session_start begins (registration will take 400ms)
A(slow): registration finished
TURN released after 180ms
TURN saw: C STILL RUNNING
C: registration finished
```

Binding the resolver at handler entry fixes it:

```ts
armSessionReady();
const settle = settleSessionReady; // this invocation's resolver
try {
  await startSession();
} finally {
  settle();
}
```

Same script with that change: `TURN released after 400ms / TURN saw: C finished`.

Whether this is live depends on something the file argues both ways about. Your new comment says
`/reload` re-imports the module, which means a fresh instance and a fresh closure; but `safeRegister`'s
existing comment says "a reload can re-fire `session_start` without resetting the registry", which
means one instance seeing two. Worth resolving that explicitly in the comment, since the barrier's
safety rests on which is true. Either way the one-liner is cheap insurance.

## 2. On cap expiry the model gets full orientation and zero tools

After `waitForSession` times out, `bridgeError` is still `""` — the handshake is in flight, not failed —
so nothing tells the model Gortex is not ready, and `decision.orientation` is pushed as normal. That is
precisely the combination your own problem statement blames for the model following bare-name guidance
into Pi's `read` and getting denied. The PR says "the turn proceeds without tools", which is true, but
the sharper statement is that it proceeds with guidance to use tools that are not there.

Parking a short "Gortex tools are still registering; retry in a moment" line when the wait times out
would close it, and it costs nothing on the happy path.

## On the harness

I'd lean toward folding it in, and the reason is narrow: the whole barrier rests on Pi's `emit()`
awaiting each listener. Pi is not vendored here and I could not check that claim from the tree. If it is
ever false — or a future Pi changes it — `before_agent_start` returns a promise nobody holds, the barrier
silently does nothing, and there is no test anywhere that notices. That is the failure mode worth buying
coverage for, more than the four scenarios themselves.

That said, "does this repo grow a second toolchain" is the maintainer's call, not mine, and you were
right to raise it separately rather than smuggle it in.

## Verified

- `READY_WAIT_MS = 15_000` is well chosen: `INIT_TIMEOUT_MS` 60s + 1s backoff + retry is ~121s worst
  case, so the cap can only bite when something is already wrong.
- The `context` fix is right — clearing after the push means a shape the hook cannot append to no longer
  eats the orientation.
- The golden drift fence passes, and regenerating with `-update-agent-render` produces no diff, so the
  committed `pi.txt` is exactly what the source renders.
- `go test ./internal/agents/... ./internal/hooks/... ./cmd/gortex/` green locally (16 packages), and all
  12 CI checks are green.

One minor note, not a blocker: a cold daemon now means up to 15 seconds of silence on the very first
prompt with no user-visible signal.

vitaliyslion:
Both taken. Pushed as `c2d6ac65`.

## 1 — resolver slot

Bound at handler entry, as you wrote it. On whether it's live: I traced every path in 0.85.1 that emits `session_start`, and each one constructs a fresh extension instance before emitting, so two invocations never share a closure.

- startup — `interactive-mode.js:712` → `bindCurrentSessionExtensions()` → `agent-session.js:1925 bindExtensions()` → `emit(_sessionStartEvent)`, once per `AgentSession`
- `/new`, `/resume`, `/fork`, JSONL import — `teardownCurrent()` then `createRuntime()`, and `createAgentSessionServices` (`agent-session-services.js:63`) builds a **new** `DefaultResourceLoader` whose `reload()` re-invokes the factory
- `/reload` — `agent-session.js:2216` builds a new `ExtensionRunner` over freshly loaded extensions before emitting at `:2237`; the interactive handler (`interactive-mode.js:5015`) does not rebind afterwards, so there's no second emission

So the failure your script reproduces isn't reachable through Pi's own modes. I took the one-liner anyway, for the reason you gave and one more: `bindExtensions()` emits unconditionally with no idempotence guard, and `core/sdk.js:69` accepts an injected `resourceLoader`, so an embedder sharing one loader across sessions *would* deliver two events to one instance. The invariant is a property of Pi's modes, not of the design.

### On the contradiction

`safeRegister`'s comment was the wrong half, and my new one was also imprecise. `registerTool` is a `Map.set` (`loader.js:238`); it never throws on a duplicate name, and the registry is per instance, so a reload resets it by construction. What the `catch` actually covers is a late registration from a superseded session, which throws via `assertActive()` once Pi has invalidated that instance. Comment rewritten to say that.

And the correction to mine: only `/reload` re-imports the module. `clearExtensionCache()` runs solely inside `resourceLoader.reload()`, so `/new` and fork hit a still-valid cache token (`loader.js:404`) and re-run **only the factory**. Same conclusion — fresh closure per session, factory-time arming required — but a different mechanism, and worth stating precisely, because it's why module-scoped `client` / `gortexToolNames` / `bridgeError` are shared with the outgoing session (whose `startSession()` isn't cancelled by teardown) while the barrier state isn't. That's what the `client !== c` check is carrying.

## 2 — cap expiry

You're right, and it's the sharper point. `waitForSession` now returns a boolean and the expiry case parks a notice ahead of the orientation. Wiring it turned up a second-order version: `orientationInjected` latches on that turn, so a handshake that *fails* after the cap never reaches the `bridgeError` branch — the timeout turn is the only turn that says anything. The notice therefore names `/reload` as the recovery, not just "retry":

> `[Gortex] graph tools were still registering when this turn began, so a tool named below may not be callable yet. If one is missing, say so and tell the user to retry, or to run /reload if it stays missing — don't fall back to native tools.`

I kept the orientation on that turn and kept the latch. Suppressing would re-run the barrier next turn and cost another 15s whenever the daemon is genuinely wedged; the wording is conditional so it stays accurate once the tools land a moment later instead of becoming a stale instruction.

## On the harness

Agreed on the reason, and it stays out of this PR. It's blocked on the bigger question you raised about moving the TS extension into its own repository; I'd rather settle that first than pick a home for the suite twice. The drafts are kept outside the tree in the meantime, so nothing is lost while that's decided.

For what it's worth, the suite does now cover your point 1 directly: driving three overlapping `session_start` invocations through the real module releases the parked turn at 1001ms on the pre-fix code, against 1500ms for the invocation it actually parked on. And your underlying concern is the one I'd most want covered — the barrier rests entirely on `emit()` awaiting each listener (`runner.js:623`, `:881`), Pi isn't vendored, and nothing in the tree would notice if that changed.

## On the cold-daemon silence

Real, and not addressed here. It'll come as its own PR dedicated to the status line, since the fix is a UI surface rather than anything in the barrier.

zzet (approved):

PR #797, opened 2026-09-11
csharp: bind receiverless calls through using static

pbednarcik:
## What

C# `using static Ns.Cls;` (and the compilation-scoped `global using static`) brings `Cls`'s static members into scope by simple name, so a bare `Clamp(1)` may be `Cls.Clamp`. The resolver read the directive for the extension-method form only. A receiverless call went through the generic function-call ladder, which has no using-static tier and binds wherever directory locality points: a same-directory instance method of the same name won at 0.9 `ast_resolved`, and with the global form only the declaring file ever saw the directive.

This adds the receiverless tier in `resolveFunctionCall` for `.cs` callers, placed behind the same-file pick (the graph records no outer-type chain, and a nested type calling an enclosing type's static member is what the same-file tier already binds right) and ahead of the directory tiers:

1. Only what the directive imports: a method carrying the extractor's static stamp that is not an extension method. Instance members are reachable only through an instance, and ECMA 14.6.4 makes extension methods available for extension invocation only, never as static methods. The stamp is written only when true, so a pre-stamp graph misses rather than mis-binds.
2. The caller's own type and its in-graph ancestors get first refusal (C# simple-name lookup order, spec 12.8.4). This reuses `csharpInstanceMemberClaims` with the caller's own class.
3. A candidate whose declaring class (`scope_ns` + `receiver`) is a visible using-static target binds at 0.9 `ast_resolved` with `resolution=using_static`. Exact match, or a dot-boundary suffix for a directive written inside a namespace block, anchored on that enclosing namespace so a top-level `using static Helpers;` cannot reach an unrelated class whose FQN ends in `.Helpers`.
4. Several same-owner overloads surviving the arity window bind the first at 0.75 `ast_inferred`. A keep-rule beside `csharpExtensionGuardKeep` carries that weak-tier bind past the cross-package guard: the directive, not an import, is the scope evidence.
5. Narrowing only: two different owners refuse (a compile-time ambiguity), a candidate set no target owns falls through unchanged, member calls never take it.

Incremental: the restub set (`restubCSharpExtensionBinds`) now includes `using_static` binds, so a `global using static` swap re-prices dependents' bare calls the way it already re-priced their extension binds. A restubbed edge drops the tag before re-resolution, so nothing the cascade binds next inherits it (the hygiene the `extension_method` tag already gets). The reuse snapshot now carries the `resolution` tag through an unrelated save of the calling file: the restub finds the edges to re-price by that tag, and a reused bind that lost it would never be re-priced again. The `extension_method` tag had the same hole.

## Review round 1

- **Local declarations shadow every candidate (P1).** A local function, a delegate-typed parameter or a delegate-typed local named like the call is the callee (spec 12.8.4), and none of those is a node the graph can point at. The extractor now records local functions in the local-scope index (in scope over their enclosing block, before their declaration too) and stamps a receiverless call whose simple name such a binding covers at the call's offset with `local_shadow` on the edge. `resolveFunctionCall` leaves a stamped stub unresolved before any tier runs, so the same-file pick, which bound the same shape at 0.9 on main, stops too. The same gate sits in `CrossRepoResolver.resolveFunctionCall`, beside the existing member-call verdict gate: in a multi-repo daemon its same-repo tier rebinds a per-repo refusal to the first same-repo function with no evidence, which a three-repo lab store showed as the stamped stub coming back bound to the same-directory decoy at confidence 0. Extraction output changes: `csharp@24`. The residual is a member inherited from an external base, which the in-graph ancestor walk cannot see; named in the file header.
- **Reuse tag carry narrowed (P2).** The reuse snapshot carries exactly `extension_method` and `using_static`, the two tags the visibility restub consumes. Every other resolver-authored tag is left to the fresh resolve as before, pinned by a table test.

## Known limits

- Adding a directive to a warm store does not re-price a receiverless call the locality cascade already bound, until its file is next edited or the store is rebuilt. Same class as the type-reference simplification the restub already documents.
- The statics set is file-flat, as it is for the extension binder: a `using static` inside `namespace A` is visible to a sibling `namespace B` in the same file.
- `using static Outer.Inner;` never matches, because a member's owner FQN is namespace plus direct owner and the graph records no outer-type chain. A silent miss, never a wrong bind. For the same reason the tier sits behind the same-file pick, which is what answers a nested type's call to its enclosing type's static member; a nested type whose outer partial lives in another file is not covered.
- An unrelated same-file type's same-named instance method still wins over the directive through the same-file pick. Pre-existing behaviour, unchanged.

## Not done, on purpose

Qualifying a bare name the repo does not declare onto an external target (`Sqrt` under `using static System.Math`). The negative evidence proves nothing: the same bare stub is what a local function (no node is emitted for it), a delegate invocation, `nameof(...)` (extracted as a call) or a member inherited from an external base (a controller's `Ok(...)`) leave behind, and the resolver cannot tell those apart from a static import. A pin test carries the rationale. That attribution needs a member model of the target or the LSP lane.

## Evidence

On my production C# codebase the store carries three `using static` directives, two of them naming in-repo classes, so the in-repo form is a real shape, not a fixture-only one. My counted C# fixture repo gained three cells (file-local directive, global directive from a sibling file, inherited member outranks the directive), with the owner classes in their own directory so the directive is the only evidence. Cold lab A/B with the LSP tiers off, base = main at this branch point:

| Cell | Base (main) | This branch |
|---|---|---|
| file-local `using static`, bare `Notch(1)` | same-directory decoy, confidence 0, untagged (first of two same-directory picks) | owner member, 0.9 `ast_resolved` |
| `global using static` from a sibling file, bare `Plumb(1)` | same-directory decoy, 0.9 `ast_resolved` | owner member, 0.9 `ast_resolved` |
| inherited member with the directive in scope | base member, 0.9 | base member, 0.9 |

The full probe outputs differ only in those two cells. With csharp-ls on, the same base store showed the same two mis-binds, so the LSP lane did not mask them there either.

The review round added two refusal cells to the same fixture: a local function and a delegate-typed parameter shadowing the directive. Both stay unresolved on this branch and on main, and they fail only against the first cut of this PR, which is how the cross-repo gap above was found. Re-run after the rebase, the probe output is byte-identical to the pre-rebase run.

## Tests

- `TestResolveCSharp_UsingStaticBindsBareCallToTargetMember`, `..._GlobalUsingStaticBindsBareCallUnitWide`, `..._GlobalUsingStaticBindsOnScopedResolve`: red at head (decoy bound), green here.
- `TestResolveCSharp_EnclosingTypeMemberOutranksUsingStatic` (base in another file): red with the enclosing-chain check removed. `..._SameFileBaseMemberStaysWithTheSameFilePick`: the same-file pin.
- `TestResolveCSharp_UsingStaticNeverBindsInstanceMembers`, `..._NeverBindsExtensionMethodsAsStatic`: each red with its exclusion removed.
- `TestResolveCSharp_UsingStaticSuffixMatchIsAnchored` (red unanchored) and `..._InsideNamespaceBlockResolvesRelative` (the positive half).
- `TestResolveCSharp_OuterTypeStaticMemberOutranksUsingStatic`: pin for the placement; red with the tier ahead of the same-file pick.
- `TestResolveCSharp_SameOwnerOverloadsSurviveTheGuard`: red without the keep-rule (the guard reverted the pick to unresolved).
- `TestResolveCSharp_RestubbedUsingStaticBindDropsStaleTag`: red without the tag drop.
- `TestResolveCSharp_TwoUsingStaticOwnersRefuse`, `TestResolveCSharp_ExternalUsingStaticTargetLeavesBareCallUnresolved`: pins for the refusals.
- `TestIncrementalSave_GlobalUsingStaticEditReresolvesReceiverlessBinds`: red with the restub widening removed. `..._UnrelatedCallerEditKeepsUsingStaticTag`: red with the reuse tag carry removed.
- Review round 1: `TestCSharpExtractor_ReceiverlessCallShadowedByLocalIsStamped` (local function before and after use, delegate parameter, delegate local, plain control), `TestResolveCSharp_LocalFunctionShadowsUsingStatic` (the review fixture verbatim), `..._LocalFunctionShadowsSameFilePick`, `..._DelegateParameterShadowsUsingStatic`, `..._LocalShadowSurvivesTheCrossRepoPass` (per-repo pass then `NewCrossRepo(g).ResolveAll()`), `TestReuseResolutionTag_CarriesOnlyTheVisibilityTags`. All red before the second commit; with the per-repo refusal removed the three per-repo refusals go red, with the local-function scope entry removed only the local-function extractor cases go red.

Salt: `csharp@24` (the `local_shadow` call-edge stamp is new extraction output). The first commit alone needed none.


zzet (changes requested):
Requesting changes for one correctness issue. The rest of this is solid work — the tier is tightly gated, and the tests run the real C# extractor rather than hand-built nodes, so `Meta[static|receiver|scope_ns]` and the file's `using_static` stamp are genuine extraction output. Nine mutants bind: the enclosing-type refusal, the extension exclusion, the static-stamp check, the suffix anchoring, the two-owner refusal, the stale-tag drop, the guard keep-rule, the tier's placement behind the same-file pick, the restub widening, and the reuse tag carry. Merged onto `main` at 54dce10d the build is clean and `internal/resolver`, `internal/indexer` and all of `internal/semantic/...` are green, golangci-lint 0 issues.

## [P1] A local function shadowing the directive now binds wrong at 0.9

C# §12.8.4 puts the enclosing method's local declaration space ahead of any using-static import, so a local function wins the simple name. The tier binds the directive's member anyway.

```csharp
// Util/Helpers.cs
namespace Util {
    public static class Helpers {
        public static int Clamp(int x) { return x; }
    }
}

// App/Caller.cs
using static Util.Helpers;
namespace App {
    public class Runner {
        public int Run() {
            int Clamp(int x) { return x + 1; }   // the real callee
            return Clamp(1);
        }
    }
}
```

A/B through `ResolveAll`, base = `main` at this branch point:

| | edge |
|---|---|
| main | `App/Caller.cs::Runner.Run -> unresolved::Clamp`, untagged, confidence 0 |
| this branch | `-> Util/Helpers.cs::Helpers.Clamp`, `ast_resolved`, **0.90**, `resolution=using_static` |

An honest unresolved becomes a confident edge to the wrong function, and everything downstream that reads `ast_resolved` — callers, impact, dead-code — inherits it.

There is no fix available inside the resolver. The C# extractor emits no node and no marker for a local function: extracting the caller above yields the file, package, `Runner`, `Runner.Run`, and the bare `unresolved::Clamp` call edge, and nothing else. Closing this needs the extractor to stamp the enclosing method with its local-function names, which is a salt bump.

The part worth arguing is the asymmetry. "Not done, on purpose" names this exact ambiguity — *"the same bare shape is what a local function (no node is emitted for it) … leaves behind, and the resolver cannot tell those apart from a static import"* — and treats it as disqualifying for an external target while binding an in-repo target at 0.9 on the same evidence. Whether the target happens to live in this repo says nothing about which of the two the caller meant. Either the in-repo bind needs the same refusal, or it needs to sit at the inferred tier where a wrong pick is cheap, or the asymmetry needs an argument the doc comment does not currently make.

Same class, lower frequency: `csharpEnclosingTypeClaims` is in-graph proof only, so a member inherited from an *external* base loses to the directive too. That one is already named in the code comment; the local-function case is not.

Not a vector, for the record: `nameof(Clamp)` extracts as `unresolved::nameof`, so it never reaches the tier.

## [P2] The reuse tag carry changes more than C#

`reuseResolutionTag` / `applyReuseResolutionTag` re-stamp *every* `resolution` value through the reuse, not just `extension_method` and `using_static` — `scope`, `import_closure`, `value_callee`, `useClass_binding` and `import_binding` ride along too.

Of those, only `useClass_binding` (resolver.go, the DI-provider tier) writes its tag without also setting `Origin`. So a reused TypeScript DI bind now carries `resolution` with `Origin == ""`, which moves it off `tstypes.claimable`'s `mc && !hasResolution && !hasSemantic` branch and onto the `OriginRank` path, where confidence 0.9 backfills to `ast_resolved` and the edge stops being claimable.

That is the behaviour `claimable`'s own comment asks for — *"Explicitly stamped tiers, DI binds (resolution marker), and provider edges (semantic_source) keep their rank"* — so this reads as fixing a reuse hole rather than opening one, and the tstypes suite is green either way. But it is a cross-language change that the PR description frames as C#-only, and nothing pins it. Worth a line in the description and a test, or an explicit narrowing of the carry to the two tags the restub actually consumes.

pbednarcik:
Both folded in the second commit.

**P1.** Correct, and the asymmetry was real: the "not done" paragraph treated the local-function ambiguity as disqualifying for an external target and ignored it for an in-repo one. Fixed at the source you named. The extractor now records local functions in the local-scope index (in scope over their enclosing block, before their declaration too) and stamps a receiverless call whose simple name a parameter, a local or a local function binds at the call's offset with `local_shadow` on the edge; `resolveFunctionCall` leaves such a stub unresolved before any tier runs. That covers your fixture, a delegate-typed parameter and a delegate-typed local with one rule, and it also stops the same-file pick, which bound the same shape at 0.9 on main. `csharp@24`. Your fixture is a test verbatim and was red before the change; the residual you named (a member inherited from an external base) stays named in the file header.

One more place needed the same gate, which the fixture caught only on a real multi-repo store: `CrossRepoResolver.resolveFunctionCall`'s same-repo tier. With the per-repo refusal in place the stamped stub still came back bound to the same-directory decoy at confidence 0, because that tier rebinds any leftover unresolved call to the first same-repo function with no evidence. The gate now sits beside the member-call verdict gate there, with a test that runs the per-repo pass and then the cross-repo pass.

That tier's wider shape is pre-existing and this PR does not close it: it overturns the two-owner refusal the same way, and any per-repo refusal that leaves no marker on the edge, whenever a daemon tracks more than one repo. A per-repo verdict marker the cross-repo pass honours would cover the class; happy to take it as a follow-up if you want it.

**P2.** Narrowed to the two tags the visibility restub consumes, with a table test over the other tag values. The description no longer claims the carry for anything else.

Rebased onto main at the current tip. Locally the resolver and semantic packages are green, as are the C# extractor and incremental tests; the parser and indexer packages show only the pre-existing Windows failures, which fail the same way on a pristine main checkout here.

zzet (approved):

Issue #803, opened 2026-09-14
Language support: VB.NET (regex tier)


conradcoffman:
**Is your feature request related to a problem?**

VB.NET files are not indexed today. On a .NET estate that mixes C# and VB, the
C# half resolves into the graph and the VB half is invisible, so call chains
and blast-radius queries stop at the language boundary rather than at a real
edge of the system.

**Describe the solution you'd like**

A regex-tier (tier 2) extractor claiming `.vb`, alongside `abap.go` /
`apex.go` / `al.go`. VB is keyword-delimited (`Class`/`End Class`,
`Sub`/`End Sub`), so block extents come from the existing
`findKeywordBlockEnd` helper in `helpers_indent.go`. No new shared helpers.

`.aspx` is out of scope. Web Forms markup is a separate concern from the VB
language, the way `razor.go` claims `.razor`/`.cshtml` separately from
`csharp.go`; nothing claims `.aspx` today and if it is wanted it belongs in
its own extractor.

**Use case**

MCP tools and CLI, same as any other language extractor: `find_usages`,
call-chain and blast-radius queries over mixed C#/VB .NET repositories, where
today the VB side contributes nothing to the graph.

**For new language support:**
- Language: VB.NET (Visual Basic .NET)
- Tree-sitter grammar available in `go-tree-sitter`: **no**. Also none in
  `go-sitter-forest`: no `vb`/`vbnet`/`basic` entry among the 170 forest
  imports in `forest_registrations.go`, and nothing in `go.mod`/`go.sum`. So
  tiers 1 and 3 are both unavailable and this has to be tier 2.
- Key constructs to extract: namespaces, classes / modules / structures /
  interfaces / enums, delegates, P/Invoke `Declare`s,
  `Sub`/`Function`/`Property`/`Event` members, `Imports`,
  `Inherits`/`Implements`, and call sites. Interfaces carry `Meta["methods"]`
  for IMPLEMENTS inference.

---

Two things I would rather hear your view on before sending a branch, since
they are the parts most likely to draw an objection.

**1. Node identity.** I followed `csharp.go` (owner-qualified
`File::Type.Member`, constructors on `Type.<init>`) rather than the flat
`filePath::name` the other regex-tier extractors use, reasoning that VB and C#
share the .NET member model and get queried side by side in one store. On a
~196-file production VB estate the flat scheme collides on 179 of 2945
declarations (6.1%) across 113 name groups, mostly `Sub New` overloads and
[...]

PR #805, opened 2026-09-14
feat(languages): add VB.NET extractor (regex tier)

conradcoffman:
﻿Refs #803.

## Summary

Adds a VB.NET extractor at the regex tier (tier 2), claiming `.vb`.

`go-sitter-forest` ships no VB grammar, so tiers 1 and 3 are unavailable:
there is no `vb`/`basic` module among the 170 forest imports in
`forest_registrations.go`, none in `go.mod` or `go.sum`, and no other
extractor currently claims `.vb`.

VB is keyword-delimited (`Class`/`End Class`, `Sub`/`End Sub`), so block
extents come from the existing `findKeywordBlockEnd` helper rather than brace
matching or indentation. No new shared helpers are introduced.

`.aspx` is deliberately not claimed. Web Forms markup is a separate concern
from the VB language, and this codebase already draws that line: `csharp.go`
handles C# while `razor.go` separately claims `.razor` and `.cshtml`, and
markup extractors have their own registration group. Nothing currently claims
`.aspx`; if it is wanted, it belongs in its own extractor the way Razor does
rather than bolted onto the language. (On the estate this was measured
against, those files also carry essentially no declarations.)

## Changes

- `internal/parser/languages/vbnet.go` (new, 474 lines). Emits namespaces,
  classes/modules/structures/interfaces/enums, delegates, P/Invoke
  `Declare`s, `Sub`/`Function`/`Property`/`Event` members, `Imports`,
  `Inherits`/`Implements`, and call sites.
- `internal/parser/languages/vbnet_test.go` (new, 865 lines). 22 tests,
  43 subtests.
- `internal/parser/languages/register.go`. One `reg.Register` line plus a
  comment, placed with the other enterprise regex extractors and before
  `registerForestLanguages`.
- `docs/languages.md`. One row in the enterprise/legacy table.

### Node identity follows `csharp.go`, not the flat scheme

Members are owner-qualified (`File::Type.Member`) and constructors land on
`Type.<init>`, matching `csharp.go` rather than the flat `filePath::name`
used by `abap.go`/`apex.go`/`al.go`. VB and C# share the .NET member model
and get queried side by side in one store, so the shapes should match.

This is load-bearing, not cosmetic. Measured over a ~196-file production VB
estate, a flat name-keyed ID collides on 179 of 2945 emitted declarations
(6.1%) across 113 distinct name groups -- `Sub New` overloads and same-named
members of two classes in one file. Owner-qualifying removes those; the 50
that remain are genuine same-signature collisions and take the shared
`disambiguateID` `_L<line>` suffix.

### Interface method sets

Interface nodes carry `Meta["methods"]`, so `InferImplements` can derive a VB
interface's contract the same way it does for C#. Only `Sub` and `Function`
join the set: a `Property` is emitted as a field and an `Event` as an event.
The accumulator is keyed on the interface's node ID rather than its name,
consistent with the owner-qualified IDs above.

### Deliberate omissions

- Plain fields / `Dim` / `WithEvents` are not emitted. On that same estate
  there are 1319 `WithEvents` lines against 2945 declaration nodes, and they
  are overwhelmingly designer-generated control fields that would swamp the
  graph.
- Bare parenthesised calls are not emitted. VB uses `(` for array indexing,
  default properties and conversions as well as invocation, so `foo(i)` is
  ambiguous without types. Only qualified (`x.M(`) and explicit (`Call M(`)
  forms produce CALLS edges.
- `New X.Y(...)` is filtered out of CALLS and re-emitted as a type
  reference, since it names a type rather than invoking a member `Y`.

### Known limitation

An indexed property access (`dt.Rows(0)`, `dr.Fields(i)`) is syntactically
identical to a call in VB and is still emitted as one. Resolving it needs
type information this extractor does not have. A hardcoded accessor
blocklist (Rows/Cells/Fields/...) was considered and rejected as too
framework-specific. The behaviour is asserted in a test so it is visible in
the suite and any future fix registers as an intentional change.

The provenance model already reflects the uncertainty. These CALLS edges
resolve to `unresolved::<name>`, so with Confidence unset `DefaultOriginFor`
classifies them `text_matched` (`EdgeTierScore` weight 0.4) rather than the
`ast_resolved` (0.9) that the structural DEFINES / IMPORTS / MEMBER_OF /
IMPLEMENTS edges receive, since `EdgeCalls` is deliberately outside the
structural case list. The extractor therefore leaves Origin unstamped and
lets the dispatcher classify, as `abap.go`, `apex.go`, `al.go` and
`razor.go` do. An ambiguous call edge is already the weakest tier rather
than being asserted as fact.

## Testing

Verified on Windows with Go 1.27.1 and golangci-lint v2.13.1 (the version
`.golangci.yaml` pins):

- `go test -count=1 -run TestVBNet ./internal/parser/languages/` -- 22
  top-level tests and 43 subtests pass, 0 failures.
- `golangci-lint run ./internal/parser/languages/` -- `0 issues.`
- `gofmt -s -l` on the committed blobs -- empty.
- Full `./internal/parser/languages/` suite -- the only failures are six
  pre-existing `TestSubprocessExtractor_*` cases, which fail identically on
  an unmodified checkout of this branch's base (they shell out to `sh` with
  `printf`/`cat`, which does not work on Windows). Byte-identical failure
  sets with and without this change.

The `-race ./...` box is left unchecked deliberately: I ran the affected
package rather than the full repo, and without `-race`, on Windows. CI on
`ubuntu-latest` is the authoritative run for that box.

Benchmarks: not added. `bench_all_langs_test.go` covers 26 languages and
none of the comparable regex-tier extractors (`abap`, `apex`, `al`,
`matlab`) appear there, so adding one only for VB looked inconsistent.
Happy to add it if you would rather every new language have one.

Coverage added: CRLF equivalence (the estate is Windows-origin, and CRLF
input must extract byte-identically to LF including line numbers), mixed and
lone-CR endings, leading BOM, determinism across 26 passes, a 26-case
modifier matrix, mixed-case keywords, the `New`-is-not-a-call filter, the
indexed-property limitation, and 17 malformed inputs (truncated,
unterminated, non-UTF8, NUL bytes, comment-only, pathological repetition)
asserting no panic plus ID-uniqueness and extent invariants.

The CRLF and `New` guards were mutation-tested rather than assumed. Removing
the `'\r'` case from `helpers_indent.trimmed` collapses every block extent
under CRLF, demoting every method to a file-scope function and dropping all
MEMBER_OF and CALLS edges; the new tests catch it and every pre-existing
LF-only fixture still passes.

## Checklist

- [x] Language extractor includes `Meta["methods"]` for interfaces (if applicable)
- [x] Methods have `EdgeMemberOf` edges to their containing type (if applicable)

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


zzet (changes requested):
Hey @conradcoffman, thank you for your interest in the project!

I found two correctness issues in the new extractor:
1. Nested classes lose their outer members. vbnet.go:168 uses findKeywordBlockEnd for the outer class’s extent; that helper stops at the first End Class. In a valid nested class, this is the inner class’s terminator. A method declared afterward in the outer class is then emitted as a free function, without its MEMBER_OF edge. Match nested terminators and add a regression test.

2. Declare procedures lose their owner. vbnet.go:279–284 always emits a file-scope function. VB requires a Declare inside a class, structure, or module, so every valid one is missing its owner and MEMBER_OF edge. Emit it through the member path and test a valid declaration inside a container.

Once those cases are fixed and CI passes again, I’d be comfortable merging.

conradcoffman:
Thanks for the precise pointers. I reproduced both before changing anything, and both are real.

**1. Nested containers.**

```vb
Public Class Outer
    Public Class Inner
    End Class
    Public Sub OuterMethod()
    End Sub
End Class
```

`Outer` came out with an extent of L1-3 instead of L1-6, so `OuterMethod` fell outside it and was emitted as a file-scope `function` with no MEMBER_OF edge, exactly as you described.

Fixed with a VB-local `vbContainerEnd` that counts nesting depth instead of stopping at the first terminator. I deliberately did not touch `findKeywordBlockEnd` itself, since Fortran and ERB depend on its first-match behaviour: both pass a bare `end`, which `hasPrefixWord` also matches against `end do` and `end if`, so depth-counting inside the shared helper would change their extents. Only the same keyword pair is counted, so a `Structure` nested inside a `Class` does not perturb the class depth, and an unterminated block still degrades to `startLine` exactly as `findKeywordBlockEnd` does.

**2. Declare.**

A `Declare` inside a `Module` was emitted at file scope with no owner. It now goes through the same member path as `Sub`/`Function`, so it picks up its owner, the owner-qualified ID and the MEMBER_OF edge, and keeps `Meta["external"]`. The free-function arm survives only for a `Declare` at file scope, which is invalid VB but still parseable, so malformed input degrades rather than disappearing.

**Regression tests** for both: `TestVBNetExtractor_NestedContainerExtent` (which also asserts the inner member still belongs to the inner class, not the outer one) and `TestVBNetExtractor_DeclareInsideContainer`.

**Blast radius**, since it confirms these were not theoretical. Re-running the extractor over the ~196-file production VB estate I had measured against:

| | before | after |
|---|---|---|
| `function` nodes | 6 | 0 |
| `method` nodes | 1812 | 1818 |
| `member_of` edges | 2184 | 2190 |

Node and edge totals are otherwise unchanged (2945 symbols, 2945 defines, 15799 calls), so nothing was gained or lost, only correctly attributed. Six real declarations had been losing their owner. Zero file-scope functions is the expected steady state for VB.NET, where every declaration must live in a type or module, so that count reaching zero is a useful invariant in its own right.



---

*Correction, added after posting.* The paragraph above originally named SQL alongside Fortran as a caller of `findKeywordBlockEnd`. That is wrong: `sql.go` and its four siblings call it zero times. I took it from the helper's own doc comment rather than checking. The restraint itself is unaffected, and rests on Fortran (5 call sites) and ERB (2), both of which pass a bare `end` that `hasPrefixWord` matches loosely. The same wrong word is in the message of commit 913f526, which I have left alone rather than rewrite a branch that is already under review.

conradcoffman:
Both requested changes are in at `913f5265`, each with a regression test: the nested-container extent (`vbContainerEnd`, depth-counted on the matching keyword pair only) and `Declare` ownership. Reproduction, reasoning and the before/after counts over the production estate are in the comment above.

Two things worth flagging rather than leaving for you to find.

**I corrected a factual error in that comment.** It named SQL alongside Fortran as a caller of `findKeywordBlockEnd`. That is wrong: `sql.go` and its four siblings call it zero times, and I had taken the claim from the helper's own doc comment instead of checking. The restraint is unaffected and rests on Fortran (5 call sites) and ERB (2), both of which pass a bare `end` that `hasPrefixWord` matches loosely against `end do` and `end if`. The same wrong word is in this head commit's own message; I left it there rather than rewrite a branch that is already under review.

**CI has never run on this head.** Both workflow runs sit at `action_required`, so `913f5265` carries zero check runs and the PR reads UNSTABLE. That needs a maintainer to approve workflows for a fork PR; there is nothing I can do about it from this side. Locally the four checks are clean: `gofmt`, `golangci-lint`, the `TestVBNet` subset, and the full `internal/parser/languages` suite.

I do not have push access here so I cannot formally re-request review. Please treat this as that request.

conradcoffman:
One more of the same class, found while verifying the two you named.

Namespaces were the remaining nestable container still on `findKeywordBlockEnd`, at `vbnet.go:144`, a little above the class call site you cited. An outer `Namespace` ended at the first inner `End Namespace`, and every type declared after that inner block lost its `scope_ns`:

```
NODE Outer       package   L1-6     <- should be L1-11
NODE Inner       package   L3-6
NODE InnerType   type      L4-5     scope_ns=Inner
NODE OuterType   type      L8-9     scope_ns=<nil>    <- should be Outer
```

Milder than the class case, and I want to be straight about that rather than oversell it: types are not members of a namespace here, so no MEMBER_OF edge is lost and graph topology is unaffected. Only the namespace node's extent and the type's `scope_ns` metadata were wrong. Fixed by pointing it at the same `vbContainerEnd`, with `TestVBNetExtractor_NestedNamespaceExtent` covering it, including the assertion that an inner type still resolves to the innermost namespace now that both cover it.

The common dotted form `Namespace Acme.Demo` is unchanged: one terminator, depth 1 to 0 at it, same result as before.

That leaves exactly one `findKeywordBlockEnd` call in the file, for `Sub`, `Function`, `Property` and `Event`. Those do not nest in VB.NET, and the bodyless cases (interface members, `MustOverride` members, auto-properties) rely on its degradation to a single-line extent, so that one is deliberate rather than overlooked.

zzet (approved):

